PR #7, opened 2015-09-17
Fixed issue#3 gc cycling problem

terrytangyuan:
We might also need to test and benchmark on larger dataset. Any suggestions?

### DIFF
--- a/metric_learn/lmnn.py
+++ b/metric_learn/lmnn.py
@@ -5,6 +5,7 @@ TODO: periodic recalculation of impostors, PCA initialization
 """
 
 import numpy as np
+import gc
 from collections import Counter
 from sklearn.metrics import pairwise_distances
 from base_metric import BaseMetricLearner
@@ -76,6 +77,7 @@ class python_LMNN(_base_LMNN):
 
     # main loop
     for it in xrange(1, self.params['max_iter']):
+      gc.collect()
       df_old = df.copy()
       a1_old = [a.copy() for a in a1]
       a2_old = [a.copy() for a in a2]


perimosocordiae:
I'd be happier if we fixed the memory leak at its source, rather than calling `gc()` to hide it.

To do so, perhaps something like http://stackoverflow.com/a/29923561/10601 might help?

terrytangyuan:
You mean to track it down to a specific problem in the for loop, right?

perimosocordiae:
Correct.

terrytangyuan:
I'll look into it. But here's the memory profiling during `TestLMNN` to see if this rings anything over your head:
```
  types |   # objects |   total size
======================================= | =========== | ============
                                   list |        6340 |    647.59 KB
                                    str |        6344 |    373.32 KB
                                    int |         546 |     12.80 KB
                                   dict |          10 |      4.23 KB
                     wrapper_descriptor |          20 |      1.56 KB
                      method_descriptor |          13 |    936     B
                                weakref |           9 |    792     B
                          numpy.ndarray |           3 |    240     B
                      member_descriptor |           3 |    216     B
                                   code |           1 |    128     B
                  function (store_info) |           1 |    120     B
                                   cell |           2 |    112     B
                      getset_descriptor |           1 |     72     B
  <class 'metric_learn.lmnn.python_LMNN |           1 |     64     B
                                  tuple |           0 |      8     B
```

perimosocordiae:
I expect that a reference cycle is getting introduced somewhere, which translates into a bunch of objects not getting cleaned up when they go out of scope.

perimosocordiae:
What happens if you call `print_diff()` after each iteration of LMNN's main loop?

perimosocordiae:
I wasn't able to find any memory leaks in the main loop. I did find one source of extreme memory usage, though, which I tried to address in commit 669456f6fcd6cf20c118a49a65197c53bc713b03.

The problem was that the `impostors` array can be huge, which meant that the `g0` computation ballooned in memory usage. We could do even better by writing `_pairwise_L2` in Cython, but I think this is good enough for now.

terrytangyuan:
Sorry for my late reply. Looks good. The [scikit-learn.metrics.pairwise.pairwise_distances()](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/metrics/pairwise.py) seems to have a parallelized option now by adding this parameter `n_jobs`, which is probably implemented in `Cython`. Maybe we should upgrade to latest version of `scikit-learn` to use this instead?

perimosocordiae:
I looked at what sklearn is using and found their equivalent here: https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/extmath.py#L57

(The issue is with row-wise / paired distances, not the all-pairs distances that `pairwise_distances` computes. The names aren't the best, I know.)

For something so simple, I just updated our function: 46a0711bd1bde3e6de77f7774ce4eb62fb5a2785

terrytangyuan:
Awesome! I was in the wrong direction. This is super quick. Feel free to just close this PR.

PR #9, opened 2015-09-22
Initial doc page

terrytangyuan:
Just some initial code. Not sure what the doc structure you prefer though.


terrytangyuan:
@perimosocordiae Travis failed due to this `ill-conditioned problem` again. Have you added a fixed seed for the test?

terrytangyuan:
Run the following and then go to `doc/build/html` to check the html page built by `Sphinx`:
```
pip install Sphinx
cd metric-learn/doc
make html
```

perimosocordiae:
Re: the flaky test, it's something that I'm aware of but haven't bothered to fix yet. I'll probably add a check in the code to prevent it at the source, if possible.
In the mean time, for non-code changes you can put the string `[ci skip]` in your commit message and it won't trigger a Travis run.

One consideration for docs: what kinds docs do we want to support building? I would probably only use the html docs, but if you have a case for other formats, that's fine.

Once we get the docs looking nice, I'll host the html docs on a gh-pages branch for easy access.

terrytangyuan:
I see. Will do. Yep gh-pages is exactly what I have in mind!

terrytangyuan:
@perimosocordiae Want to merge this first and host it? We can finish the docs later and I might need your help on this too since you are more familiar with the algorithms implemented.

perimosocordiae:
Merged. Thanks for getting started on this!

PR #14, opened 2015-11-07
Adding Error Message Rather Than Backtrace Upon Python 3 Import

seewalker:
Modified metric_learn/\__init\__.py to fail with more informative error when importing metric_learn is attempted with python 3.*

Finding out that I should be using python2 rather than python3 with metric-learn took me about 2 minutes, and with this update somebody in that position could be immediately told that that's the issue.

When running `import metric_learn` with anaconda python 3.4.3, the error message I added does in fact get printed on stderr and the program exits. When running `import metric_learn` with anaconda python 2.7.10, the error condition is not satisfied and the import succeeds.

### DIFF
--- a/metric_learn/__init__.py
+++ b/metric_learn/__init__.py
@@ -1,3 +1,7 @@
+import sys
+if sys.version_info >= (3, 0):
+    sys.stderr.write("metric-learn works with python 2.6+, but not python 3.*. Exiting.")
+    sys.exit(1)
 from itml import ITML
 from lmnn import LMNN
 from lsml import LSML


perimosocordiae:
Thanks for the PR! This gave me the push to finally just add Python 3 support, which I committed in 63372d0c710e44d5366e818ec16de5ebbc0d6651.

As you [can see here](https://travis-ci.org/all-umass/metric-learn/builds/89868560), there are still a few kinks to work out, but I'm hoping to have everything working cross-version soon.

perimosocordiae:
Python 3 problems have been fixed, so I'm making a new release. Let me know if you have any other issues.

seewalker:
Great thanks, I was hoping to use it with python 3.

PR #18, opened 2016-05-16
Added get_params, set_params and params moved to dictionary

svecon:
I wanted to use the metric-learners in a scikit Pipeline and GridSearchCV. GridSeachCV uses get_params and set_params and so I implemented both methods on the BaseMetricLearner.

Consequently I had to move all learner parameters to dictionaries so that these two methods would get to their params.


perimosocordiae:
Looks good, thanks!

PR #19, opened 2016-05-27
Fit constraints

svecon:
I needed to use the Metric Learners inside scikit Pipeline, which only calls .fit(X,y).

Therefore I renamed original .fit() to .fit_constraints(), added all of its parameters to constructor and created new .fit(), which uses the parameters from constructor to build constraints and then call .fit_constraints().

I understand this is a breaking change. If there is a better approach I want to hear it :)

This only affects methods which need special constraints: ITML, SDML, LSML, RCA

### DIFF
--- a/metric_learn/__init__.py
+++ b/metric_learn/__init__.py
@@ -1,9 +1,14 @@
 from __future__ import absolute_import
 
 from .itml import ITML
+from .itml import ITML_Supervised
 from .lmnn import LMNN
 from .lsml import LSML
+from .lsml import LSML_Supervised
 from .sdml import SDML
+from .sdml import SDML_Supervised
 from .nca import NCA
 from .lfda import LFDA
 from .rca import RCA
+from .rca import RCA_Supervised
+from .constraints import Constraints
--- a/metric_learn/__init__.py
+++ b/metric_learn/__init__.py
@@ -1,14 +1,10 @@
 from __future__ import absolute_import
 
-from .itml import ITML
-from .itml import ITML_Supervised
+from .itml import ITML, ITML_Supervised
 from .lmnn import LMNN
-from .lsml import LSML
-from .lsml import LSML_Supervised
-from .sdml import SDML
-from .sdml import SDML_Supervised
+from .lsml import LSML, LSML_Supervised
+from .sdml import SDML, SDML_Supervised
 from .nca import NCA
 from .lfda import LFDA
-from .rca import RCA
-from .rca import RCA_Supervised
-from .constraints import Constraints
+from .rca import RCA, RCA_Supervised
+from .constraints import adjacencyMatrix, positiveNegativePairs, relativeQuadruplets, chunks
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -1,0 +1,61 @@
+import random
+import numpy as np
+from six.moves import xrange
+
+class Constraints(object):
+
+  @staticmethod
+  def adjacencyMatrix(labels, num_points, num_constraints):
+    a, c = np.random.randint(len(labels), size=(2,num_constraints))
+    b, d = np.empty((2, num_constraints), dtype=int)
+    for i,(al,cl) in enumerate(zip(labels[a],labels[c])):
+      b[i] = random.choice(np.nonzero(labels == al)[0])
+      d[i] = random.choice(np.nonzero(labels != cl)[0])
+    W = np.zeros((num_points,num_points))
+    W[a,b] = 1
+    W[c,d] = -1
+    # make W symmetric
+    W[b,a] = 1
+    W[d,c] = -1
+    return W
+
+  @staticmethod
+  def positiveNegativePairs(labels, num_points, num_constraints):
+    ac,bd = np.random.randint(num_points, size=(2,num_constraints))
+    pos = labels[ac] == labels[bd]
+    a,c = ac[pos], ac[~pos]
+    b,d = bd[pos], bd[~pos]
+    return a,b,c,d
+
+  @staticmethod
+  def relativeQuadruplets(labels, num_constraints):
+    C = np.empty((num_constraints,4), dtype=int)
+    a, c = np.random.randint(len(labels), size=(2,num_constraints))
+    for i,(al,cl) in enumerate(zip(labels[a],labels[c])):
+      C[i,1] = random.choice(np.nonzero(labels == al)[0])
+      C[i,3] = random.choice(np.nonzero(labels != cl)[0])
+    C[:,0] = a
+    C[:,2] = c
+    return C
+
+  @staticmethod
+  def chunks(Y, num_chunks=100, chunk_size=2, seed=None):
+    random.seed(seed)
+    chunks = -np.ones_like(Y, dtype=int)
+    uniq, lookup = np.unique(Y, return_inverse=True)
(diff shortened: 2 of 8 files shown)


perimosocordiae:
Thanks for the PR! This is a tricky situation, partly due to the breaking change, but also because it brings up a good question about how people are expected to use this API.

From my perspective, the main use-cases for metric learning are unsupervised/semisupervised learning. In those settings, you want to use all of the constraint information available, because it's likely pretty hard to get.

The existing `prepare_constraints` functions are mostly there for testing purposes. After all, if you have full label information for all your data, why would you select some random subset of it to convert to constraints?

Even so, I agree that there should be a better / more compatible API for use in the greater scikit-learn ecosystem. The fact that most of the metric learners use different representations of their constraint information is a big pain, and the existing way of generating random constraints isn't useful for real world learning.

In light of all this, I think it makes sense to:

- scrap the `prepare_constraints` functions as they currently exist, and move their sampling functionality to the test suite.
- create a `ConstraintTransformer` class for each kind of constraint, which should be sklearn Pipeline-compatible and do the work of converting partial label vectors to the appropriate format.
- look for places where the `fit()` methods are expecting parameters that should be moved to the constructor, and fix them.

Let me know what you think.

svecon:
Problem with ConstraintTransformer classes is, that .transform() in scikit only transforms X, not y. And it would also not be a solution for me, because my Pipeline currently looks like this (notice the kNN after the LMNN - I need original labels after the metric learner):
```
        Pipeline([
            ('imputer', defaultImputer),
            ('standardizer', defaultStandardizer),
            ('lmnn', LMNN(k=3, min_iter=50, ...)),
            ('knn', KNeighborsClassifier()),
        ]),
```
The only solution which comes to my mind for this scenario is to create `ConstraintWrapper` classes for each type of constraint and they would expect the metric-transformer in their constructor.
This would separate the concerns (original metric learner classes would expect only the constraints, not labels) and it would allow me to use my current pipelines.

What do you think?

svecon:
I thinking about this some more and even started implementing a skeleton. However I came to a conclusion that it adds up too much unnecessary complexity. Having two different methods (.fit(), .fit_with_constraints()) is the simplest solution and I think it works well in most scenarios.

.fit_with_constraints() works exactly as it used to
.fit() follows scikit-learn conventions and essentially is only a helper function, which calls original .fit_with_constraints()

I think it makes more sense than having separate classes for different constraints.

svecon:
I have reverted the classes to their original functionality (where `.fit()` takes constraints) and created new classes `XXX_Supervised` which are sub-classes of the original classes, and use supervised `.fit(X,y)` in which the constraints are created.
I also moved generating constraints to separate class `Constraints`.
Does this make sense?

perimosocordiae:
I really like the new direction you've taken, thank you! I think this fits with the overall design of the package much better, and supports your use-case without too much friction.

I have a number of small nitpicks, which I'll address inline, but I'm +1 for this idea.

metric_learn/__init__.py
```diff
@@ -1,9 +1,14 @@
 from __future__ import absolute_import
 
 from .itml import ITML
+from .itml import ITML_Supervised
```
perimosocordiae: Here and below, I'd put these into one import statement:

    from .itml import ITML, ITML_Supervised

metric_learn/constraints.py
```diff
@@ -0,0 +1,63 @@
+""" Helper class that can generate different types of constraints from supervised data labels."""
+
+import random
+import numpy as np
+from six.moves import xrange
+
+class Constraints(object):
```
perimosocordiae: A class containing only static methods isn't all that useful. Either:

 1. Ditch the class and access the functions from the module namespace directly
 2. Make a stateful constraint-manager class, which handles some internal representation of constraints and has methods for emitting them in various formats. This class would then have a single method for initializing from supervised labels.

Option 2 would be ideal, but I'm not sure how feasible it would be. It might also be outside the scope of this PR, and could be added later.

metric_learn/constraints.py
```diff
@@ -0,0 +1,63 @@
...
+    a, c = np.random.randint(len(labels), size=(2,num_constraints))
+    for i,(al,cl) in enumerate(zip(labels[a],labels[c])):
+      C[i,1] = random.choice(np.nonzero(labels == al)[0])
+      C[i,3] = random.choice(np.nonzero(labels != cl)[0])
+    C[:,0] = a
+    C[:,2] = c
+    return C
+
+  @staticmethod
+  def chunks(Y, num_chunks=100, chunk_size=2, seed=None):
```
perimosocordiae: This function takes a seed, others just use the default random state. We should (eventually) move to the numpy-style of using `RandomState` objects, which would default to the global state but also support user-defined seeds, etc.

This can wait until after the PR merges, though.

metric_learn/itml.py
```diff
@@ -138,3 +135,47 @@ def _vector_norm(X):
...
+    ----------
+    gamma : float, optional
+        value for slack variables
+    max_iters : int, optional
+    convergence_threshold : float, optional
+    num_constraints: int, needed for .fit()
+    verbose : bool, optional
+        if True, prints information while learning
+    """
+    self.params = {
```
perimosocordiae: I'd prefer to call the base constructor here:

    ITML.__init__(self, gamma=gamma, max_iters=max_iters, convergence_threshold=convergence_threshold,
                  bounds=bounds, A0=A0, verbose=verbose)
    self.params['num_constraints'] = num_constraints

metric_learn/itml.py
```diff
@@ -138,3 +135,47 @@ def _vector_norm(X):
...
+    X : (n x d) data matrix
+        each row corresponds to a single instance
+    labels : (n) data labels
+    """
+    num_constraints = self.params['num_constraints']
+    if num_constraints is None:
+      num_constraints = 20*(len(set(labels)))**2 # 20* number of classes**2
+
+    C = Constraints.positiveNegativePairs(labels, X.shape[0], num_constraints)
+    return super(ITML_Supervised,self).fit(X, C, bounds=self.params['bounds'], A0=self.params['A0'])
```
perimosocordiae: Since there's only one base class, I typically would do:

    return ITML.fit(self, X, C, bounds=self.params['bounds'], A0=self.params['A0'])

metric_learn/itml.py
```diff
@@ -138,3 +135,47 @@ def _vector_norm(X):
...
+
+    Parameters
+    ----------
+    X : (n x d) data matrix
+        each row corresponds to a single instance
+    labels : (n) data labels
+    """
+    num_constraints = self.params['num_constraints']
+    if num_constraints is None:
+      num_constraints = 20*(len(set(labels)))**2 # 20* number of classes**2
```
perimosocordiae: Calling `set()` on a big numpy array is going to be slow. Instead, I'd write:

    num_classes = len(np.unique(labels))
    num_constraints = 20 * num_classes**2

metric_learn/lsml.py
```diff
@@ -62,6 +66,7 @@ def fit(self, X, constraints, weights=None, prior=None, verbose=False):
     verbose : bool, optional
```
perimosocordiae: This docstring needs to be updated.

svecon:
I have fixed the problems you pointed out and marked the code with #TODO where you suggested bigger changes.

Anything else to do here before merge?

perimosocordiae:
Looks good now, merging. Thanks!

PR #21, opened 2016-06-17
Simple Covariance metric-learner

svecon:
Some papers mention using inverse of covariance matrix as a simple "metric-learning" method.
I thought it would a good idea to have this explicitly in a class.

### DIFF
--- a/metric_learn/__init__.py
+++ b/metric_learn/__init__.py
@@ -8,3 +8,4 @@ from .nca import NCA
 from .lfda import LFDA
 from .rca import RCA, RCA_Supervised
 from .constraints import adjacencyMatrix, positiveNegativePairs, relativeQuadruplets, chunks
+from .covariance import Covariance
--- a/metric_learn/covariance.py
+++ b/metric_learn/covariance.py
@@ -1,0 +1,23 @@
+"""
+Simple Covariance metric (learner)
+"""
+
+from __future__ import absolute_import
+import numpy as np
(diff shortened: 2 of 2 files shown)


perimosocordiae:
I might call this a Mahalanobis metric: https://www.cs.princeton.edu/courses/archive/fall08/cos436/Duda/PR_Mahal/M_metric.htm

A little more explanation in the module docstring would be nice, ideally explaining that this is just a simple baseline method.

I agree that this is a useful addition.

svecon:
I disagree with the name:
According to the original paper http://www.unt.edu/rss/class/Jon/MiscDocs/1936_Mahalanobis.pdf (1936) they would indeed call this a Mahalanobis distance.
However this name has been generalized for any matrix (positive semi-definite). Therefore all currently implemented algorithms in this package are a variation of Mahalanobis distance.
Sources:
Xing (2002) http://www.cs.cmu.edu/~epxing/papers/Old_papers/xing_nips02_metric.pdf
Bellet's survey (2014) https://arxiv.org/pdf/1306.6709.pdf (page 8)

perimosocordiae:
Fair enough. Merging the other PR caused a merge conflict, so you'll need to rebase this change against master before I can merge it.

svecon:
Updated docstring and rebased to master. Should be ready to merge.

perimosocordiae:
Thanks, merged.

PR #27, opened 2016-09-08
[MRG] Metric Learning Tutorial Notebook

bhargavvader:
@perimosocordiae , thought this might be handy in visualising what the algorithms are doing. It's very useful for me, so I thought I'd clean up the code I have and make a notebook.

Let me know if this is ok, and whether you might want more things added to this.
Once #26 is completed, we can change this later to reflect `fit_transform` as well.

Edit: also wanted to know where the right place for this would be, whether in the `examples` folder or in  the `docs` folder.


perimosocordiae:
Very cool! I think this would fit nicely in `examples/`.

Some of the code in the notebook can be cleaned up, and it might be better to use an alternate dataset to really show how the methods differ. Some of the papers have illustrative examples that we might be able to use.

The one big omission from the notebook is how you might specify constraints manually for each method. (Rather than just using the _Supervised variants.) This might be a bit harder to show clearly, but I think it's worthwhile.

bhargavvader:
Could you mention which part exactly could be cleaned up?
I would ideally want to use a different data set too, but thought of using just one for all the examples would be easier in making a comparison - do you have any ideas of other data sets to use, in particular?

I like the idea of specifying constraints manually, I'll add what I can regarding that to the notebook.

perimosocordiae:
Regarding code cleanup, you keep reassigning the `X` variable, which makes things confusing to read. Maybe have each method assign to a separate variable (i.e., `x_lmnn` or `x_itml`) when calling `.transform()`?

The docs aren't very clear about manual constraint specification yet, but the docstrings for each method should give you a starting point.  For example, `ITML`'s constraints are a "4-tuple of arrays (a,b,c,d)" which are "indices into X, such that d(X[a],X[b]) < d(X[c],X[d])". You can also look at the constraints class to find examples of making each kind of constraint.

bhargavvader:
TODO:

- clean up notebook with regard to old comments
- add example of `fit_transform` as well
- briefly describe each algorithm as well as link to paper
- give example of manual constraints
- maybe add nice motivating example to use any one also (e.g clustering)

Just some notes for myself. I'll get to this soon :)

perimosocordiae:
This is looking good! A few lines of description about each method will be nice, as well as a mention that (for the most part) we follow scikit-learn conventions.

It might also be good to explain why running the same learner multiple times might not always give the same results. The current API for the supervised methods hides all of the randomness, which might be confusing to a new user.

bhargavvader:
Added descriptions and links for all the algorithms, and mentioned about scikit-learn and using random seeds.

Only have to do the manual constraints bit, just going to spend some time coming up with a good example for it.

bhargavvader:
@perimosocordiae, apart from ITML and LSML all the methods use labels as a way to supervise the learning, right?

So the example of manual constraints, it's only for them, right?

bhargavvader:
@perimosocordiae I'm done with the tutorial! 
If you think it needs any other additions let me know; but from my side, right now, you can merge.

perimosocordiae:
Great, thanks! This will be a useful resource for people new to the project.

terrytangyuan:
Awesome! It would be better if we include the associated .py file as part of the test suite.

perimosocordiae:
We could have an automated test run [`jupyter nbconvert --to script`](http://nbconvert.readthedocs.io/en/latest/usage.html#executable-script) on it, then make sure it runs without errors. I think that could be added to the TravisCI tests without too much difficulty.

bhargavvader:
I haven't made an associated .py file, but could if needed; but I think adding it to the TravisCI tests would be better, right?

terrytangyuan:
CJ's idea sounds good without the need to maintain a separate py file. Yeah it would be better to continuously test it, which is critical for future users.

Issue #13, opened 2015-10-14
Algorithm wish list


perimosocordiae:
Methods that we haven't implemented yet, but would like to:

 - [ ] [Probablistic Global Distance Metric Learning (PGDM)](http://ai.stanford.edu/~ang/papers/nips02-metric.pdf)
  * [reference implementation](http://www.cs.cmu.edu/~epxing/papers/Old_papers/code_Metric_online.tar.gz)
 - [ ] [Discriminative Component Analysis (DCA)](http://ieeexplore.ieee.org/iel5/10925/34374/01641007.pdf)
  * [reference implementation](http://www.cs.cmu.edu/~liuy/dca.zip)
 - [ ] [Localized Distance Metric Learning (LDM)](http://www.cse.msu.edu/~yangliu1/aaai2006-distance-v7.pdf)
  * [reference implementation](http://www.cs.cmu.edu/~liuy/ldm_scripts_2.zip)
 - [ ] [DistBoost](http://www.cs.huji.ac.il/~tomboy/papers/RCA_ICML03.pdf) and [KernelBoost](http://www.cs.huji.ac.il/~tomboy/papers/KernelBoost_ICML06.pdf)
  * [reference implementation](http://www.cs.huji.ac.il/~tomboy/code/DistBoost.zip)
 - [ ] [Active Distance Metric Learning](http://www.cse.msu.edu/~yangliu1/uai2007_bayesian.pdf)
  * [reference implementation](http://www.cs.cmu.edu/~liuy/script_bayesian_dist.zip)
 - [ ] [Metric Learning for Kernel Regression (MLKR)](http://www.cs.cornell.edu/~kilian/papers/weinberger07a.pdf)
  * [reference implementation](http://www.cse.wustl.edu/~kilian/code/files/MLKR0.9.zip)

bhargavvader:
Hello, any tips or further documentation on how to go about doing this? And which one of these would be highest priority?

perimosocordiae:
I don't personally have a priority, but if you want to rank them you could look at their citation counts on Google Scholar. That should give you an idea of what's more popular in the literature. I would start with one that seems easy to implement, though.

[...]

PR #28, opened 2016-09-09
Added MLKR algorithm

dsquareindia:
Implements MLKR (sorry I've named the branch mklr :P) algorithm mentioned in #13. @perimosocordiae could you please review once? Sorry for the loops! I'll try to make it more vectorized. I'm also checking if any mathematical mistakes have inadvertently crept in. Also, what could be a good test for this? Should it just be on the lines of `testRCA`? Sorry again for the poor quality of the code. This will surely go through many revisions.

### DIFF
--- a/metric_learn/mlkr.py
+++ b/metric_learn/mlkr.py
@@ -1,0 +1,124 @@
+"""
+Metric Learning for Kernel Regression (MLKR), Weinberger et al.,
+
+MLKR is an algorithm for supervised metric learning, which learns a distance
+function by directly minimising the leave-one-out regression error. This
+algorithm can also be viewed as a supervised variation of PCA and can be used
+for dimensionality reduction and high dimensional data visualization.
+"""
+from __future__ import division
+import numpy as np
+from six.moves import xrange
+
+from .base_metric import BaseMetricLearner
+
+class MLKR(BaseMetricLearner):
+    """Metric Learning for Kernel Regression (MLKR)"""
+    def __init__(self, A=None, epsilon=0.01):
+        """
+        MLKR initialization
+
+        Parameters
+        ----------
+        A: Initialization of matrix A. Defaults to the identity matrix.
+        epsilon: Step size for gradient descent
+        """
+        self.params = {
+            "A": A,
+            "epsilon": epsilon
+        }
+
+    def _process_inputs(self, X, y):
+        if X.ndim == 1:
+            X = X[:, np.newaxis]
+        if y.ndim == 1:
+            y == y[:, np.newaxis]
+        self.X = X
+        n, d = X.shape
+        assert y.shape[0] == n
+        return y, n, d
+
+    def fit(self, X, y):
+        """
+        Fit MLKR model
+
+        Parameters:
+        ----------
+        X : (n x d) array of samples
+        y : (n) data labels
+
+        Returns:
+        -------
+        self: Instance of self
+        """
+        y, n, d = self._process_inputs(X, y)
+        alpha = 0.0001  # Stopping criterion
+        if self.params['A'] is None:
+            A = np.identity(d)  # Initialize A as eye matrix
+        else:
+            A = self.params['A']
+            assert A.shape == (d, d)
+        cost = np.Inf
+        # Gradient descent procedure
+        while cost > alpha:
+            K = self._computeK(X, A, n)
+            yhat = self._computeyhat(y, K)
+            sum_i = 0
+            for i in xrange(n):
+                sum_j = 0
+                for j in xrange(n):
+                    sum_j += (yhat[j] - y[j]) * K[i][j] * \
+                             (X[i, :] - X[j, :])[:, np.newaxis].dot \
+                                 ((X[i, :] - X[j, :])[:, np.newaxis].T)
+                sum_i += (yhat[i] - y[i]) * sum_j
+            gradient = 4 * A.dot(sum_i)
+            A -= self.params['epsilon'] * gradient
+            cost = np.sum(np.square(yhat - y))
+        self._transformer = A
+        return self
+
+    def _computeK(self, X, A, n):
+        """
+        Internal helper function to compute K matrix.
+
+        Parameters:
+        ----------
+        X: (n x d) array of samples
+        A: (d x d) 'A' matrix
+        n: number of rows in X
+
+        Returns:
+        -------
+        K: (n x n) K matrix where Kij = exp(-distance(x_i, x_j)) where
+           distance is defined as squared L2 norm of (x_i - x_j)
+        """
+        dist_mat = np.zeros(shape=(n, n))
+        for i in xrange(n):
+            for j in xrange(n):
+                if i == j:
+                    dist = 0
+                else:
+                    dist = np.sum(np.square(A.dot((X[i, :] - X[j, :]))))
+                dist_mat[i, j] = dist
+        return np.exp(-dist_mat)
+
+    def _computeyhat(self, y, K):
+        """
+        Internal helper function to compute yhat matrix.
+
+        Parameters:
+        ----------
+        y: (n) data labels
+        K: (n x n) K matrix
+
+        Returns:
+        -------
+        yhat: (n x 1) yhat matrix
+        """
+        numerator = K.dot(y)
+        denominator = np.sum(K, 1)[:, np.newaxis]
+        yhat = numerator / denominator
+        return yhat
+
+    def transformer(self):
+        return self._transformer
--- a/metric_learn/mlkr.py
+++ b/metric_learn/mlkr.py
@@ -9,33 +9,39 @@ for dimensionality reduction and high dimensional data visualization.
 from __future__ import division
 import numpy as np
 from six.moves import xrange
+from scipy.spatial.distance import pdist, squareform
 
 from .base_metric import BaseMetricLearner
 
 class MLKR(BaseMetricLearner):
     """Metric Learning for Kernel Regression (MLKR)"""
-    def __init__(self, A=None, epsilon=0.01):
+    def __init__(self, A0=None, epsilon=0.01, alpha=0.0001):
         """
         MLKR initialization
 
         Parameters
         ----------
-        A: Initialization of matrix A. Defaults to the identity matrix.
-        epsilon: Step size for gradient descent
+        A0: Initialization of matrix A. Defaults to the identity matrix.
+        epsilon: Step size for gradient descent.
+        alpha: Stopping criterion for loss function in gradient descent.
         """
         self.params = {
-            "A": A,
-            "epsilon": epsilon
+            "A0": A0,
+            "epsilon": epsilon,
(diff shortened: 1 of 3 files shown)


metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+        A: Initialization of matrix A. Defaults to the identity matrix.
+        epsilon: Step size for gradient descent
+        """
+        self.params = {
+            "A": A,
+            "epsilon": epsilon
+        }
+
+    def _process_inputs(self, X, y):
+        if X.ndim == 1:
```
perimosocordiae: It's usually a good idea to convert inputs to numpy arrays here, just in case the user provided a plain Python list, or some other sequence.

```python
self.X = np.array(X, copy=False)
y = np.array(y, copy=False)
```

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+            X = X[:, np.newaxis]
+        if y.ndim == 1:
+            y == y[:, np.newaxis]
```
perimosocordiae: typo here: `==` should be `=`

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+        self.X = X
+        n, d = X.shape
+        assert y.shape[0] == n
```
perimosocordiae: In general, prefer to raise an actual exception instead of using `assert`. (I know that some of the other code isn't very good about following this rule yet, sorry!)

For example:
```python
if y.shape[0] != n:
    raise ValueError('data and label lengths mismatch: %d != %d' % (n, y.shape[0]))
```

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+        ----------
+        X : (n x d) array of samples
+        y : (n) data labels
+
+        Returns:
+        -------
+        self: Instance of self
+        """
+        y, n, d = self._process_inputs(X, y)
+        alpha = 0.0001  # Stopping criterion
```
perimosocordiae: This should be a user-settable parameter (in `self.params`).

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+        while cost > alpha:
+            K = self._computeK(X, A, n)
+            yhat = self._computeyhat(y, K)
+            sum_i = 0
+            for i in xrange(n):
+                sum_j = 0
+                for j in xrange(n):
+                    sum_j += (yhat[j] - y[j]) * K[i][j] * \
+                             (X[i, :] - X[j, :])[:, np.newaxis].dot \
+                                 ((X[i, :] - X[j, :])[:, np.newaxis].T)
```
perimosocordiae: This might end up getting vectorized, but for now I'd recommend writing out each major subexpression on a separate line (assigned to a separate variable) for readability.

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+        cost = np.Inf
+        # Gradient descent procedure
+        while cost > alpha:
+            K = self._computeK(X, A, n)
+            yhat = self._computeyhat(y, K)
+            sum_i = 0
+            for i in xrange(n):
+                sum_j = 0
+                for j in xrange(n):
+                    sum_j += (yhat[j] - y[j]) * K[i][j] * \
```
perimosocordiae: Should be `K[i,j]` here?

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+                sum_i += (yhat[i] - y[i]) * sum_j
+            gradient = 4 * A.dot(sum_i)
+            A -= self.params['epsilon'] * gradient
+            cost = np.sum(np.square(yhat - y))
+        self._transformer = A
+        return self
+
+    def _computeK(self, X, A, n):
+        """
+        Internal helper function to compute K matrix.
```
perimosocordiae: I suspect that this function could be replaced with a call to [`scipy.spatial.pdist`](http://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.pdist.html#scipy.spatial.distance.pdist) with the `'mahalanobis'` metric.

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+
+class MLKR(BaseMetricLearner):
+    """Metric Learning for Kernel Regression (MLKR)"""
+    def __init__(self, A=None, epsilon=0.01):
+        """
+        MLKR initialization
+
+        Parameters
+        ----------
+        A: Initialization of matrix A. Defaults to the identity matrix.
```
perimosocordiae: I might call this `A0`, to distinguish it from the learned matrix `A`.

perimosocordiae:
A good starting point is to add a test to the existing ones based on the Iris dataset, just to exercise the code and make sure that it produces reasonable-looking results.

It might also be useful to try to replicate the paper's result on the "twin-peaks" synthetic dataset. (See figure 2 in the PDF.)

dsquareindia:
@perimosocordiae I've addressed the initial comments. I'll try to reproduce the twin-peaks dataset too. Any notes on how I should be going about it? We'll look into how we can vectorize the gradient descent further since the test takes an awful lot of time. Can we switch to SGD?

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+        Returns:
+        -------
+        self: Instance of self
+        """
+        y, n, d = self._process_inputs(X, y)
+        if self.params['A0'] is None:
+            A = np.identity(d)  # Initialize A as eye matrix
+        else:
+            A = self.params['A0']
+            assert A.shape == (d, d)
```
perimosocordiae: This `assert` should be a `ValueError`

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+        cost = np.Inf
+        # Gradient descent procedure
+        while cost > self.params['alpha']:
```
perimosocordiae: For efficiency, it's better to get `alpha` from the params dictionary once, rather than every iteration:
```python
alpha = self.params['alpha']
while cost > alpha:
    ...
```

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+            for i in xrange(n):
+                sum_j = 0
+                for j in xrange(n):
+                    diffK = (yhat[j] - y[j]) * K[i, j]
+                    x_ij = (X[i, :] - X[j, :])[:, np.newaxis]
+                    x_ijT = x_ij.T
+                    sum_j += diffK * x_ij.dot(x_ijT)
+                sum_i += (yhat[i] - y[i]) * sum_j
+            gradient = 4 * A.dot(sum_i)
+            A -= self.params['epsilon'] * gradient
```
perimosocordiae: Same thing for `epsilon` as above with `alpha`.

metric_learn/mlkr.py
```diff
@@ -0,0 +1,124 @@
...
+
+        Returns:
+        -------
+        K: (n x n) K matrix where Kij = exp(-distance(x_i, x_j)) where
+           distance is defined as squared L2 norm of (x_i - x_j)
+        """
+        dist_mat = pdist(X, metric='mahalanobis', VI=A.T.dot(A))
+        dist_mat = np.square(dist_mat)
+        dist_mat = squareform(dist_mat)
+        return np.exp(-dist_mat)
```
perimosocordiae: It's faster to do all the math on the distances before calling `squareform`:
```python
dist = pdist(X, metric='mahalanobis', VI=A.T.dot(A))
return squareform(np.exp(-dist**2))
```

dsquareindia: I just had a doubt about how to apply the rotation matrix in the final step after generating X. Is it necessary?

dsquareindia:
I just had a doubt about how to apply the rotation matrix in the final step after generating X. Is it necessary?

perimosocordiae:
Oh right, I forgot about that part! That's the most important step:

```python
rotation = scipy.linalg.orth(np.random.normal(size=(10,10)))
return X.dot(rotation)
```

dsquareindia:
Thanks! Btw what do you think about adding an "SGD mode"? Running on this dataset is taking forever!

perimosocordiae:
It might be a good idea, but let's make sure the non-stochastic version works first (even if it's really slow). SGD is much more difficult to debug!

dsquareindia:
On twin-peaks, with the default values for alpha and epsilon, the cost is always increasing. There should be an internal check and adjustment for this right? How do we fix this? The gradient and cost computation seems to be correct. Would it be better to use `scipy.optimize` routines for minimizing?

dsquareindia:
ping @perimosocordiae :)

perimosocordiae:
I was away last week, so I'm just getting to look at this now. If possible, it would be great to use `scipy.optimize` here. That would give us a lot of support for changing convergence conditions and iteration/evaluation limits.

dsquareindia:
Alright awesome! I'll get to it in a while.

dsquareindia:
There seems to be some problem somewhere in my code. My cost keeps rising on the twin peaks dataset. Here's what my cost looks like on consecutive iterations on twin_peaks(100):
```
X, y = twin_peaks(100)
mlkr = MLKR()
mlkr.fit(X, y)
```
```
inf
18.6375498396
18.638362056
18.6391722602
18.6399804524
18.6407866332
18.641590803
18.6423929623
18.6431931116
18.6439912514
18.6447873824
18.6455815052
18.6463736204
18.6471637286
18.6479518307
18.6487379272
18.6495220191
18.6503041071
18.651084192
18.6518622747
18.6526383561
18.6534124371
18.6541845186
18.6549546017
18.6557226873
18.6564887764
18.6572528701
18.6580149696
18.6587750759
18.6595331901
18.6602893135
18.6610434473
18.6617955926
18.6625457507
18.6632939229
```
I've been trying to debug it but not reaching anywhere. My gradient computation also seems alright :/
@perimosocordiae could you please have a look? Thanks!

perimosocordiae:
I'm looking into this now.

dsquareindia:
I'm also trying switching to maximum number of line searches rather than keeping an `epsilon`. Atleast with that we'll be sure of getting out of the loop. This is what is done in the octave code. They have an option of number of line searches or number of function evaluations but not a minimum loss value.

perimosocordiae:
I'm merging this now so that I can apply some fixups that I've made locally.

perimosocordiae:
And thanks @dsquareindia for making this happen!

dsquareindia:
Oh I was experimenting with the line searches on twin peaks and getting pretty good results :P
No problem and thanks a lot for the help! Once you merge in MLKR with your patches, I'll put in the PR for twin peaks (if you're not doing it already) using the final MLKR. Are you working on efficient optimisation too?

perimosocordiae:
I've committed all of my changes now, so feel free to work with the current code.

Issue #33, opened 2016-09-15
Enable user-given random state for constraint generation


perimosocordiae:
Each method in the `Constraints` class which generates random numbers should have a `random_state=None` keyword argument, which will make constraint generation reproducible.

This will be useful for the test cases added in #26.

bhargavvader:
Sounds good, I'll do this.

bhargavvader:
This can be closed too!

PR #35, opened 2016-09-19
[MRG] Added random_states

bhargavvader:
Added options to pass a `random_seed` for all the algorithms which use it.
Addresses #33, and will help the tests for #26.

### DIFF
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -63,7 +63,6 @@ class Constraints(object):
     """
     the random state object to be passed must be a numpy random seed
     """
-    state = random_state
     chunks = -np.ones_like(self.known_label_idx, dtype=int)
     uniq, lookup = np.unique(self.known_labels, return_inverse=True)
     all_inds = [set(np.where(lookup==c)[0]) for c in xrange(len(uniq))]
@@ -88,7 +87,6 @@ class Constraints(object):
     """
     the random state object to be passed must be a numpy random seed
     """
-    state = random_state
     n = len(all_labels)
     num_ignored = max(0, n - num_preserved)
     idx = random_state.randint(n, size=num_ignored)
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -84,7 +84,7 @@ class Constraints(object):
     return chunks
 
   @staticmethod
-  def random_subset(all_labels, num_preserved=np.infm, random_state=None):
+  def random_subset(all_labels, num_preserved=np.inf, random_state=None):
     """
     the random state object to be passed must be a numpy random seed
     """
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -19,8 +19,8 @@ class Constraints(object):
     self.known_label_idx, = np.where(partial_labels >= 0)
     self.known_labels = partial_labels[self.known_label_idx]
 
-  def adjacency_matrix(self, num_constraints):
-    a, b, c, d = self.positive_negative_pairs(num_constraints)
+  def adjacency_matrix(self, num_constraints, random_state=np.random):
+    a, b, c, d = self.positive_negative_pairs(num_constraints, random_state=random_state)
     row = np.concatenate((a, c))
     col = np.concatenate((b, d))
     data = np.ones_like(row, dtype=int)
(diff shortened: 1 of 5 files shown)


bhargavvader:
ping @perimosocordiae . 
A review on this would mean I can wrap up my other PRs as well. :)

bhargavvader:
ping @perimosocordiae.
Could you please review this?

metric_learn/constraints.py
```diff
@@ -59,7 +59,11 @@ def _pairs(self, num_constraints, same_label=True, max_iter=10):
     ab = np.array(list(ab)[:num_constraints], dtype=int)
     return self.known_label_idx[ab.T]
 
-  def chunks(self, num_chunks=100, chunk_size=2):
+  def chunks(self, num_chunks=100, chunk_size=2, random_state=None):
+    """
+    the random state object to be passed must be a random.seed() object and not a numpy random seed
+    """
+    state = random_state
```
perimosocordiae: The `random_state` object isn't used in this function, so I'd drop it entirely.

metric_learn/constraints.py
```diff
@@ -80,7 +84,11 @@ def chunks(self, num_chunks=100, chunk_size=2):
...
   @staticmethod
-  def random_subset(all_labels, num_preserved=np.inf):
+  def random_subset(all_labels, num_preserved=np.inf, random_state=None):
+    """
+    the random state object to be passed must be a numpy random seed
+    """
+    state = random_state
     n = len(all_labels)
     num_ignored = max(0, n - num_preserved)
     idx = np.random.randint(n, size=num_ignored)
```
perimosocordiae: The point of passing in a random state is to generate the random numbers we need from it. So in this case, it would look like:

```python
def random_subset(all_labels, num_preserved=np.inf, random_state=np.random):
   ...
   idx = random_state.randint(n, size=num_ignored)
   ...
```

We can use the `np.random` module as the default value, which means it'll call `np.random.randint` by default.
This wouldn't support passing in a raw numeric seed, but that's ok by me: users can just pass in `np.random.RandomState(1234)` if they want to use seed `1234`.

metric_learn/constraints.py
bhargavvader: Isn't it used because random is used in `random.randint` and `random.sample`?
I did not mention `numpy` on purpose, because while `numpy` randomstate isn't being used, a random seed still is.

metric_learn/constraints.py
bhargavvader: Okay, cool. I'll add the `np.random` as the default in all the other classes as well.

bhargavvader:
@perimosocordiae could you let me know how to handle random seeds in the `chunks` method? It uses random but not the numpy one so the seed passed is different.

Could you also have a look at the other classes to see if it is fine?

metric_learn/constraints.py
perimosocordiae: Ah, right. We should convert those calls to use the numpy random module. Then, the same technique used in the other methods will apply.

bhargavvader:
@perimosocordiae , I addressed the comments and added `np.random` as default.

metric_learn/constraints.py
```diff
@@ -59,18 +59,22 @@ def _pairs(self, num_constraints, same_label=True, max_iter=10):
...
+    the random state object to be passed must be a numpy random seed
+    """
+    state = random_state
     chunks = -np.ones_like(self.known_label_idx, dtype=int)
     uniq, lookup = np.unique(self.known_labels, return_inverse=True)
     all_inds = [set(np.where(lookup==c)[0]) for c in xrange(len(uniq))]
     idx = 0
     while idx < num_chunks and all_inds:
-      c = random.randint(0, len(all_inds)-1)
+      c = np.random.randint(0, high=len(all_inds)-1)
```
perimosocordiae: This is a good start, but we need to be using `random_state` here:

    c = random_state.randint(0, high=len(all_inds)-1)

(Same comment for all the other places where we're directly calling functions on `np.random` in this module.)

bhargavvader: Of course - I feel rather silly now. Pushing the right commit!

bhargavvader:
@perimosocordiae, again, incorporated suggestions and changes (silly, stupid oversights).

perimosocordiae:
Looking much better, thanks! The only remaining issue is the `positive_negative_pairs` method (and its helper method `_pairs`) on the Constraints object. These methods should have a `random_state` keyword argument just like the others.

bhargavvader:
Ah, I hadn't noticed it. Made the changes.

PR #36, opened 2016-09-29
adds code for safe return in case of no impostors for lmnn

### LINKED ISSUE #17: Error when impostors list is empty
Numpy throws an error in np.hstack(impostors) when impostors is empty.
What would be the recommended way to fix this?

anirudt:
Attempt to fix #17 .

I have done it for LMNN, is there a need for the others?

### DIFF
--- a/metric_learn/lmnn.py
+++ b/metric_learn/lmnn.py
@@ -207,8 +207,8 @@ class python_LMNN(_base_LMNN):
         i,j = np.unravel_index(np.unique(tmp), shape)
       impostors.append(np.vstack((in_inds[j], out_inds[i])))
     if len(impostors) == 0:
-        """No impostors detected"""
-        return []
+        # No impostors detected
+        return impostors
     return np.hstack(impostors)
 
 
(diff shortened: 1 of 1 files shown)


metric_learn/lmnn.py
```diff
@@ -203,6 +206,9 @@ def _find_impostors(self, furthest_neighbors):
         tmp = np.ravel_multi_index((i,j), shape)
         i,j = np.unravel_index(np.unique(tmp), shape)
       impostors.append(np.vstack((in_inds[j], out_inds[i])))
+    if len(impostors) == 0:
+        """No impostors detected"""
+        return []
```
perimosocordiae: No need for a docstring here. Also, you can just return `impostors` instead of returning a new empty list.

perimosocordiae:
Thanks! One small comment, but the change looks good overall. If something similar needs to be done for the other methods, we can handle those fixes separately.

metric_learn/lmnn.py
anirudt: Sure, will do.

anirudt:
Sure, @perimosocordiae . Looking at the CI Build Details, the RCA test hasn't gone through. Does it need  a look into it?

perimosocordiae:
The test failure is unrelated, though it did make me smile: `0.25054383490356519 not less than 0.25`. So close!

Thanks for your contribution!

PR #42, opened 2016-11-07
Removed random import

bhargavvader:
Removed the `random` import as we've now completely shifted to the numpy random class for any random functions.


perimosocordiae:
Thanks, merged.

PR #44, opened 2016-12-20
Nca

ab-anssi:
1) The current implementation of NCA does not work when the variance of a feature is null. Indeed, the matrix A is then initialized with some nan values (1 / max - min is undefined).
I propose to initialize A with 1 / max(max - min, EPS). This solution is similar to the one used in MLKR.

2) The current implementation of NCA does not support dimension reduction. I propose to:
     - Add a parameter num_dim to the constructor (as in MLKR)
     - Initialize A with a matrix of dimension (num_dim, d)

### DIFF
--- a/metric_learn/nca.py
+++ b/metric_learn/nca.py
@@ -9,6 +9,7 @@ from six.moves import xrange
 
 from .base_metric import BaseMetricLearner
 
+EPS = np.finfo(float).eps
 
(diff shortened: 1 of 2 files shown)


perimosocordiae:
Looks good! Could you add a test case that provides a non-default setting for `num_dims`?

perimosocordiae:
Merged. Thanks for the contribution!

PR #47, opened 2017-01-10
Added __repr__ to BaseMetricLearner

svecon:
I took the __repr__ from scikit and put it to base_metric together with _pprint.

### DIFF
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -1,4 +1,6 @@
+import numpy as np
 from numpy.linalg import inv,cholesky
+from six import iteritems
 
 
 class BaseMetricLearner(object):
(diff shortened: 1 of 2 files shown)


metric_learn/base_metric.py
```diff
@@ -95,3 +95,67 @@ def set_params(self, **kwarg):
...
+  # https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/base.py#L287
+  def __repr__(self):
+    class_name = self.__class__.__name__
+    return '%s(%s)' % (class_name, _pprint(self.get_params(deep=False),
+                                               offset=len(class_name),),)
+
+###############################################################################
+
+# _iteritems taken from:
+# https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/externals/six.py#L437
```
perimosocordiae: We already depend on `six`, so you can just do: `from six import iteritems`

metric_learn/base_metric.py
```diff
@@ -95,3 +95,67 @@ def set_params(self, **kwarg):
...
+    params : dict
+        The dictionary to pretty print
+    offset : int
+        The offset in characters to add at the begin of each line.
+    printer : callable
+        The function to convert entries to strings, typically
+        the builtin str or repr
+    """
+    # Do a multi-line justified repr:
+    import numpy as np
```
perimosocordiae: No need to import this here, just do the import at the top.

metric_learn/base_metric.py
```diff
@@ -95,3 +95,67 @@ def set_params(self, **kwarg):
     """
     self.params.update(kwarg)
     return self
+  
+  # https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/base.py#L287
+  def __repr__(self):
+    class_name = self.__class__.__name__
+    return '%s(%s)' % (class_name, _pprint(self.get_params(deep=False),
+                                               offset=len(class_name),),)
```
perimosocordiae: I would split this into two lines:
```python
params = self.get_params(deep=False)
return '%s(%s)' % (class_name, _pprint(params, offset=len(class_name)))
```

perimosocordiae:
Great, thanks!

PR #48, opened 2017-01-26
Fixes ITML performance 

arikpoz:
This pull request fixes a performance bug in the ITML (Information-Theoretic Metric Learning) code.
I've replaced the formula: A*(v*v)*A with (A*v)*(A*v), which improves the complexity from O(n^3) to O(n^2), as described in the original ITML paper.
note that one mathematical justification is in order, which is A=A' in all stages of the algorithm since A is a Mahalanobis matrix.
I've tested it and got the exact same numerical results, with the fraction of the time needed (hours vs. days).


perimosocordiae:
Looks great, thanks!

Issue #49, opened 2017-01-30
RCA - Non-invertible inner covariance matrix 


ab-anssi:
The current implementation of RCA does not check whether the inner covariance matrix is invertible. However, when this matrix is not invertible, it has some null eigen values, and so the transformation matrix `transformer_` contains Nan values. 
I provide an example illustrating this problem. [rca_non_invertible_covariance.txt](https://github.com/all-umass/metric-learn/files/738990/rca_non_invertible_covariance.txt)

The solution is to select only eigenvalues that are NOT NULL.
There is a problem when the rank of the inner covariance matrix, denoted by `r`, is smaller than `dim`, the number of dimension requested. I think a good solution is to return a matrix with `r` dimensions, and to display a warning to the user.

What do you think of this bug fix ? Can I send a pull request ?

perimosocordiae:
I like your proposed solution, and I look forward to the pull request! 👍

perimosocordiae:
Closed by #50.

PR #50, opened 2017-02-02
Handle rank-deficient inner covariance in RCA

ab-anssi:
It solves the issue #49 .

The input data is projected with a PCA if the inner covariance matrix is not full rank as advised in the paper "Learning a Mahalanobis Metric from Equivalence Constraints" (JMLR 2005).

I have added two tests that do not work with the current implementation :
  - A feature with a null variance
  - Digits dataset

I have also tested on my own datasets, and this implementation seems robust.

### DIFF
--- a/metric_learn/rca.py
+++ b/metric_learn/rca.py
@@ -103,6 +103,10 @@ class RCA(BaseMetricLearner):
 
     # The embedding dimension must be smaller than the rank of the inner covariance matrix
     dim = min(self.params['dim'], rank)
+    if dim < self.params['dim']:
+      mess  = 'WARNING : The embedding dimension must be smaller than the rank of the inner covariance matrix. '
+      mess += 'dim is set to ' + str(dim) + '.'
+      print mess
 
     # Fisher Linear Discriminant projection
     if dim < rank:
--- a/metric_learn/rca.py
+++ b/metric_learn/rca.py
@@ -8,6 +8,7 @@ Those relevant dimensions are estimated using "chunklets",
 subsets of points that are known to belong to the same class.
 
 'Learning distance functions using equivalence relations', ICML 2003
+'Learning a Mahalanobis metric from equivalence constraints', JMLR 2005
 """
 
 from __future__ import absolute_import
--- a/metric_learn/rca.py
+++ b/metric_learn/rca.py
@@ -98,7 +98,7 @@ class RCA(BaseMetricLearner):
       d = data.shape[1]
       M_pca = pca.components_
     else:
-        data -= data.mean(axis=0)
+      data -= data.mean(axis=0)
 
     chunk_mask, chunk_data = self._process_chunks(data, chunks)
     inner_cov = np.cov(chunk_data, rowvar=0, bias=1)
--- a/metric_learn/rca.py
+++ b/metric_learn/rca.py
@@ -54,12 +54,9 @@ class RCA(BaseMetricLearner):
     self.params = {
       'dim': dim,
     }
-
+  
   def transformer(self):
-    if self.pca is None:
-        return self._transformer
-    else:
-        return self._transformer.dot(self.pca.components_)
+    return self._transformer
 
   def _process_inputs(self, X, Y):
     X = np.asanyarray(X)
@@ -93,11 +90,12 @@ class RCA(BaseMetricLearner):
 
     # If the inner covariance matrix is not full rank,
     # the input data are first projected with a PCA to a space of dimension rank.
-    self.pca = None
+    M_pca = np.identity(d)
     if rank < d:
-      self.pca = decomposition.PCA(n_components = rank)
-      self.pca.fit(data)
-      data = self.pca.transform(data)
+      pca = decomposition.PCA(n_components = rank)
+      pca.fit(data)
+      M_pca = pca.components_
+      data = pca.transform(data)
       chunk_mask, chunk_data, chunk_labels = _process_chunks(data, chunks, num_chunks)
       inner_cov, rank = _compute_inner_cov(chunk_data)
 
@@ -119,6 +117,8 @@ class RCA(BaseMetricLearner):
       self._transformer = _inv_sqrtm(inner_cov).dot(A.T)
     else:
       self._transformer = _inv_sqrtm(inner_cov).T
+        
+    self._transformer = self._transformer.dot(M_pca)
 
     return self
 
--- a/metric_learn/rca.py
+++ b/metric_learn/rca.py
@@ -33,13 +33,6 @@ def _process_chunks(data, chunks, num_chunks):
 
   return chunk_mask, chunk_data, chunk_labels
 
-# "inner" covariance of chunk deviations
-def _compute_inner_cov(chunk_data):
-  inner_cov = np.cov(chunk_data, rowvar=0, bias=1)
-  rank = np.linalg.matrix_rank(inner_cov)
-
-  return inner_cov, rank
-
 
 class RCA(BaseMetricLearner):
   """Relevant Components Analysis (RCA)"""
@@ -86,7 +79,8 @@ class RCA(BaseMetricLearner):
     """
     data, chunks, num_chunks, d = self._process_inputs(data, chunks)
     chunk_mask, chunk_data, chunk_labels = _process_chunks(data, chunks, num_chunks)
-    inner_cov, rank = _compute_inner_cov(chunk_data)
+    inner_cov = np.cov(chunk_data, rowvar=0, bias=1)
+    rank = np.linalg.matrix_rank(inner_cov)
 
     # If the inner covariance matrix is not full rank,
     # the input data are first projected with a PCA to a space of dimension rank.
@@ -97,7 +91,8 @@ class RCA(BaseMetricLearner):
       M_pca = pca.components_
       data = pca.transform(data)
       chunk_mask, chunk_data, chunk_labels = _process_chunks(data, chunks, num_chunks)
-      inner_cov, rank = _compute_inner_cov(chunk_data)
+      inner_cov = np.cov(chunk_data, rowvar=0, bias=1)
+      rank = np.linalg.matrix_rank(inner_cov)
 
     # The embedding dimension must be smaller than the rank of the inner covariance matrix
     dim = min(self.params['dim'], rank)
--- a/metric_learn/rca.py
+++ b/metric_learn/rca.py
@@ -20,6 +20,7 @@ import warnings
 from .base_metric import BaseMetricLearner
 from .constraints import Constraints
 
+
 def _process_chunks(data, chunks, num_chunks):
   # mean center
   data -= data.mean(axis=0)
@@ -45,10 +46,8 @@ class RCA(BaseMetricLearner):
     dim : int, optional
         embedding dimension (default: original dimension of data)
     """
-    self.params = {
-      'dim': dim,
-    }
-  
+    self.params = {'dim': dim}
+
   def transformer(self):
     return self._transformer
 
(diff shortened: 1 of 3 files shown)


ab-anssi:
I have solved the first three issues.

I have added the digits test because this dataset is "more complex" than the Iris one. I think it is important to test on more than one dataset. For instance, the current implementation of RCA does not work on the digits dataset.
I understand your concern about the execution time of the tests. RCA is very fast so it does not really impact the execution time of the test :
    - Without the digits test :  26.723 sec
    - With the digits test    :  27,763 sec
However, I agree that expensive algorithms, such as NCA, should not be executed on the digits dataset in the tests.

Do you want me to remove the test of RCA on the digits dataset ?

metric_learn/rca.py
```diff
@@ -62,32 +79,42 @@ def fit(self, data, chunks):
...
+      chunk_mask, chunk_data, chunk_labels = _process_chunks(data, chunks, num_chunks)
+      inner_cov = np.cov(chunk_data, rowvar=0, bias=1)
+      rank = np.linalg.matrix_rank(inner_cov)
+
+    # The embedding dimension must be smaller than the rank of the inner covariance matrix
+    dim = min(self.params['dim'], rank)
+    if dim < self.params['dim']:
+      mess  = 'The embedding dimension must be smaller than the rank of the inner covariance matrix. '
+      mess += 'dim is set to ' + str(dim) + '.'
+      warnings.warn(mess)
```
perimosocordiae: It seems like we can clean up the conditionals a bit here:
```python
dim = self.params['dim']
if dim > rank:
  warnings.warn('...')
  dim = rank
```

metric_learn/rca.py
```diff
@@ -62,32 +79,42 @@ def fit(self, data, chunks):
...
       vals, vecs = np.linalg.eig(tmp)
-      inds = np.argsort(vals)[:self.params['dim']]
+      inds = np.argsort(vals)[:dim]
       A = vecs[:,inds]
       inner_cov = A.T.dot(inner_cov).dot(A)
       self._transformer = _inv_sqrtm(inner_cov).dot(A.T)
     else:
       self._transformer = _inv_sqrtm(inner_cov).T
+        
+    self._transformer = self._transformer.dot(M_pca)
```
perimosocordiae: I'd rather have `M_pca` default to `None` instead of the identity matrix, then for this line do:
```python
if M_pca is not None:
  self._transformer = self._transformer.dot(M_pca)
```

test/metric_learn_test.py
```diff
@@ -119,6 +122,19 @@ def test_iris(self):
     rca.fit(self.iris_points, self.iris_labels)
     csep = class_separation(rca.transform(), self.iris_labels)
     self.assertLess(csep, 0.25)
+  def test_feature_null_variance(self):
```
perimosocordiae: Add a blank line between methods.

metric_learn/rca.py
```diff
@@ -62,32 +79,42 @@ def fit(self, data, chunks):
...
+    # the input data are first projected with a PCA to a space of dimension rank.
+    M_pca = np.identity(d)
+    if rank < d:
+      pca = decomposition.PCA(n_components = rank)
+      pca.fit(data)
+      M_pca = pca.components_
+      data = pca.transform(data)
+      chunk_mask, chunk_data, chunk_labels = _process_chunks(data, chunks, num_chunks)
+      inner_cov = np.cov(chunk_data, rowvar=0, bias=1)
+      rank = np.linalg.matrix_rank(inner_cov)
```
ab-anssi: I have addressed all your remarks apart from this one. I have a few remarks/questions.

   The 3 papers presenting RCA (see below) emphasize that it is really important to perform a PCA before RCA to scale down noise dimensions. However, several methods are proposed to apply the PCA.

- Embedding dimension, denoted by `r`
   - In the paper ECCV 2002, `r` is set to the effective rank of the inner covariance matrix, i.e. the number of singular values that are significantly greater than 0 (see algorithm p.7).
   - Currently, I set `r` to the rank of the inner covariance matrix, i.e. the number of singular values that are not null. Perhaps, we should compute the effective rank as proposed in ECCV 2002. But how do we set the threshold `t` ? We should perhaps run some tests.
   - The paper JMLR 2005 presents another value for `r` computed from the size of the chunklets (see Algorithm 2 p.12). I do not really understand why this value is a good choice, and the computation depends on a parameter `alpha` (between 0 and 1). How do we set this parameter ?

- Rank of the inner covariance matrix after PCA
The PCA components are computed from the inner covariance matrix and are used to project the original data, and finally the inner covariance matrix is recomputed. The final inner covariance matrix is not the projection of the original covariance matrix on its PCA components.
Thus, as stated in the paper ECCV 2002,  the rank of the inner covariance matrix may become smaller than `r` after the projection. Perhaps, setting the threshold `t` high enough may prevent this issue. Another solution may be to keep projecting the original data to fewer principal components until the inner covariance matrix is full rank.

What I want to do :
  - Setting `r` to the the number of singular values of the inner covariance matrix that are significantly greater than 0. The remaining question is "how do I set the threshold" ?
  - Keeping on projecting the original data to fewer principal components until the inner covariance matrix is full rank. This way the implementation is robust, and we make no optimistic assumption about the rank of the inner covariance matrix.

What do you think of this proposal ?

The papers presenting RCA :
1. Adjustment learning and relevant component analysis, [ECCV 2002](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.19.2871&rep=rep1&type=pdf)
2. Learning distance functions using equivalence relations, [ICML 2003](https://pdfs.semanticscholar.org/0029/cf474bce70cfab2e944c4be01f99e741f1f4.pdf)
3. Learning a mahalanobis metric from equivalence constraints, [JMLR 2005](http://www.jmlr.org/papers/volume6/bar-hillel05a/bar-hillel05a.pdf)

perimosocordiae: Thanks for the meta-analysis, this really helps. I like the idea of starting with a PCA transformation unconditionally. The principal components are sorted by how much variance they explain, so we can just do this once and take the first `d` dimensions that produce a full-rank inner covariance fairly cheaply.

We should still let the user choose an embedding dimension, so the process might look like:

1. If the user asks for `d` dimensions (defaulting to `X.shape[1]`), compute the first `d` principal components.
2. Compute the inner covariance matrix and find its rank, `r`, using the standard non-null check.
3. If `r < d`, search for the appropriate `d` principal components and warn the user.

I think this avoids the need for any additional parameters and keeps the implementation pretty straightforward. If the user wants to prune small-but-non-null singular values, they can do so by setting `d`.

ab-anssi: I like the idea to warn the user if the inner covariance matrix is not full rank, and to give advice about how to solve the issue.
However, if the user wants `d` dimensions, I don't think we should keep only the  `d` first principal components because we may lose some discriminative information. For instance, if the user wants only 2 dimensions for data visualization, keeping only the first two components is not enough : they may represent a very small part of the variance of the data, and the discriminative information may be in the unselected components.

My proposal :

1. Compute PCA and select the first components such that 90% of the variance is explained (easy to obtain with scikit-learn).
2. Compute the inner covariance matrix of the projected data and compute its rank using the standard non-null check.
3. If the inner covariance matrix is not full rank, then warn the user: 
"The inner covariance matrix is not invertible, and so the transformation matrix may contain Nan values. You should preprocess the data, with PCA for instance, to remove noise and redundant information."
We check that the inner covariance matrix is full rank, and not that `d <= r`. Indeed, if `d <= r` but the inner covariance matrix is not full rank, then we are going to select null eigenvalues (or very small ones).
4. If the inner covariance matrix is full rank, but `d > r`, warn the user that the embedding dimension has been reduced to `r`. 

What do you think of this proposal ?

perimosocordiae: I think taking 90% of the variance is too low, and telling the user to preprocess with PCA when we're already preprocessing with PCA seems strange. Maybe we could add a parameter `pca_comps=None` which would then be used for the PCA preprocessing:

```python
data = PCA(n_components=pca_comps).fit_transform(data)
```

Then, if the inner-cov is not full rank, we can tell the user to adjust `pca_comps`. We can also allow it to be dual-purpose (like sklearn's `n_components`), where a value between 0 and 1 indicates %variance, and an integer >= 1 indicates the number of components to use.

Other than that, I like your proposal.

ab-anssi: I like this solution ! I will update the pull request.

ab-anssi:
I have added the `pca_comps` parameter. Can you please review the code ?

metric_learn/rca.py
```diff
@@ -8,47 +8,68 @@
...
-  def __init__(self, dim=None):
+  def __init__(self, dim=None, pca_comps=None):
     """Initialize the learner.
 
     Parameters
     ----------
     dim : int, optional
         embedding dimension (default: original dimension of data)
+    pca_comps : int, float, None or string
+        see the sklearn documentation
```
perimosocordiae: We should just explain what this is, and then maybe give a reference to sklearn. Maybe something like:
```
pca_comps : int or float, optional
    Number of components to keep during PCA preprocessing.
    If None (default), does not perform PCA.
    If ``0 < pca_comps < 1``, it is used as the minimum explained variance ratio.
    See sklearn.decomposition.PCA for more details.
```

metric_learn/rca.py
```diff
@@ -8,47 +8,68 @@
...
+    data = np.asanyarray(data)
+    self.X = data
+    n, d = data.shape
+    return data, d
 
+  def _process_chunks(self, data, chunks):
+    chunks = np.asanyarray(chunks)
+    num_chunks = chunks.max() + 1
+    chunk_mask, chunk_data = _chunk_mean_centering(data, chunks, num_chunks)
+    return chunk_mask, chunk_data
```
perimosocordiae: `return _chunk_mean_centering(data, chunks, num_chunks)`

metric_learn/rca.py
```diff
@@ -8,47 +8,68 @@
...
-
-    Y = np.asanyarray(Y)
-    num_chunks = Y.max() + 1
-
-    return X, Y, num_chunks, d
+    if self.params['dim'] > d:
+      self.params['dim'] = d
+      warnings.warn('dim must be smaller than the data dimension. ' +
+                    'dim is set to %d.' % (d))
+    return self.params['dim']
```
perimosocordiae: We should also check that `self.params['dim'] >= 1`, and raise a ValueError if it's not.

metric_learn/rca.py
```diff
@@ -61,34 +82,44 @@ def fit(self, data, chunks):
...
-    chunk_labels = chunks[chunk_mask]
-    for c in xrange(num_chunks):
-      mask = chunk_labels == c
-      chunk_data[mask] -= chunk_data[mask].mean(axis=0)
+    # PCA projection to remove noise and redundant information.
+    M_pca = None
+    if self.params['pca_comps'] is not None:
+      pca = decomposition.PCA(n_components=self.params['pca_comps'],
+                              svd_solver='full')
+      data, d = pca.fit_transform(data), self.params['pca_comps']
```
perimosocordiae: This should be on two separate lines. Also, if `pca_comps` is a float (like 0.9, for instance), this will cause a problem.

metric_learn/rca.py
```diff
@@ -61,34 +82,44 @@ def fit(self, data, chunks):
...
+      M_pca = pca.components_
 
-    # "inner" covariance of chunk deviations
+    data -= data.mean(axis=0)
```
perimosocordiae: If we did PCA, the data will already be mean-centered. This line should be in an `else` condition for the PCA check.

metric_learn/rca.py
```diff
@@ -61,34 +82,44 @@ def fit(self, data, chunks):
...
+    chunk_mask, chunk_data = self._process_chunks(data, chunks)
     inner_cov = np.cov(chunk_data, rowvar=0, bias=1)
+    rank = np.linalg.matrix_rank(inner_cov)
+
+    if rank < d:
+      warnings.warn('The inner covariance matrix is not invertible, ' +
+                    'so the transformation matrix may contain Nan values. ' +
+                    'You should adjust pca_comps to remove noise and ' +
+                    'redundant information.')
```
perimosocordiae: No need to use `+` here. The string literals will automatically concatenate without any operators between them.

test/metric_learn_test.py
```diff
@@ -120,6 +120,15 @@ def test_iris(self):
...
 
+  def test_feature_null_variance(self):
+    rca = RCA_Supervised(dim=2, pca_comps=3, num_chunks=30, chunk_size=2)
+    X = np.hstack((self.iris_points,
+        np.eye(self.iris_points.shape[0], M = 1)))
+    rca.fit(X, self.iris_labels)
+    csep = class_separation(rca.transform(), self.iris_labels)
+    self.assertLess(csep, 0.30)
+
+
```
perimosocordiae: We should also have a test using `pca_comps=0.95`. It can use the same `X` matrix and check the same result.

test/metric_learn_test.py
```diff
@@ -120,6 +120,15 @@ def test_iris(self):
     csep = class_separation(rca.transform(), self.iris_labels)
     self.assertLess(csep, 0.25)
 
+  def test_feature_null_variance(self):
+    rca = RCA_Supervised(dim=2, pca_comps=3, num_chunks=30, chunk_size=2)
+    X = np.hstack((self.iris_points,
+        np.eye(self.iris_points.shape[0], M = 1)))
```
perimosocordiae: `X = np.hstack((self.iris_points, np.eye(len(self.iris_points), M=1)))`

ab-anssi:
Thank you for reviewing so quickly ! I have addressed your comments.

metric_learn/rca.py
```diff
@@ -61,34 +86,46 @@ def fit(self, data, chunks):
...
+    # PCA projection to remove noise and redundant information.
+    M_pca = None
+    if self.params['pca_comps'] is not None:
+      pca = decomposition.PCA(n_components=self.params['pca_comps'],
+                              svd_solver='full')
+      data = pca.fit_transform(data)
+      d = data.shape[1]
+      M_pca = pca.components_
+    else:
+        data -= data.mean(axis=0)
```
perimosocordiae: This line is indented too far.

PR #56, opened 2017-04-24
DML using Evolution

svecon:
This is an evolution based approach to DML. This is a code that I used for my novel method using CMAES and kNN to evolve a Mahalanobis metric. It is based on ideas from Fukui et al. [Evolutionary distance metric learning approach to semi-supervised clustering with neighbor relations].

I think it will need some more refactoring, I am open to suggestions.


perimosocordiae:
Wow, this is quite the PR! I haven't read through the whole thing yet, but my initial reaction is that including a complete implementation of CMAES seems like overkill. I'd be much happier adding an optional dependency on a CMAES library and having the code here just focus on the metric learning parts. Similarly, I'd rather use an existing neural network library than include an implementation here.

svecon:
I am not including a complete implementation of CMAES nor Neural Networks. I am using DEAP library for evolution algorithms, that CMAES class is a necessary boilerplate code to run it.

For the neural networks, I only use a forward inference, which is only a simple for-cycle. I dont think a library dependency is needed just because of that.

svecon:
Hello,
I have refactored the whole method, I hope it will be more understandable this way. Any feedback is appreciated, I would be very happy having this merged eventually :)

I don't know why the unittests fail in autoloader:
AttributeError: 'module' object has no attribute 'metric_learn_test'

Any ideas?

svecon:
Managed to fix the Travis build. One of the libraries was not available in Python 2.

test/test_sklearn_compat.py
```diff
@@ -56,6 +58,14 @@ def test_lsml(self):
   def test_itml(self):
     check_estimator(dITML)
 
+  # This fails in check_dtype_object, not sure why
+  # TODO !!!!!
```
perimosocordiae: Any followup on this? I had to do some debugging when adding these tests initially, and it's not always obvious why they fail.

svecon:
Yes, the test passes after the last refactoring.

Also exposed more parameters in the main CMAES and JDE class wrappers.

bellet:
Given that this has been inactive for a long time, and that meanwhile the package has undergone massive changes, I will close this for now

PR #59, opened 2017-05-16
Clarified documentation of ITML

### LINKED ISSUE #58: Misleading documentation of ITML parameter "constraints"
The documentation of `ITML.fit()` states the following about the parameter `constraints`:

    constraints : 4-tuple of arrays
        (a,b,c,d) indices into X, such that d(X[a],X[b]) < d(X[c],X[d])

But as far as it is obvious to me from the code, the criterion mentioned there is neither used in the ITML paper nor in the code. The only relevant criterion is the one mentioned in the documentation of `bounds`:

    bounds : list (pos,neg) pairs, optional
        bounds on similarity, s.t. d(X[a],X[b]) < pos and d(X[c],X[d]) > neg

The current docstring of `constraints` might be interpreted as if there actually was some kind of dependence between the pairs `(a[i], b[i])` and `(c[i], d[i])`. However, as far as I understand, the order of the pairs in `(c,d)` is completely independent from the one of the pairs in `(a,b)`.

Callidior:
Resolves #58

### DIFF
--- a/metric_learn/itml.py
+++ b/metric_learn/itml.py
@@ -80,7 +80,8 @@ class ITML(BaseMetricLearner):
     X : (n x d) data matrix
         each row corresponds to a single instance
     constraints : 4-tuple of arrays
-        (a,b,c,d) indices into X, such that d(X[a],X[b]) < d(X[c],X[d])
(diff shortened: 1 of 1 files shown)


perimosocordiae:
Looks good, thanks for the patch!

Issue #57, opened 2017-05-15
Inconsistent conversion between transformer and metric


Callidior:
According to the documentation, the learned distance between `x` and `y` is expressed as `sqrt((x-y).dot(inv(M)).dot(x-y))`, which is equivalent to evaluating the Euclidean distance in the transformed space obtained from `dot(X, L.T)` for `n x d` data matrices `X` with `L = inv(cholesky(M))`.

However, if an algorithm does not learn a metric `M`, but its corresponding transformer `L` directly, the current implementation of `BaseMetricLearner.metric()` computes `M` as `L.T.dot(L)`, which actually equals `inv(M)`, according to the definition of `L` given above.

If I got the documentation right, then the following *should* hold:

    learner.transformer().T.dot(learner.transformer()) == learner.metric()

This is, however, currently not the case for algorithms learning `M` (e.g., ITML, LSML or even the covariance method).

----

In addition, the definition of a "metric" is not even consistent across different learners: ITML and LSML learn a metric `sqrt((x-y).dot(M).dot(x-y))`, while the covariance method and SDLM (I guess) use `inv(M)` as stated in the readme.

Thus, **the transformers returned by ITML and LSML are wrong!**

I would suggest harmonizing the terminology towards the one currently employed by ITML and LSML, since it seems also predominant in literature, in my opinion. This would require removing the `inv()` from `BaseMetricLearner.transformer()` as well as changing all learners relying on the current terminology, such as ITML ans LSML.

perimosocordiae:
Thanks for bringing this issue up. I agree that we need to be more consistent with the mapping between transformers and metrics.

I want to make sure we get this right, so I'm going to re-read some of the literature and see if I come to the same conclusion you did.

Callidior:
Thanks for looking into this!
I completely agree that a change like this needs to be made very carefully.

Just in case you come to the same conclusion as I, I've already tried to make the necessary changes in the Pull Request referenced above.
Since ITML and LSML return just wrong transformers at the moment, their performance could be significantly improved by fixing this issue.

perimosocordiae:
After letting this languish for too long, I think I'm convinced. Thanks for pointing this out!

PR #61, opened 2017-05-23
Implementation of MMC

Callidior:
This PR implements Probabilistic Global Distance Metric Learning (PGDM) mentioned in #13, sometimes also referred to as MMC, originally proposed in the following paper:

> Eric P. Xing, Andrew Y. Ng, Michael I. Jordan and Stuart Russell
> "Distance metric learning with application to clustering with side-information."
> NIPS. Vol. 15. No. 505-512. 2002.

The implementation is mainly a translation of the [Matlab reference code](http://www.cs.cmu.edu/%7Eepxing/papers/Old_papers/code_Metric_online.tar.gz) into Python, only optimized a bit to make use of vectorization and the power of numpy. This makes this implementation 6 times faster than the reference code on the iris dataset.

Both variants of PGDM proposed in the original paper are supported: learning a full or a diagonal Mahalanobis matrix.

The implementation has been validated against the results obtained from the reference code on the iris dataset. Corresponding unit tests are provided.

The class separation score of less than 0.15 on the iris dataset is one of the best scores currently reported in the unit tests for this dataset.

----

Note that PGDM learns only a positive *semi*-definite metric `M`, which precludes obtaining a transformer `L` with `L.T*L=M`  using the Cholesky decomposition. Instead, `PGDM` overrides `transformer()` and uses an eigenvector decomposition `M = V * W * V.T` to obtain a transformer `L = V.T * W^(-1/2)`.

----

**Important note:** PGDM learns a distance function `(x-y)^T * M * (x-y)`, as opposed to `(x-y)^T * inv(M) * (x-y)` as currently stated in the documentation of `metric_learn`. Thus, this PR depends on solving issue #57, e.g., by merging PR #60. If you decide to stick to the current definition of a "metric", `PGDM.metric()` would need to be changed in order to return the inverse of the learned matrix.

### DIFF
--- a/metric_learn/mmc.py
+++ b/metric_learn/mmc.py
@@ -1,7 +1,7 @@
 """
-Probabilistic Global Distance Metric Learning, Xing et al., NIPS 2002
+Mahalanobis Metric Learning with Application for Clustering with Side-Information, Xing et al., NIPS 2002
 
-PGDM minimizes the sum of squared distances between similar examples,
+MMC minimizes the sum of squared distances between similar examples,
 while enforcing the sum of distances between dissimilar examples to be
 greater than a certain margin.
 This leads to a convex and, thus, local-minima-free optimization problem
@@ -9,7 +9,7 @@ that can be solved efficiently.
 However, the algorithm involves the computation of eigenvalues, which is the
 main speed-bottleneck.
 Since it has initially been designed for clustering applications, one of the
-implicit assumptions of PGDM is that all classes form a compact set, i.e.,
+implicit assumptions of MMC is that all classes form a compact set, i.e.,
 follow a unimodal distribution, which restricts the possible use-cases of
 this method. However, it is one of the earliest and a still often cited technique.
 
@@ -28,11 +28,11 @@ from ._util import vector_norm
 
 
 
-class PGDM(BaseMetricLearner):
-  """Probabilistic Global Distance Metric Learning (PGDM)"""
+class MMC(BaseMetricLearner):
+  """Mahalanobis Metric for Clustering (MMC)"""
   def __init__(self, max_iter=100, max_proj=10000, convergence_threshold=1e-3,
                A0=None, diagonal=False, diagonal_c=1.0, verbose=False):
-    """Initialize PGDM.
+    """Initialize MMC.
     Parameters
     ----------
     max_iter : int, optional
@@ -59,7 +59,7 @@ class PGDM(BaseMetricLearner):
     self.verbose = verbose
   
   def fit(self, X, constraints):
-    """Learn the PGDM model.
+    """Learn the MMC model.
     Parameters
     ----------
     X : (n x d) data matrix
@@ -85,9 +85,9 @@ class PGDM(BaseMetricLearner):
     no_ident = vector_norm(X[c] - X[d]) > 1e-9
     c, d = c[no_ident], d[no_ident]
     if len(a) == 0:
-      raise RuntimeError('No similarity constraints given for PGDM.')
+      raise RuntimeError('No similarity constraints given for MMC.')
     if len(c) == 0:
-      raise RuntimeError('No dissimilarity constraints given for PGDM.')
+      raise RuntimeError('No dissimilarity constraints given for MMC.')
     
     # init metric
     if self.A0 is None:
@@ -102,7 +102,7 @@ class PGDM(BaseMetricLearner):
     return a,b,c,d
 
   def _fit_full(self, X, constraints):
-    """Learn full metric using PGDM.
+    """Learn full metric using MMC.
     Parameters
     ----------
     X : (n x d) data matrix
@@ -208,22 +208,22 @@ class PGDM(BaseMetricLearner):
       if delta < self.convergence_threshold:
         break
       if self.verbose:
-        print('pgdm iter: %d, conv = %f, projections = %d' % (cycle, delta, it+1))
+        print('mmc iter: %d, conv = %f, projections = %d' % (cycle, delta, it+1))
 
     if delta > self.convergence_threshold:
       self.converged_ = False
       if self.verbose:
-        print('pgdm did not converge, conv = %f' % (delta,))
+        print('mmc did not converge, conv = %f' % (delta,))
     else:
       self.converged_ = True
       if self.verbose:
-        print('pgdm converged at iter %d, conv = %f' % (cycle, delta))
+        print('mmc converged at iter %d, conv = %f' % (cycle, delta))
     self.A_[:] = A_old
     self.n_iter_ = cycle
     return self
   
   def _fit_diag(self, X, constraints):
-    """Learn diagonal metric using PGDM.
+    """Learn diagonal metric using MMC.
     Parameters
     ----------
     X : (n x d) data matrix
@@ -275,7 +275,7 @@ class PGDM(BaseMetricLearner):
       w[:] = w_previous
       error = np.abs((obj_previous - obj_initial) / obj_previous)
       if self.verbose:
-        print('pgdm iter: %d, conv = %f' % (it, error))
+        print('mmc iter: %d, conv = %f' % (it, error))
       it += 1
     
     self.A_ = np.diag(w)
@@ -361,7 +361,7 @@ class PGDM(BaseMetricLearner):
     L = V.T * w^(-1/2), with A = V*w*V.T being the eigenvector decomposition of A with
     the eigenvalues in the diagonal matrix w and the columns of V being the eigenvectors.
     
-    The Cholesky decomposition cannot be applied here, since PGDM learns only a positive
+    The Cholesky decomposition cannot be applied here, since MMC learns only a positive
     *semi*-definite Mahalanobis matrix.
     
     Returns
@@ -375,8 +375,8 @@ class PGDM(BaseMetricLearner):
       return V.T * np.sqrt(np.maximum(0, w[:,None]))
 
 
-class PGDM_Supervised(PGDM):
-  """Probabilistic Global Distance Metric Learning (PGDM)"""
+class MMC_Supervised(MMC):
+  """Mahalanobis Metric for Clustering (MMC)"""
   def __init__(self, max_iter=100, max_proj=10000, convergence_threshold=1e-6,
                num_labeled=np.inf, num_constraints=None,
                A0=None, diagonal=False, diagonal_c=1.0, verbose=False):
@@ -402,15 +402,15 @@ class PGDM_Supervised(PGDM):
     verbose : bool, optional
         if True, prints information while learning
     """
-    PGDM.__init__(self, max_iter=max_iter, max_proj=max_proj,
-                  convergence_threshold=convergence_threshold,
-                  A0=A0, diagonal=diagonal, diagonal_c=diagonal_c,
-                  verbose=verbose)
+    MMC.__init__(self, max_iter=max_iter, max_proj=max_proj,
+                 convergence_threshold=convergence_threshold,
+                 A0=A0, diagonal=diagonal, diagonal_c=diagonal_c,
+                 verbose=verbose)
     self.num_labeled = num_labeled
     self.num_constraints = num_constraints
 
   def fit(self, X, y, random_state=np.random):
-    """Create constraints from labels and learn the PGDM model.
+    """Create constraints from labels and learn the MMC model.
     Parameters
     ----------
     X : (n x d) matrix
@@ -430,4 +430,4 @@ class PGDM_Supervised(PGDM):
                                   random_state=random_state)
     pos_neg = c.positive_negative_pairs(num_constraints,
                                         random_state=random_state)
-    return PGDM.fit(self, X, pos_neg)
+    return MMC.fit(self, X, pos_neg)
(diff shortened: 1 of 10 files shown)


Callidior:
I'm not sure whether PGDM is the correct name for this algorithm. The [survey of Yang](http://www.cs.cmu.edu/%7Eliuy/frame_survey_v2.pdf) introduces this method by Xing et al. as "Supervised Global Distance Metric Learning" and then extends this approach in a subsection to "Probabilistic Global Distance Metric Learning". But the paper and the reference implementation referred to in #13 are not probabilistic at all.

Maybe we should change the name to just "Global Distance Metric Learning", which could be abbreviated as GDML, or to MMC, as suggested by [Weinberger et al.](http://www.jmlr.org/papers/volume10/weinberger09a/weinberger09a.pdf)

How do you think about this?

perimosocordiae:
I agree that the wish list was wrongly conflating this approach with the probabilistic version from the survey, and I'm happy to rename this version to something better.

Despite the referenced usage, I think MMC is a bit of a bad name, as the original paper uses clustering as an example application, not the entire basis for the work. Do any other citations propose nice names for this algorithm? In lieu of any better ideas, GDML can work, even though it's pretty uninformative.

test/metric_learn_test.py
```diff
@@ -149,5 +149,48 @@ def test_iris(self):
...
+    n = self.iris_points.shape[0]
+    a, b, c, d = [], [], [], []
+    for i in range(n):
+      for j in range(i+1, n):
+        if self.iris_labels[i] == self.iris_labels[j]:
+          a.append(i)
+          b.append(j)
+        else:
+          c.append(i)
+          d.append(j)
```
perimosocordiae: This can be made a bit nicer with vectorization:
```python
mask = self.iris_labels[None] == self.iris_labels[:,None]
a, b = np.nonzero(np.triu(mask, k=1))
c, d = np.nonzero(np.triu(~mask, k=1))
```

test/metric_learn_test.py
```diff
@@ -149,5 +149,48 @@ def test_iris(self):
...
+
+    # Full metric
+    pgdm = PGDM(convergence_threshold = 0.01)
```
perimosocordiae: Style nitpick: don't leave spaces around the `=` in keyword arguments. (here and throughout)

metric_learn/pgdm.py
```diff
@@ -0,0 +1,436 @@
...
+import numpy as np
+from six.moves import xrange
+from sklearn.metrics import pairwise_distances
+from sklearn.utils.validation import check_array, check_X_y
+
+from .base_metric import BaseMetricLearner
+from .constraints import Constraints
+
+
+# hack around lack of axis kwarg in older numpy versions
```
perimosocordiae: This code is used in ITML as well, so let's pull this hack out into a new file, `metric_learn/_util.py`.

metric_learn/pgdm.py
```diff
@@ -0,0 +1,436 @@
...
+    else:
+      return self._fit_full(X, constraints)
+  
+  def _process_inputs(self, X, constraints):
+    
+    self.X_ = X = check_array(X)
+    
+    # check to make sure that no two constrained vectors are identical
+    a,b,c,d = constraints
+    ident = _vector_norm(X[a] - X[b]) > 1e-9
```
perimosocordiae: In this usage, `ident` is actually a mask where True implies non-identical. Maybe `non_ident` instead?

metric_learn/pgdm.py
```diff
@@ -0,0 +1,436 @@
...
+    a, b = a[ident], b[ident]
+    ident = _vector_norm(X[c] - X[d]) > 1e-9
+    c, d = c[ident], d[ident]
```
perimosocordiae: We should also ensure that our constraint vectors aren't empty after this sanity check.

Callidior:
I agree with you that both MMC and GDML are not the most informative names. Unfortunately, I haven't spotted any other term for this algorithm in the existing literature yet and Xing et al. completely avoid giving their algorithm a name (besides just "metric learning").

Although this algorithm is, for sure, applicable to more applications than just clustering, one could also argue that learning a metric that is particularly suitable for clustering is an essential part of the method's objective. It minimizes the total intra-class distance, while keeping the total inter-class distance larger than a given margin. Thus, the classes would be tightly clustered together in the space enforced by such a metric.

However, I leave this decision to you. As soon as you notify me about the final name, I will rename everything and update this PR.

metric_learn/pgdm.py
```diff
@@ -0,0 +1,436 @@
...
+    pos_diff = X[a] - X[b]
+    w = np.einsum('ij,ik->jk', pos_diff, pos_diff).ravel()
+    # `w` is the sum of all outer products of the rows in `pos_diff`.
+    # The above `einsum` is equivalent to the much more inefficient:
+    # w = np.apply_along_axis(
+    #         lambda x: np.outer(x,x).ravel(),
+    #         1,
+    #         X[a] - X[b]
+    #     ).sum(axis = 0)
+    t = w.dot(A.ravel() / 100.0)
```
perimosocordiae: Is there an explanation for this division by 100? Combined with the division by 10 in `_prepare_inputs`, there's a lot of magic happening to `A`...

metric_learn/pgdm.py
```diff
@@ -0,0 +1,436 @@
...
+main speed-bottleneck.
+Since it has initially been designed for clustering applications, one of the
+implicit assumptions of PGDM is that all classes form a compact set, i.e.,
+follow a unimodal distribution, which restricts the possible use-cases of
+this method. However, it is one of the earliest and a still often cited technique.
+
+Adapted from Matlab code at http://www.cs.cmu.edu/%7Eepxing/papers/Old_papers/code_Metric_online.tar.gz
+"""
+
+from __future__ import print_function, absolute_import
```
perimosocordiae: Importing `division` is a good idea as well.

metric_learn/pgdm.py
```diff
@@ -0,0 +1,436 @@
...
+    # The above `einsum` is equivalent to the much more inefficient:
+    # w = np.apply_along_axis(
+    #         lambda x: np.outer(x,x).ravel(),
+    #         1,
+    #         X[a] - X[b]
+    #     ).sum(axis = 0)
+    t = w.dot(A.ravel() / 100.0)
+    
+    w1 = w / np.linalg.norm(w) # make `w` a unit vector
+    t1 = t / np.linalg.norm(w) # distance from origin to `w^T*x=t` plane
```
perimosocordiae: We should compute the norm of `w` once and reuse it.

perimosocordiae: Also, style nitpick: use two spaces between the end of the code and the comment character:

    foo()  # comment

metric_learn/pgdm.py
```diff
@@ -0,0 +1,436 @@
...
+          break
+      
+      # third constraint: gradient ascent
+      # max: g(A) >= 1
+      # here we suppose g(A) = fD(A) = \sum_{I,J \in D} sqrt(d_ij' A d_ij)
+      
+      obj_previous = self._fD(X, c, d, A_old) # g(A_old)
+      obj = self._fD(X, c, d, A)              # g(A)
+      
+      if ((obj > obj_previous) or (cycle == 0)) and (satisfy):
```
perimosocordiae: Don't need so many parens here:

    if satisfy and (obj > obj_previous or cycle == 0):

metric_learn/pgdm.py
```diff
@@ -0,0 +1,436 @@
...
+        
+        # If projection of 1 and 2 is successful, and such projection
+        # imprives objective function, slightly increase learning rate
```
perimosocordiae: imprives -> improves

metric_learn/pgdm.py
```diff
@@ -0,0 +1,436 @@
...
+    
+    Note that d_ij*A*d_ij' = tr(d_ij*A*d_ij') = tr(d_ij'*d_ij*A)
+    so, d(d_ij*A*d_ij')/dA = d_ij'*d_ij
+        df/dA = f'(\sum_{ij \in D} \sqrt{tr(d_ij'*d_ij*A)})
+                * 0.5*(\sum_{ij \in D} (1/sqrt{tr(d_ij'*d_ij*A)})*(d_ij'*d_ij))
+    """
+    dim = X.shape[1]
+    diff = X[c] - X[d]
+    M = np.einsum('ij,ik->ijk', diff, diff) # outer products of all rows in `diff`
+    dist = np.sqrt(np.sum(M * A[None,:,:], axis = (1,2)))
```
perimosocordiae: This line might also benefit from conversion to an `np.einsum` call.

perimosocordiae:
Okay, I can live with MMC.

metric_learn/pgdm.py
Callidior: I totally agree. I've copied this part from ITML. We should also change it there.

metric_learn/pgdm.py
Callidior: I also don't like this, but that magic is in the original reference code without any comment.

The division by 10 in `_prepare_inputs` could be removed and the learned matrix `A` would then be scaled by a factor of 10. That doesn't hurt the learned metric in any way, but the result wouldn't be identical to that of the reference implementation any more, so I left it this way.

Callidior:
Thanks for your valuable comments!

I hope to find time for implementing the requested changes until next week.

Callidior:
@perimosocordiae I've addressed your change requests and renamed to algorithm to MMC.

metric_learn/mmc.py
```diff
@@ -89,10 +80,14 @@ def _process_inputs(self, X, constraints):
...
-    ident = _vector_norm(X[c] - X[d]) > 1e-9
-    c, d = c[ident], d[ident]
+    no_ident = vector_norm(X[a] - X[b]) > 1e-9
+    a, b = a[no_ident], b[no_ident]
+    no_ident = vector_norm(X[c] - X[d]) > 1e-9
+    c, d = c[no_ident], d[no_ident]
+    if len(a) == 0:
+      raise RuntimeError('No similarity constraints given for MMC.')
+    if len(c) == 0:
+      raise RuntimeError('No dissimilarity constraints given for MMC.')
```
perimosocordiae: I think a `ValueError` makes more sense here. We should also specify that no _non-trivial_ constraints were provided, otherwise the user might be confused.

metric_learn/mmc.py
```diff
@@ -308,9 +304,9 @@ def _fD1(self, X, c, d, A):
     """
     dim = X.shape[1]
     diff = X[c] - X[d]
-    M = np.einsum('ij,ik->ijk', diff, diff) # outer products of all rows in `diff`
-    dist = np.sqrt(np.sum(M * A[None,:,:], axis = (1,2)))
-    sum_deri = np.sum(M / (2 * (dist[:,None,None] + 1e-6)), axis = 0)
+    M = np.einsum('ij,ik->ijk', diff, diff)    # outer products of all rows in `diff`
+    dist = np.sqrt(np.einsum('ijk,jk', M, A))  # equivalent to: np.sqrt(np.sum(M * A[None,:,:], axis=(1,2)))
+    sum_deri = np.einsum('ijk,i->jk', M, 0.5 / (dist + 1e-6))  # equivalent to: np.sum(M / (2 * (dist[:,None,None] + 1e-6)), axis=0)
```
perimosocordiae: Lines shouldn't go longer than 80 chars, so I'd put the comments on the previous line.

metric_learn/mmc.py
```diff
@@ -344,10 +340,11 @@ def _D_constraint(self, X, c, d, w):
...
     dist = np.sqrt(diff_sq.dot(w))
-    sum_deri1 = np.sum(diff_sq / (2 * np.maximum(dist, 1e-6))[:,None], axis = 0)
-    sum_deri2 = np.sum(
-      np.einsum('ij,ik->ijk', diff_sq, diff_sq) / (-4 * np.maximum(1e-6, dist**3)[:,None,None]),
-      axis = 0
+    sum_deri1 = np.einsum('ij,i', diff_sq, 0.5 / np.maximum(dist, 1e-6))
+    sum_deri2 = np.einsum(
+      'ijk,i',
+      np.einsum('ij,ik->ijk', diff_sq, diff_sq),
+      -0.25 / np.maximum(1e-6, dist**3)
```
perimosocordiae: Nested calls to `np.einsum` are pretty hard to read. I think the previous version was fine, but you could also pull out the inner `einsum` to a local variable if this version is significantly faster.

It's also possible to do `einsum` with >2 arrays, but that can have unintended consequences (see https://github.com/numpy/numpy/issues/5366). As of numpy 1.12 the situation is a bit better (see https://github.com/numpy/numpy/pull/5488), but I don't think we need to go down that route.

Callidior: Neither do I think so. However, this version is about 4 times faster, since it does not require broadcasting `dist` from 1-D to 3-D. So, I will just pull out the inner `einsum` to a local variable.

Callidior: I have found an even 10 times faster way to compute this without intermediate result in a single `einsum` call:

    sum_deri2 = np.einsum(
        'ij,ik->jk',
        diff_sq,
        diff_sq / (-4 * np.maximum(1e-6, dist**3))[:,None]
    )

Callidior:
@perimosocordiae Thanks, fixed :)

perimosocordiae:
Merged. Thanks for the PR, @Callidior!

Callidior:
Thanks for your helpful comments!

But please be aware that the metric learned by this algorithm defines a distance `(x-y)^T * M * (x-y)` and not `(x-y)^T * M^(-1) * (x-y)` as stated in the readme. But some other existing metric learner implementations do the same and hopefully we can get rid of that inconsistency soon :)

PR #62, opened 2017-05-23
Preventing a crash for LSML in a verbose mode

searchivarius:
If LSML is created with ``verbose=True``, the code crashes here https://github.com/searchivarius/metric-learn/blob/fac5ed5d4f0ae24d274e0956f4b7d2d6d09d2841/metric_learn/lsml.py#L95 due to uninitialized variable ``s_best``. This is I believe a way to fix the problem.

### DIFF
--- a/metric_learn/lsml.py
+++ b/metric_learn/lsml.py
@@ -69,7 +69,7 @@ class LSML(BaseMetricLearner):
     """
     self._prepare_inputs(X, constraints, weights)
     prior_inv = scipy.linalg.inv(self.M_)
-    s_best = self._total_loss(self.M_, prior_inv)
+    l_best = s_best = self._total_loss(self.M_, prior_inv)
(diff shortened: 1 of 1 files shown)


Callidior:
It's an interesting case if the program gets there with `l_best` uninitialized, because it means that no single gradient descent step could improve the loss. Thus, maybe `l_best = 0` is a more appropriate initialization?

searchivarius:
@Callidior I am torturing your software using substantially non-metric data :-) Whatever initialization is best in this case is up to you.

perimosocordiae:
`l_best` is recording the length of the best step size, so I agree that it makes sense to initialize it to zero.

searchivarius:
Hi guys, sure, fell free to make this modification.

perimosocordiae:
Just pushed a fix. Thanks, @searchivarius!

PR #64, opened 2017-05-24
Fixed parameter name in docstring of RCA.fit()

Callidior:
The docstring mentioned a parameter `X`, although it was actually called `data` in the function definition.


perimosocordiae:
Thanks, merged.

PR #69, opened 2017-08-03
Accelerate ITML on high-dimensional data

Callidior:
Admittedly, the changes proposed in this PR are very small, but in my test-case with 1024-dimensional data and 90 constraints, the run-time of ITML was reduced from 27 to 12 seconds.

Besides the changes in this PR, ITML could be accelerated further by not learning the metric `A` directly, but the decomposition `L` with `A = L^T * L` instead, as suggested in the ITML paper. However, that would require an efficient function for rank-1 updates/downdates of a Cholesky decomposition which is currently not available in `scipy` or `numpy`. And I don't think you would like to make another third-party module such as [`choldate`](https://github.com/jcrudy/choldate) a dependency just because of this.

### DIFF
--- a/metric_learn/itml.py
+++ b/metric_learn/itml.py
@@ -95,30 +95,30 @@ class ITML(BaseMetricLearner):
     gamma_proj = 1. if gamma is np.inf else gamma/(gamma+1.)
     pos_bhat = np.zeros(num_pos) + self.bounds_[0]
     neg_bhat = np.zeros(num_neg) + self.bounds_[1]
+    pos_vv = self.X_[a] - self.X_[b]
+    neg_vv = self.X_[c] - self.X_[d]
     A = self.A_
 
     for it in xrange(self.max_iter):
       # update positives
-      vv = self.X_[a] - self.X_[b]
-      for i,v in enumerate(vv):
+      for i,v in enumerate(pos_vv):
         wtw = v.dot(A).dot(v)  # scalar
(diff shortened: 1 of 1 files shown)


perimosocordiae:
I'm fine with adding an optional dependency it the speedup is significant, similar to the use of Shogun for LMNN.

Callidior:
The speedup resulting from working with the Cholesky decomposition of the metric instead of the metric itself depends on the dimensionality of the data. For the Iris dataset, there is no speedup. In my scenario with 1024-dimensional data, the runtime could further be reduced from 12 seconds to 9 seconds, so it's at least a slight acceleration.
However, since the implementation of ITML would have to be changed from estimating `A` to estimating `L`, it would be difficult to make this an *optional* dependency.

perimosocordiae:
Could have a completely separate implementation, `_CholeskyUpdateITML` for instance, and just swap out like we do with LMNN:

```python
try:
  import choldate
except ImportError:
  ITML = _ITML
else:
  ITML = _CholeskyUpdateITML
```

Common code could be shared via a base class.

perimosocordiae:
That said, I'm fine with leaving things as-is, unless there's a strong desire for the alternate version.

Callidior:
I agree. If some needs to process high-dimensional data with maximum possible speed, a `C`-implementation could be used for better efficiency. I've written one for my one purposes and it can handle my scenario in less than 2 seconds now.

Maybe its useful for some people ending up here: https://github.com/Callidior/libitml

PR #78, opened 2017-11-24
LMNN: fix mistake and improve performances

aje:
**Fix mistake in LMNN**
Issue in function _find_impostors:
 - the squared euclidean distance is used to compute the margins in variable "margin_radii"
 - the euclidean distance is used (through the function sklearn.metrics.pairwise.pairwise_distances) to compute distances between samples of different labels in variable "dist"
 - the issue is that the impostors are found by testing "dist < margin_radii" which is wrong because "dist" represent distances, and "margin_radii" represent squared distances.

I propose to solve this problem by computing always the squared distances.

**Faster LMNN** 
The use of the function sklearn.metrics.pairwise_distances gives horrible performances.
Replace it by a faster function.

Below is a little script to show the computation time of LMNN

import time
from sklearn.datasets import load_breast_cancer
from metric_learn import lmnn
dataset = load_breast_cancer()
s = time.time()
l1 = lmnn.LMNN(k=3)
l1.fit(dataset["data"], dataset["target"])
e = time.time()
print(e - s)

Before this commit, this script executes in 32 seconds
After this commit, this script executes in 1 second

### DIFF
--- a/metric_learn/lmnn.py
+++ b/metric_learn/lmnn.py
@@ -185,7 +185,7 @@ class python_LMNN(_base_LMNN):
     target_neighbors = np.empty((self.X_.shape[0], self.k), dtype=int)
     for label in self.labels_:
       inds, = np.nonzero(self.label_inds_ == label)
-      dd = euclidean_distances(self.X_[inds], self.X_[inds], squared=True)
+      dd = euclidean_distances(self.X_[inds], squared=True)
       np.fill_diagonal(dd, np.inf)
       nn = np.argsort(dd)[..., :self.k]
       target_neighbors[inds] = inds[nn]
--- a/metric_learn/lmnn.py
+++ b/metric_learn/lmnn.py
@@ -185,7 +185,7 @@ class python_LMNN(_base_LMNN):
     target_neighbors = np.empty((self.X_.shape[0], self.k), dtype=int)
     for label in self.labels_:
       inds, = np.nonzero(self.label_inds_ == label)
-      dd = pairwise_distances(self.X_[inds])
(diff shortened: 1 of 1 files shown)


perimosocordiae:
Thanks for the contribution! I agree with your analysis of the bug, though I was surprised to hear that sklearn's `pairwise_distances` was so slow.

So I looked a little closer and found that the issue is actually `metric='seuclidean'`. Regular L2 distance uses essentially the same algorithm as your `pairwiseEuclidean`:
https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/metrics/pairwise.py#L246

But if you use the `seuclidean` metric, it ends up calling Scipy's `cdist` function instead, which then has to make a copy of both input arrays to make them C-contiguous:
https://github.com/scipy/scipy/blob/v1.0.0/scipy/spatial/distance.py#L2361

So I think the easiest solution is, instead of rolling our own distance function, importing `euclidean_distances` from sklearn and calling it with `squared=True` anywhere that we want squared L2 distances. I did some quick testing and found that this also prevents the very slow running time.

metric_learn/lmnn.py
```diff
@@ -185,7 +185,7 @@ def _select_targets(self):
     target_neighbors = np.empty((self.X_.shape[0], self.k), dtype=int)
     for label in self.labels_:
       inds, = np.nonzero(self.label_inds_ == label)
-      dd = pairwise_distances(self.X_[inds])
+      dd = euclidean_distances(self.X_[inds], self.X_[inds], squared=True)
```
perimosocordiae: You can leave out the second argument, similar to how `pairwise_distances` works.

PR #85, opened 2018-02-26
[WIP] New API proposal

wdevazelhes:
This PR intends to provide a new API for the package, that would allow to use scikit-learn's tools for model selection and pipelining not only for supervised algorithms but also for weakly supervised ones. Indeed we would like to be able for instance to have a dataset of pairs, an algorithm that works on labeled pairs (i.e. pairwise constraints) and plug it into say scikit-learn's 'roc_auc' cross-validation scoring. This cross-validation scoring would split on the **pairs** and return the score of a classification of **pairs**. However we don't want to build an explicit dataset of pairs: indeed if we duplicate a point as many times as it is involved in a pair, the dataset would be huge. Instead, we would build a object that acts **as if** it was a list of pairs, but in fact stores the initial dataset as well as indices. The behaviour of such an object and use basic use cases can be found in this binder notebook: [Binder](https://mybinder.org/v2/gh/wdevazelhes/api_metric_learn/master).

This PR will be complete when the following will be done: 

- [X] Keep the same API but introduce the objects that will be used in the new API
- [ ] Create an object ConstrainedDataset that will allow the behaviour presented in the description
- [X] Modify existing algorithms to use this object. The API will then be changed (only for algorithms that take as an input an some labeled pairs of samples instead of labeled samples).
- [X] Write the appropriate documentation

### DIFF
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -59,7 +59,7 @@ class SupervisedMetricLearner(BaseMetricLearner):
 
 class WeaklySupervisedMetricLearner(BaseMetricLearner):
 
-  def fit(self, constrained_dataset, y):
+  def fit(self, X_constrained, y):
     return NotImplementedError
 
 
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -3,7 +3,7 @@ from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils.validation import check_array
 
 
-class BaseMetricLearner(BaseEstimator):
+class BaseMetricLearner(BaseEstimator, TransformerMixin):
   def __init__(self):
     raise NotImplementedError('BaseMetricLearner should not be instantiated')
 
@@ -51,7 +51,7 @@ class BaseMetricLearner(BaseEstimator):
     return X.dot(L.T)
 
 
-class SupervisedMetricLearner(BaseMetricLearner, TransformerMixin):
+class SupervisedMetricLearner(BaseMetricLearner):
 
   def fit(self, X, y):
     return NotImplementedError
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -2,14 +2,11 @@ from numpy.linalg import cholesky
 from sklearn.base import BaseEstimator
 from sklearn.utils.validation import check_array
 
-
-class TransformerMixin(object):
-  """Mixin class for all transformers in metric-learn. Same as the one in
-  scikit-learn, but the documentation is changed: this Transformer is
-  allowed to take as y a non array-like input"""
+class BaseMetricLearner(BaseEstimator):
 
   def __init__(self):
-    raise NotImplementedError('TransformerMixin should not be instantiated')
+    raise NotImplementedError('BaseMetricLearner should not be instantiated')
+
 
   def fit_transform(self, X, y=None, **fit_params):
     """Fit to data, then transform it.
@@ -19,7 +16,7 @@ class TransformerMixin(object):
 
     Parameters
     ----------
-    X : numpy array of shape [n_samples, n_features]
+    X : array-like of shape [n_samples, n_features], or ConstrainedDataset
         Training set.
 
     y : numpy array of shape [n_samples] or 4-tuple of arrays
@@ -40,11 +37,6 @@ class TransformerMixin(object):
     else:
       # fit method of arity 2 (supervised transformation)
       return self.fit(X, y, **fit_params).transform(X)
-
-class BaseMetricLearner(BaseEstimator, TransformerMixin):
-
-  def __init__(self):
-    raise NotImplementedError('BaseMetricLearner should not be instantiated')
 
   def metric(self):
     """Computes the Mahalanobis matrix from the transformation matrix.
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -23,7 +23,7 @@ class BaseMetricLearner(BaseEstimator):
     X : array-like of shape [n_samples, n_features], or ConstrainedDataset
         Training set.
 
-    y : numpy array of shape [n_samples] or 4-tuple of arrays
+    y : numpy array of shape [n_samples] or quadruplet of arrays
         Target values, or constraints (a, b, c, d) indices into X, with
         (a, b) specifying similar and (c,d) dissimilar pairs).
 
@@ -70,7 +70,7 @@ class BaseMetricLearner(BaseEstimator):
 
     Parameters
     ----------
-    X : (n x d) matrix or ConstrainedDataset, optional
+    X : (n x d) matrix or `ConstrainedDataset` , optional
         Data to transform. If not supplied, the training data will be used.
         In the case of a ConstrainedDataset, X_constrained.X is used.
 
@@ -113,9 +113,6 @@ class WeaklySupervisedMixin(object):
     raise NotImplementedError('WeaklySupervisedMixin should not be '
                               'instantiated')
 
-  def fit(self, X, constraints, **kwargs):
-    return self._fit(X, constraints, **kwargs)
-
   def decision_function(self, X_constrained):
       return self.predict(X_constrained)
 
@@ -125,15 +122,72 @@ class PairsMixin(WeaklySupervisedMixin):
   def __init__(self):
     raise NotImplementedError('PairsMixin should not be instantiated')
 
+  def fit(self, X_constrained, y_constraints, **kwargs):
+    """Fit a pairs based metric learner.
+
+    Parameters
+    ----------
+    X_constrained : `ConstrainedDataset`, shape=(n_constraints, 2, n_features)
+      Training `ConstrainedDataset`.
+
+    y_constraints : array-like, shape=(n_constraints,)
+      Labels of constraints (0 for similar pairs, 1 for dissimilar).
+
+    kwargs : Any
+      Algorithm specific parameters.
+
+    Returns
+    -------
+    self : The fitted estimator.
+    """
+    return self._fit(X_constrained, y_constraints, **kwargs)
+
   def predict(self, X_constrained):
+    """Predicts the learned similarity between input pairs
+
+    Returns the learned metric value between samples in every pair. It should
+    ideally be low for similar samples and high for dissimilar samples.
+
+    Parameters
+    ----------
+    X_constrained : `ConstrainedDataset`, shape=(n_constraints, 2, n_features)
+      A constrained dataset of paired samples.
+
+    Returns
+    -------
+    y_predicted : `numpy.ndarray` of floats, shape=(n_constraints,)
+      The predicted learned metric value between samples in every pair.
+    """
     # TODO: provide better implementation
     pairwise_diffs = (X_constrained.X[X_constrained.c[:, 0]] -
                       X_constrained.X[X_constrained.c[:, 1]])
     return np.sqrt(np.sum(pairwise_diffs.dot(self.metric()) * pairwise_diffs,
                                   axis=1))
 
-  def score(self, X_constrained, y):
-      return roc_auc_score(y, self.decision_function(X_constrained))
+  def score(self, X_constrained, y_constraints):
+    """Computes score of pairs similarity prediction.
+
+    Returns the ``roc_auc`` score of the fitted metric learner. It is
+    computed in the following way: for every value of a threshold
+    ``t`` we classify all pairs of samples where the predicted distance is
+    inferior to ``t`` as belonging to the "similar" class, and the other as
+    belonging to the "dissimilar" class, and we count false positive and
+    true positives as in a classical ``roc_auc`` curve.
+
+    Parameters
+    ----------
+    X_constrained : `ConstrainedDataset`, shape=(n_constraints, 2, n_features)
+      Constrained dataset of paired samples.
+
+    y_constraints : array-like, shape=(n_constraints,)
+      The corresponding labels.
+
+    Returns
+    -------
+    score : float
+      The ``roc_auc`` score.
+    """
+    return roc_auc_score(y_constraints, self.decision_function(X_constrained))
 
 
 class TripletsMixin(WeaklySupervisedMixin):
@@ -142,7 +196,45 @@ class TripletsMixin(WeaklySupervisedMixin):
     raise NotImplementedError('TripletsMixin should not be '
                               'instantiated')
 
+  def fit(self, X_constrained, y=None, **kwargs):
+    """Fit a triplets based metric learner.
+
+    Parameters
+    ----------
+    X_constrained : `ConstrainedDataset`, shape=(n_constraints, 3, n_features)
+      Training `ConstrainedDataset`. To give the right supervision to the
+      algorithm, the first two points should be more similar than the first
+      and the third.
+
+    y : Ignored, for scikit-learn compatibility.
+
+    kwargs : Any
+      Algorithm specific parameters.
+
+    Returns
+    -------
+    self : The fitted estimator.
+    """
+    return self._fit(X_constrained, **kwargs)
+
+
   def predict(self, X_constrained):
+    """Predict the difference between samples similarities in input triplets.
+
+    For each triplet of samples in ``X_constrained``, returns the
+    difference between the learned similarity between the first and the
+    second point, minus the similarity between the first and the third point.
+
+    Parameters
+    ----------
+    X_constrained : `ConstrainedDataset`, shape=(n_constraints, 3, n_features)
+      Input constrained dataset.
+
+    Returns
+    -------
+    prediction : `numpy.ndarray` of floats, shape=(n_constraints,)
+      Predictions for each triplet.
+    """
     # TODO: provide better implementation
     similar_diffs = X_constrained.X[X_constrained.c[:, 0]] - \
                     X_constrained.X[X_constrained.c[:, 1]]
@@ -154,6 +246,24 @@ class TripletsMixin(WeaklySupervisedMixin):
                           dissimilar_diffs, axis=1))
 
   def score(self, X_constrained, y=None):
+    """Computes score of triplets similarity prediction.
+
+    Returns the accuracy score of the following classification task: a record
+    is correctly classified if the predicted similarity between the first two
+    samples is higher than that between the first and the third.
+
+    Parameters
+    ----------
+    X_constrained : `ConstrainedDataset`, shape=(n_constraints, 3, n_features)
+      Constrained dataset of triplets of samples.
+
+    y: Ignored (for scikit-learn compatibility).
+
+    Returns
+    -------
+    score: float
+      The triplets score.
+    """
     predicted_sign = self.decision_function(X_constrained) < 0
     return np.sum(predicted_sign) / predicted_sign.shape[0]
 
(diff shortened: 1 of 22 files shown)


metric_learn/covariance.py
```diff
@@ -12,10 +12,10 @@
 import numpy as np
 from sklearn.utils.validation import check_array
 
-from .base_metric import BaseMetricLearner
+from .base_metric import SupervisedMetricLearner
 
 
-class Covariance(BaseMetricLearner):
+class Covariance(SupervisedMetricLearner):
```
perimosocordiae: I understand why this was chosen, but this particular base class made me stop and consider a moment. We may eventually want a base class for unsupervised methods as well.

bellet: I agree, this simple covariance method should not be a SupervisedMetricLearner (as it is completely unsupervised). Whether we will really need an unsupervised class in the long run is unclear, but maybe the best for now is to create an UnsupervisedMetricLearner class which takes only X in fit.

wdevazelhes: Thanks for the review ! I agree, I did not notice but indeed Covariance is unsupervised, so I will change this in a following PR

wdevazelhes:
> 1/ It may be useful to add ways to update a ConstraintDataset object, although I am not sure how far we should go (append new constraints? append new points in X? ...). I think it would probably be useful to be able to add new constraint indices in c without modifying X, and maybe to merge two ConstrainedDataset objects (by concatenating the two X's, and the two constraints after re-indexing). A potential use case to keep in mind is to create constraints from a small set of labeled points and then complement this with constraints on a different/larger set of points.
 
I agree, I will implement it in a next PR

> 2/ Do we want to consider a ConstrainedDataset object to be characterized primarily by (a) a fixed dataset X, along with some constraints on X, or (b) by a set of constraints with X giving a memory-efficient representation? In (b), one could remove points in X which are not used in the constraints (at creation of the object, or when slicing etc). I think (a) makes more sense, and probably simplifies the implementation of stuff in 1/


Yes indeed, while in the original notebook (b) is implemented, (a) (the solution of this PR) has a lot of advantages: it is simpler (I agree it will simplify 1/ implementation), also the behaviour is consistent between creation and slicing (in the original notebook X was pruned at slicing but we had to  had to keep the full X at creation in the case of pipelining transformers, (ex: ``make_pipeline(MetricLearner(), PCA())``) because they will be applied to points and not constraints). Plus, in case of (b), if we want to prune X when slicing, in fact we will often have to copy the samples X that we keep (except in the case where slicing on constraints gives a real slicing on X (ex: ``X[5, 6, 7, 8]`` that we could transform in ``X[5:9]`` to have a view)): indeed we cannot have a view of X except if we take a real slice of it (fancy indexing of a numpy array returns a copy). Therefore, (a) allows to keep only the reference to an initial dataset, which is memory efficient.

metric_learn/constraints.py
```diff
@@ -100,3 +101,57 @@ def random_subset(all_labels, num_preserved=np.inf, random_state=np.random):
...
+    return self.toarray().__str__()
+
+  def __repr__(self):
+    return self.toarray().__repr__()
+
+  def toarray(self):
+    return self.X[self.c]
+
+  @staticmethod
+  def _check_index(length, indices):
```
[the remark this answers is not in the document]
wdevazelhes: I agree, I will implement it in a next PR

wdevazelhes:
I just commited some API changes in the algorithms, making the Weakly Supervised algorithms use ConstrainedDataset and y as an input in ``fit``. For now I just wrapped the previous behaviour inside ``fit`` with some helper functions (``unwrap_pairs``, ``wrap_pairs``, ``unwrap_to_graph``) as a proof of concept, and made the tests work, but I will probably make a cleaner implementation later on.

metric_learn/base_metric.py
```diff
@@ -59,7 +59,7 @@ def fit(self, X, y):
 
 class WeaklySupervisedMetricLearner(BaseMetricLearner):
 
-  def fit(self, X, constraints):
+  def fit(self, constrained_dataset, y):
```
bellet: maybe think of a more concise naming convention instead of `constrained_dataset` as it is gonna be used all over the place. `X_constrained` perhaps?

metric_learn/itml.py
```diff
@@ -73,19 +73,19 @@ def _process_inputs(self, X, constraints, bounds):
...
     Parameters
     ----------
-    X : (n x d) data matrix
-        each row corresponds to a single instance
-    constraints : 4-tuple of arrays
-        (a,b,c,d) indices into X, with (a,b) specifying positive and (c,d)
-        negative pairs
+    constrained_dataset : ConstrainedDataset
+        with constraints being an array of shape [n_constraints, 2]
+    y : array-like, shape (n x 1)
```
bellet: should be `n_constraints` instead of `n`

metric_learn/mmc.py
```diff
@@ -58,17 +59,17 @@ def __init__(self, max_iter=100, max_proj=10000, convergence_threshold=1e-3,
...
     Parameters
     ----------
-    X : (n x d) data matrix
-        each row corresponds to a single instance
-    constraints : 4-tuple of arrays
-        (a,b,c,d) indices into X, with (a,b) specifying similar and (c,d)
-        dissimilar pairs
+    constrained_dataset : ConstrainedDataset
+        with constraints being an array of shape [n_constraints, 2]
+    y : array-like, shape (n x 1)
```
bellet: same as above

metric_learn/sdml.py
```diff
@@ -56,21 +56,22 @@ def _prepare_inputs(self, X, W):
...
 
     Parameters
     ----------
-    X : array-like, shape (n, d)
-        data matrix, where each row corresponds to a single instance
-    W : array-like, shape (n, n)
-        connectivity graph, with +1 for positive pairs and -1 for negative
+    constrained_dataset : ConstrainedDataset
+        with constraints being an array of shape [n_constraints, 2]
+    y : array-like, shape (n x 1)
```
bellet: same as above

metric_learn/base_metric.py
wdevazelhes: I agree, ``X_constrained`` seems good

metric_learn/itml.py
wdevazelhes: Yes, indeed, will change this

metric_learn/mmc.py
wdevazelhes: will do

metric_learn/sdml.py
wdevazelhes: will do

metric_learn/lsml.py
```diff
@@ -181,4 +186,7 @@ def fit(self, X, y, random_state=np.random):
                                   random_state=random_state)
     pairs = c.positive_negative_pairs(num_constraints, same_length=True,
                                       random_state=random_state)
-    return LSML.fit(self, X, pairs, weights=self.weights)
+    X_constrained = ConstrainedDataset(X, np.hstack([pairs[i][:, None]
+                                                           for i in
+                                                           range(4)]))
```
perimosocordiae: `np.column_stack(pairs)` seems to be what you want here.

metric_learn/mmc.py
```diff
@@ -22,13 +22,14 @@
 from sklearn.metrics import pairwise_distances
 from sklearn.utils.validation import check_array, check_X_y
 
-from .base_metric import BaseMetricLearner
-from .constraints import Constraints
+from .base_metric import PairsMetricLearner, SupervisedMetricLearner
+from .constraints import Constraints, ConstrainedDataset, unwrap_pairs, \
+  wrap_pairs
```
perimosocordiae: Style nitpick: Use parens to break up long import lists:

```python
from .constraints import (
    Constraints, ConstrainedDataset, unwrap_pairs, wrap_pairs)
```

test/test_constrained_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+        binary_index = np.random.randint(0, 2, c_shape)
+        boolean_index = binary_index.astype(bool)
+        items = [0, c_shape - 1, i, slice(i), slice(0, begin), slice(begin,
+                 end), slice(end, c_shape), slice(0, c_shape), fancy_index,
+                 binary_index, boolean_index]
+        for item in items:
+            self.check_indexing(item)
+
+    def test_repr(self):
+        assert repr(cd) == repr(X[c])
```
perimosocordiae: Use `self.assertEqual(...)` instead of bare asserts.

metric_learn/lsml.py
wdevazelhes: Great, thanks !

test/test_constrained_dataset.py
wdevazelhes: Thanks !

metric_learn/mmc.py
wdevazelhes: Thanks !

test/test_constrained_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @staticmethod
+    def check_indexing(idx):
+        # checks that an indexing returns the data we expect
+        np.testing.assert_array_equal(cd[idx].c, c[idx])
+        np.testing.assert_array_equal(cd[idx].toarray(), X[c[idx]])
+        np.testing.assert_array_equal(cd[idx].toarray(), X[c][idx])
+
+    def test_inputs(self):
+        # test the allowed and forbidden ways to create a ConstrainedDataset
```
perimosocordiae: I'd split this into two separate tests, `test_allowed_inputs` and `test_invalid_inputs`.

test/test_constrained_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+        ConstrainedDataset(X, c)
+        two_points = [[1, 2], [3, 5]]
+        out_of_range_constraints = [[1, 2], [0, 1]]
+        msg = "ConstrainedDataset cannot be created: the length of " \
+              "the dataset is 2, so index 2 is out of " \
+              "range."
```
perimosocordiae: Similar to imports, use parentheses here:
```python
msg = ("First part of string. "
       "Second part of string.")
```

test/test_constrained_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+from sklearn.model_selection import StratifiedKFold, KFold
+from sklearn.utils.testing import assert_raise_message
+
+X = np.random.randn(20, 5)
+c = np.random.randint(0, X.shape[0], (15, 2))
+cd = ConstrainedDataset(X, c)
+y = np.random.randint(0, 2, c.shape[0])
+group = np.random.randint(0, 3, c.shape[0])
+
+c_shape = c.shape[0]
```
perimosocordiae: This is only used by `test_getitem`, so I'd move it there. (And probably name it `num_constraints`)

metric_learn/constraints.py
```diff
@@ -100,3 +90,88 @@ def random_subset(all_labels, num_preserved=np.inf, random_state=np.random):
...
+
+  @staticmethod
+  def triplets_from_labels(y):
+    # TODO: to be implemented
+    raise NotImplementedError
+
+
+def unwrap_pairs(X_constrained, y):
+  a = X_constrained.c[(y == 0)[:, 0]][:, 0]
+  b = X_constrained.c[(y == 0)[:, 0]][:, 1]
```
perimosocordiae: This seems redundant. How about:
```python
y_zero = (y == 0).ravel()
a, b = X_constrained[y_zero].T
c, d = X_constrained[~y_zero].T
```

metric_learn/constraints.py
```diff
@@ -100,3 +90,88 @@ def random_subset(all_labels, num_preserved=np.inf, random_state=np.random):
...
+  X = X_constrained.X
+  return X, [a, b, c, d]
+
+def wrap_pairs(X, constraints):
+  a = np.array(constraints[0])
+  b = np.array(constraints[1])
+  c = np.array(constraints[2])
+  d = np.array(constraints[3])
+  constraints = np.vstack([np.hstack([a[:, None], b[:, None]]),
+                           np.hstack([c[:, None], d[:, None]])])
```
perimosocordiae:
```python
a, b, c, d = constraints
constraints = np.vstack((np.column_stack((a, b)), np.column_stack((c, d))))
# or if we have numpy 1.13+
constraints = np.block([[a, b], [c, d]])
```

test/test_constrained_dataset.py
```diff
@@ -0,0 +1,74 @@
+import unittest
+import numpy as np
+from metric_learn.constraints import ConstrainedDataset
+from numpy.testing import assert_array_equal
+from sklearn.model_selection import StratifiedKFold, KFold
+from sklearn.utils.testing import assert_raise_message
+
+X = np.random.randn(20, 5)
```
perimosocordiae: It would be nice to test sparse `X` as well. To do so, I'd write something like this:
```python
class _BaseTestConstrainedDataset(unittest.TestCase):
   # everything currently under TestConstrainedDataset, but using self.X instead of X

class TestDenseConstrainedDataset(_BaseTestConstrainedDataset):
  def setUp(self):
    self.X = np.random.randn(20, 5)
    self.c = ... # and so on

class TestSparseConstrainedDataset(_BaseTestConstrainedDataset):
   # similar, but setUp creates a dense X
```

test/test_constrained_dataset.py
bellet: actually this is also used in `test_shape` (although not through this variable)

test/test_constrained_dataset.py
bellet: maybe it would make sense to also test other data types? like lists instead of numpy arrays

test/test_constrained_dataset.py
wdevazelhes: I agree, will do

test/test_constrained_dataset.py
wdevazelhes: thanks, will do

metric_learn/constraints.py
wdevazelhes: thanks, will do

metric_learn/constraints.py
wdevazelhes: thanks, will do

test/test_constrained_dataset.py
wdevazelhes: I agree, I will add these tests in a next commit

test/test_constrained_dataset.py
wdevazelhes: I agree I will rename it and put it before the definitions that use it

wdevazelhes:
Hi, I just pushed a modification of the class structure. As discussed with @nvauquie and @bellet, the following fact was problematic: ``MMC_Supervised`` for instance, inherited from both ``SupervisedMetricLearner`` (through ``MMC`` inheritance), and ``WeaklySupervisedMetricLearner``. This was weird and could also cause some bugs for instance if I want to put an abstract  ``predict`` in ``WeaklySupervisedMetricLearning``. Indeed, when testing ``MMC_Supervised`` with scikit-learn's check-estimator, scikit-learn will list all the functions that are implemented by ``MMC_Supervised``, and as it implements ``predict`` (through ``WeaklySupervisedMetricLearner`` inheritance), it would run some tests on it with synthetic a dataset. But the ``predict`` function is not a real scikit-learn's one as it takes as input a ``ConstrainedDataset`` and not numpy array so it will bug. 
One way to solve this problem is to replace the class hierarchy with mixins, like it is the case in many of scikit-learn's estimators: see for instance ``KNeighborsClassifier`` https://github.com/scikit-learn/scikit-learn/blob/788a458bba353c2cf3cfa5a15d6f68315149ef9e/sklearn/neighbors/classification.py#L23 and ``KNeighborsRegressor`` https://github.com/scikit-learn/scikit-learn/blob./788a458bba353c2cf3cfa5a15d6f68315149ef9e/sklearn/neighbors/regression.py#L19 Also the same is done for ``LinearRegression``: https://github.com/scikit-learn/scikit-learn/blob/788a458bba353c2cf3cfa5a15d6f68315149ef9e/sklearn/linear_model/logistic.py#L952 https://github.com/scikit-learn/scikit-learn/blob/788a458bba353c2cf3cfa5a15d6f68315149ef9e/sklearn/linear_model/base.py#L284
For now this architecture will allow to develop the next features (like the ``predict`` function in weakly supervised metric learners), but it can be changed later.

metric_learn/base_metric.py
```diff
@@ -1,8 +1,43 @@
-from numpy.linalg import inv, cholesky
-from sklearn.base import BaseEstimator, TransformerMixin
+from numpy.linalg import cholesky
+from sklearn.base import BaseEstimator
 from sklearn.utils.validation import check_array
 
 
+class TransformerMixin(object):
```
perimosocordiae: Does any class other than `BaseMetricLearner` use this mixin? If not, I'd just inline it into `BaseMetricLearner`.

metric_learn/base_metric.py
```diff
@@ -49,3 +84,45 @@ def transform(self, X=None):
...
+
+class TripletsMixin(WeaklySupervisedMixin):
+
+  def __init__(self):
+    raise NotImplementedError('TripletsMixin should not be '
+                              'instantiated')
+  # TODO: introduce specific scoring functions etc
+
+
+class QuadrupletsMixin(UnsupervisedMixin):
```
perimosocordiae: Quadruplets imply weak supervision.

metric_learn/base_metric.py
```diff
@@ -49,3 +84,45 @@ def transform(self, X=None):
...
+    return NotImplementedError
+
+
+class UnsupervisedMixin(object):
+
+  def fit(self, X, y=None):
+    return NotImplementedError
+
+
+class WeaklySupervisedMixin(object):
```
perimosocordiae: Add `NotImplementedError` for the `__init__` methods of this and the above abstract mixins.

metric_learn/covariance.py
```diff
@@ -34,3 +34,8 @@ def fit(self, X, y=None):
     else:
       self.M_ = np.linalg.inv(self.M_)
     return self
+
+class Covariance(_Covariance, UnsupervisedMixin):
```
perimosocordiae: If nothing else will use `_Covariance`, I'd prefer to eliminate it and use the `UnsupervisedMixin` directly.

perimosocordiae: In general, I don't think it's worth separating out the algorithm implementation unless/until we have >1 class that needs it.

metric_learn/base_metric.py
wdevazelhes: I agree
Done

metric_learn/base_metric.py
wdevazelhes: Done

metric_learn/base_metric.py
wdevazelhes: Done

metric_learn/base_metric.py
```diff
@@ -1,12 +1,43 @@
...
+
+    Fits transformer to X and y with optional parameters fit_params
+    and returns a transformed version of X.
+
+    Parameters
+    ----------
+    X : array-like of shape [n_samples, n_features], or ConstrainedDataset
+        Training set.
+
+    y : numpy array of shape [n_samples] or 4-tuple of arrays
```
bellet: with the new API `y` should only be constraint labels or `None`, shouldn't it?

wdevazelhes:
I just pushed a commit that allows to do predictions and decision functions with weakly supervised learners. I added some testing but we cannot do ``check_estimator`` on them (as they are training on ConstrainedDatasets rather than array-like data). So I tried to take a relevant subset of scikit-learn's tests that are runned by ``check_estimator``, and adapt it to WeaklySupervisedMetricLearners. Some are maybe too much now, but I thought they might be useful later when developing new algorithms. Let me know if they add some confusion to the project and should rather be removed.
Note that some tests fail for now due to numerical errors, but they will be fixed in the future (I think the synthetic random dataset I used for testing is not really learnable so I will replace it by a better one).

In another branch, I am also trying to use scikit-learn's ``check_estimator`` anyway, trying to wrap weakly supervised metric learners into estimators compatible with ``check_estimator`` (but not through supervised metric learners, which is already tested in ``test_sklearn_compat``). This would allow to test the prediction, scoring etc behaviour of weakly supervised metric learners, through scikit-learn's ``check_estimator``. However, this is not easy because ``check_estimator`` routines launch a serie of tests with classical array-like data as inputs, so I am trying to do a wrapper estimator that create some synthetic ConstrainedDataset from an array-like input, and forcing the y to be boolean in order to look like the label of a pair of constraints. I am not sure about this ... The branch is here https://github.com/wdevazelhes/metric-learn/commits/feat/predictions_more_tests

metric_learn/base_metric.py
wdevazelhes: Thanks, indeed I forgot to update the 4-tuple part 
Originally I wanted to allow ``fit_transform`` to be used in a classical way (so with labels as y) but also in the weakly supervised way (with constraints labels) (and it could be None in both cases). So something like     
```python
y : None or numpy array of shape [n_samples] or [n_constraints]
        Target values, or constraints labels
```
However now I am not sure it it a good idea: maybe we should make this ``fit_transform`` specific to weakly supervised learning (so with ``y`` being constraints labels or None as you say) and inline it into ``WeaklySupervisedMixin`` rather than ``BaseEstimator``, and make ``SupervisedMixin`` inherit from scikit-learn's ``TransformerMixin`` to have the classical version of ``fit_transform`` for supervised estimators, like this: 

```python
class WeaklySupervisedMixin(object):
     def fit_transform(self, X_constrained, y=None, **fit_params):
    """
[...]
    y : None, or numpy array of shape [n_constraints]
        Constraints labels.
    """
class SupervisedMixin(TransformerMixin):
```

wdevazelhes:
Hi, I just added some documentation. Most of it is in the Weakly Supervised page, but there is also some in the docstrings of ``ConstrainedDataset`` and that of ``PairsMixin``, ``TripletsMixin`` and ``QuadrupletsMixin``.

bellet:
For clarity I will close this one since it is about a previous API attempt

PR #89, opened 2018-04-27
[MRG] uniformize indent to 4 spaces

wdevazelhes:
This PR reindents all files with 4 indents, according to PEP8 style guide (http://pep8.readthedocs.io/en/release-1.7.x/intro.html#error-codes)

### DIFF
--- a/bench/benchmarks/iris.py
+++ b/bench/benchmarks/iris.py
@@ -17,22 +17,22 @@ CLASSES = {
 }
 
 try:
-  from metric_learn.lmnn import python_LMNN
-  if python_LMNN is not metric_learn.LMNN:
-    CLASSES['python_LMNN'] = python_LMNN(k=5, learn_rate=1e-6, verbose=False)
+    from metric_learn.lmnn import python_LMNN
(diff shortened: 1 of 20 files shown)


wdevazelhes:
We discussed about this with @bellet and @nvauquie and it appeared that for instance in scikit-learn they are more permissive than a full compliance with automatic PEP8 checkers: indeed they distinguish 3 PEP8 issues: those that should be fixed (like mixing spaces and tabs), those that should be fixed if this part of the code is touched, and those where they do not care (see https://github.com/scikit-learn/scikit-learn/issues/4043#issuecomment-69101349). (See also https://github.com/scikit-learn/scikit-learn/issues/5816#issuecomment-156722866)
So maybe we could follow that policy ? Like just checking newly modified code in PRs and check that it does not introduce new PEP8 issues ? (And we would consider that the indent should be 2 spaces)
But it also appeared that as there are a few PEP8 issues at the moment in the project, it could be the occasion to fix them at once. 
Anyway I agree that it is not so important and consistency is the more important.

PR #90, opened 2018-04-27
[MRG] move from unittest to pytest

wdevazelhes:
This PR replaces unittest by pytest. Pytest is more and more used and provides a flexible and advanced framework for testing. It is still compatible with unittests so the project testing would still work after this PR is merged. It would allow to code future tests for the new API in pytest.


wdevazelhes:
Yes I agree, it is not necessary to update existing tests (they might use pytest later but only if needed). I just updated the PR message.
Thanks for the README.txt, I'll update it.

Issue #91, opened 2018-05-17
New API to be more compatible with scikit-learn


wdevazelhes:
Hi, this is the new proposal for metric-learn's API. 

The previously proposed API (https://github.com/metric-learn/metric-learn/pull/85) , used a custom data object (ConstrainedDataset), but this is to avoid. This new API proposal, on the contrary, finds a way to use only usual data format as inputs of algorithms (arrays etc).

The main goal of this new API, like the previous attempt, is to change the input data of metric learners that learn on tuple of points (pairs + labels of pairs, triplets, quadruplets), to be compatible with scikit-learn's utilities like cross-validation. 

## UML Diagram 
Here is the current UML diagram of classes

<details>

![uml diagram](http://yuml.me/46bb7842.jpg)

</details>

## Quick overview
<details>
Let's consider the case of a metric learning algorithm fitting on tuples of points. The data points X (or a lazy equivalent like the path to a folder on disk) would be given first as an argument to a Transformer object that would memorize it. 

Then, when creating a TupleLearner (let's call this way a Metric Learner that learns and predicts on tuple of points), we would give as an argument the previous Transformer. 

Then the TupleLearner would be able to take as input an array of indices (of shape (n_constraints, 2) for PairsLearner for instance), and under the hood form tuples from these input indices plus the Transformer.

 Therefore the input to TuplesLearner would be a unique object (the array of tuples), that can be splitted (slicing in two the dataset of tuples along the first dimension slices it in two other dataset of tuples ), which could therefore stand for the usual X in scikit learn, that allows cross validation etc.

A short snippet to sum up this main use case: 

```python
class ArrayIndexer(TransformerMixin):
  def __init__(self, X):
    self.X = X
  def transform(self, list_of_indices):
    return self.X[list_of_indices]
# ArrayIndexer would be an object coded in metric-learn

pairs_extractor = ArrayIndexer(X) # X: data, ArrayIndexer: the aforementioned Transformer
mmc = MMC(preprocessor=pairs_extractor)
mmc.fit(pairs, y) # pairs: indices of pairs (2D array of shape (n_constraints, 2) of integers), y: label of constraints (1 if similar, 0 if not) (array of shape (n_constraints)) 
```
</details>

## More details :
Here are more details on the API, including more cases that we would want to deal with: 
<details>

[...]

PR #92, opened 2018-05-17
[MRG + 1] Allow already formed tuples as an input.

wdevazelhes:
This PR changes the API with the first and easier case to implement of the new api (see issue https://github.com/metric-learn/metric-learn/issues/91 ): the case where there is no preprocessor but tuples are already formed and given to the algorithm as such. The main question is for algorithms that needs a covariance matrix to initialize the metric (LSML), or that compute bounds (ITML). For now I just reconstruct a dataset of points from the pairs thanks to np.unique, in order to have the same results and pass the tests, but this is probably not the best thing to do. For SDML, I could express the term involved in the loss only with formed pairs instead of datapoints, and here is a snippet that tests it: https://gist.github.com/wdevazelhes/3c349e13976613d15ebc46178c942474

Left TODO:
- [x] Update docstrings of methods where signature changed

### DIFF
--- a/metric_learn/itml.py
+++ b/metric_learn/itml.py
@@ -54,10 +54,9 @@ class ITML(BaseMetricLearner):
   def _process_pairs(self, pairs, y, bounds):
     pairs, y = check_X_y(pairs, y, accept_sparse=False,
                                       ensure_2d=False, allow_nd=True)
-    y = y.astype(bool)
 
     # check to make sure that no two constrained vectors are identical
-    pos_pairs, neg_pairs = pairs[y], pairs[~y]
+    pos_pairs, neg_pairs = pairs[y == 1], pairs[y == -1]
     pos_no_ident = vector_norm(pos_pairs[:, 0, :] - pos_pairs[:, 1, :]) > 1e-9
     pos_pairs = pos_pairs[pos_no_ident]
     neg_no_ident = vector_norm(neg_pairs[:, 0, :] - neg_pairs[:, 1, :]) > 1e-9
@@ -76,8 +75,7 @@ class ITML(BaseMetricLearner):
     else:
       self.A_ = check_array(self.A0)
     pairs = np.vstack([pos_pairs, neg_pairs])
-    y = np.hstack([np.ones(len(pos_pairs)), np.zeros(len(neg_pairs))])
-    y = y.astype(bool)
+    y = np.hstack([np.ones(len(pos_pairs)), - np.ones(len(neg_pairs))])
     return pairs, y
 
 
@@ -100,7 +98,7 @@ class ITML(BaseMetricLearner):
     """
     pairs, y = self._process_pairs(pairs, y, bounds)
     gamma = self.gamma
-    pos_pairs, neg_pairs = pairs[y], pairs[~y]
+    pos_pairs, neg_pairs = pairs[y == 1], pairs[y == -1]
     num_pos = len(pos_pairs)
     num_neg = len(neg_pairs)
     _lambda = np.zeros(num_pos + num_neg)
--- a/metric_learn/itml.py
+++ b/metric_learn/itml.py
@@ -87,7 +87,7 @@ class ITML(BaseMetricLearner):
     pairs: array-like, shape=(n_constraints, 2, n_features)
         Array of pairs. Each row corresponds to two points.
     y: array-like, of shape (n_constraints,)
-        Labels of constraints. Should be 0 for dissimilar pair, 1 for similar.
+        Labels of constraints. Should be -1 for dissimilar pair, 1 for similar.
     bounds : list (pos,neg) pairs, optional
         bounds on similarity, s.t. d(X[a],X[b]) < pos and d(X[c],X[d]) > neg
 
--- a/metric_learn/itml.py
+++ b/metric_learn/itml.py
@@ -64,7 +64,7 @@ class ITML(BaseMetricLearner):
     neg_pairs = neg_pairs[neg_no_ident]
     # init bounds
     if bounds is None:
-      X = np.unique(pairs.reshape(-1, pairs.shape[2]), axis=0)
+      X = np.vstack({tuple(row) for row in pairs.reshape(-1, pairs.shape[2])})
       self.bounds_ = np.percentile(pairwise_distances(X), (5, 95))
     else:
       assert len(bounds) == 2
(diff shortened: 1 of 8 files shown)


metric_learn/sdml.py
```diff
@@ -42,21 +42,22 @@ def __init__(self, balance_param=0.5, sparsity_param=0.01, use_cov=True,
...
 
-  def _prepare_inputs(self, X, W):
-    self.X_ = X = check_array(X)
-    W = check_array(W, accept_sparse=True)
+  def _prepare_pairs(self, pairs, y):
+    pairs, y = check_X_y(pairs, y, accept_sparse=False,
+                                      ensure_2d=False, allow_nd=True)
     # set up prior M
     if self.use_cov:
+      X = np.unique(pairs.reshape(-1, pairs.shape[2]), axis=0)
```
perimosocordiae: This is causing test failures, because the `axis` argument for `np.unique` was added in version 1.13.0 but we test with an earlier version.

wdevazelhes: Yes, indeed, I will change this to something compatible with version 1.12.1

wdevazelhes: I'll go for this solution: https://stackoverflow.com/questions/16970982/find-unique-rows-in-numpy-array/22941699#22941699

wdevazelhes:
I also updated the docstring. Note also that for quadruplets learning, for now I asked for the quadruplets to be ordered so that we know which samples are more similar to others. We could also imagine to let the ordering between the first two and the last two pairs be any (most similar pairs before the least similar, or the contrary), and then use a label to specify the ordering. This would be coherent with the supervision for pairs (where there is also a label of constraints), but is redundant. However this can be easily changed later.

metric_learn/itml.py
```diff
@@ -51,52 +51,65 @@ def __init__(self, gamma=1., max_iter=1000, convergence_threshold=1e-3,
     self.A0 = A0
     self.verbose = verbose
 
-  def _process_inputs(self, X, constraints, bounds):
-    self.X_ = X = check_array(X)
+  def _process_pairs(self, pairs, y, bounds):
```
bellet: I think it would make sense that this function `_process_pairs` is shared across the class of pair metric learners. For instance, ruling the potential pairs that are identical is useful for all algorithms

metric_learn/itml.py
```diff
@@ -51,52 +51,65 @@ def __init__(self, gamma=1., max_iter=1000, convergence_threshold=1e-3,
...
-    a,b,c,d = constraints
-    no_ident = vector_norm(X[a] - X[b]) > 1e-9
-    a, b = a[no_ident], b[no_ident]
-    no_ident = vector_norm(X[c] - X[d]) > 1e-9
-    c, d = c[no_ident], d[no_ident]
+    pos_pairs, neg_pairs = pairs[y], pairs[~y]
+    pos_no_ident = vector_norm(pos_pairs[:, 0, :] - pos_pairs[:, 1, :]) > 1e-9
+    pos_pairs = pos_pairs[pos_no_ident]
+    neg_no_ident = vector_norm(neg_pairs[:, 0, :] - neg_pairs[:, 1, :]) > 1e-9
+    neg_pairs = neg_pairs[neg_no_ident]
```
bellet: maybe showing a warning to the user when such pair is found and discarded is useful. in particular if a negative pair is made of two identical points, probably there is a problem with the way the user generated the pairs, or the dataset

metric_learn/sdml.py
```diff
@@ -131,5 +132,8 @@ def fit(self, X, y, random_state=np.random):
 
     c = Constraints.random_subset(y, self.num_labeled,
                                   random_state=random_state)
-    adj = c.adjacency_matrix(num_constraints, random_state=random_state)
-    return SDML.fit(self, X, adj)
+    pos_neg = c.positive_negative_pairs(num_constraints,
+                                              random_state=random_state)
+    pairs, y = wrap_pairs(X, pos_neg)
+    y = 2 * y - 1
```
bellet: shouldn't the pair labels be 0/1 and not -1/1?

that said, this raises the question of whether pair labels should be 0/1 or -1/1. If -1/1 is always more convenient for the algorithm implementation, we could switch to that. potentially allow both and convert to -1/1 in the process pairs helper function

metric_learn/mmc.py
```diff
@@ -58,51 +58,58 @@ def __init__(self, max_iter=100, max_proj=10000, convergence_threshold=1e-3,
...
     ----------
-    X : (n x d) data matrix
-        each row corresponds to a single instance
-    constraints : 4-tuple of arrays
-        (a,b,c,d) indices into X, with (a,b) specifying similar and (c,d)
-        dissimilar pairs
+    pairs: array-like, shape=(n_constraints, 2, n_features)
+        Array of pairs. Each row corresponds to two points.
+    y: array-like, of shape (n_constraints,)
+        Labels of constraints. Should be 0 for dissimilar pair, 1 for similar.
```
bellet: missing changes for label -1/1
maybe make sure to check that no change has been missed, including within the algorithms and tests

wdevazelhes: Thanks indeed I forgot docstrings :p

metric_learn/sdml.py
wdevazelhes: Yes I agree
Done (see commits https://github.com/metric-learn/metric-learn/pull/92/commits/374a8512ff13b8c427c03ed03a36a782fa8ef55c and https://github.com/metric-learn/metric-learn/pull/92/commits/b4bdec4bf30875253e6a60e8f6c22863daa2bf7d). (Note that for now I did not try to simplify the code of the algorithms using these labels but I just added it in the new API issue main message)

metric_learn/itml.py
wdevazelhes: Yes I agree, I added it to the small features TODO list at the end of the main issue: https://github.com/metric-learn/metric-learn/issues/91#user-content-featurestodo

metric_learn/itml.py
wdevazelhes: Yes I agree. I added it to the TODO

wdevazelhes:
Thanks for the review. I am merging this PR. This will allow to review the next PR more easily. If some change remains to do, they can be changed later on, as we are working on a branch separate from master.

PR #95, opened 2018-05-22
[MRG] New API should allow prediction functions and scoring

wdevazelhes:
The new API should allow metric learning algorithms that fit on tuples of points to also predict, score, etc on tuples, Therefore being usable in scikit-learn's cross-validation routines. This is part of the PRs need for issue #91.

- [x] Take previous tests from https://github.com/metric-learn/metric-learn/pull/85, and refactor them using pytest, and to allow already formed tuples (3D arrays) instead of ``ConstrainedDatasets``.
- [x] Take previous code from https://github.com/metric-learn/metric-learn/pull/85 and adapt it to formed tuples
- [x] Make tests work
- [x] Make some modifications if needed
    - <strike>Move docstrings from ``_fit`` to ``fit``</strike>
    - <strike>Remove unused imports</strike>

PR #96, opened 2018-05-25
[MRG] Create new Mahalanobis mixin

wdevazelhes:
This PR creates a new Mahalanobis Mixin (cf issue https://github.com/metric-learn/metric-learn/issues/91), which is a common interface for all algorithms that learn a Mahalanobis type (pseudo) distance (of the form `(x - x')^T M (x - x')`) (right now all algorithms are this form but there might be others in the future). 

This interface will enforce that an attribute `metric_` exists, add documentation for it in the docstring of child classes, and will allow to factorize computations of `embed` functions (similar to what is done now with `transform`), and `score_pairs` function (these functions will come in later PRs, therefore right now this Mixin seems a bit artificial but it is temporary).

I also used the opportunity of this PR to improve the way the argument `metric_` is returned, checking that the matrix indeed exists (i.e. it has been explicitely initialized or the estimator has been fitted), and raising a warning otherwise. Don't hesitate to comment on this last part, or to tell me if it should belong to a separate PR. 

TODO: 
- [x] Create the class 
- [x] Make current algorithms inherit from it
- [x] Use this opportunity to improve the ``metric_`` property
- [x] Maybe add some more tests
- [X] <strike> Fix docstrings to render nicely (as if metric_ was a regular ``Attribute`` of the class) </strike> done by copying, see https://github.com/metric-learn/metric-learn/pull/96#issuecomment-415036218
- [x] Check array at predict, embed etc, to only allow arrays of the right shape =>EDIT: full checking will be done in the "preprocessor" PR

EDIT: 
Initially we were thinking of doing also an ``ExplicitMixin`` that would be for metric learning algorithms that have a way to embed points in a space where the metric is the learned one. Since all algorithms are of this form for now, we will not implement it but rather implement all the functions in ``MahalanobisMixin`` (see https://github.com/metric-learn/metric-learn/pull/95#issuecomment-394689505) 
- <strike> [ ] Add ``embed`` function </strike> => EDIT: for now we will let only a transform function (see https://github.com/metric-learn/metric-learn/pull/96#issuecomment-407118297)
- [x] Add ``score_pairs`` function

### DIFF
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -38,7 +38,7 @@ def check_tuples(tuples):
       "Expected 3D array, got 1D array instead:\ntuples={}.\n"
       "Reshape your data using tuples.reshape(1, -1, 1) if it contains a "
       "single tuple and the points in the tuple have a single "
-      "feature.").format(tuples)
+      "feature.".format(tuples))
   # If input is 2D raise error
   if len(tuples.shape) == 2:
     raise ValueError(
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -28,7 +28,7 @@ def check_tuples(tuples):
     The validated input.
   """
   # If input is scalar raise error
-  if len(tuples.shape) == 0:
+  if np.isscalar(tuples):
     raise ValueError(
       "Expected 3D array, got scalar instead. Cannot apply this function on "
       "scalars.")
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -10,3 +10,41 @@ except TypeError:
 else:
   def vector_norm(X):
     return np.linalg.norm(X, axis=1)
+
+
+def check_tuples(tuples):
+  """Check that the input is a valid 3D array representing a dataset of tuples.
+
+  Equivalent of `check_array` in scikit-learn.
+
+  Parameters
+  ----------
+  tuples : object
+    The tuples to check.
+
+  Returns
+  -------
+  tuples_valid : object
+    The validated input.
+  """
+  # If input is scalar raise error
+  if len(tuples.shape) == 0:
+    raise ValueError(
+      "Expected 3D array, got scalar instead. Cannot apply this function on "
+      "scalars.")
+  # If input is 1D raise error
+  if len(tuples.shape) == 1:
+    raise ValueError(
+      "Expected 3D array, got 1D array instead:\ntuples={}.\n"
+      "Reshape your data using tuples.reshape(1, -1, 1) if it contains a "
+      "single tuple and the points in the tuple have a single "
+      "feature.").format(tuples)
+  # If input is 2D raise error
+  if len(tuples.shape) == 2:
+    raise ValueError(
+      "Expected 3D array, got 2D array instead:\ntuples={}.\n"
+      "Reshape your data either using tuples.reshape(-1, {}, 1) if "
+      "your data has a single feature or tuples.reshape(1, {}, -1) "
+      "if it contains a single tuple.".format(tuples, tuples.shape[1],
+                                              tuples.shape[0]))
+  return tuples
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -9,8 +9,6 @@ from ._util import check_tuples
 
 
 class BaseMetricLearner(six.with_metaclass(ABCMeta, BaseEstimator)):
-  def __init__(self):
-    raise NotImplementedError('BaseMetricLearner should not be instantiated')
 
   @abstractmethod
   def score_pairs(self, pairs):
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -50,19 +50,31 @@ class MahalanobisMixin(six.with_metaclass(ABCMeta)):
   """Mahalanobis metric learning algorithms.
 
   Algorithm that learns a Mahalanobis (pseudo) distance :math:`d_M(x, x')`,
-  defined between two column vectors :math:`x` and :math:`x'` by:
-  :math:`d_M(x, x') = \sqrt{(x-x')^T M (x-x')}`, where :math:`M` is the
-  learned square matrix.
+  defined between two column vectors :math:`x` and :math:`x'` by: :math:`d_M(x,
+  x') = \sqrt{(x-x')^T M (x-x')}`, where :math:`M` is a learned symmetric
+  positive semi-definite (PSD) matrix. The metric between points can then be
+  expressed as the euclidean distance between points embedded in a new space
+  through a linear transformation. Indeed, the above matrix can be decomposed
+  into the product of two transpose matrices (through SVD or Cholesky
+  decomposition): :math:`d_M(x, x')^2 = (x-x')^T M (x-x') = (x-x')^T L^T L
+  (x-x') = (L x - L x')^T (L x- L x')`
 
   Attributes
   ----------
-  metric_: `np.ndarray`, shape=(n_features, n_features)
-      The learned Mahalanobis matrix.
+  metric_: `np.ndarray`, shape=(n_features_out, n_features)
+    The learned metric ``M``.
+  transformer_: `np.ndarray`, shape=(n_features_out, n_features)
+    The learned linear transformation ``L``.
   """
 
   @property
   @abstractmethod
   def metric_(self):
+    pass
+
+  @property
+  @abstractmethod
+  def transformer_(self):
     pass
 
 
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -9,18 +9,6 @@ class BaseMetricLearner(BaseEstimator, TransformerMixin):
   def __init__(self):
     raise NotImplementedError('BaseMetricLearner should not be instantiated')
 
-  def metric(self):
-    """Computes the Mahalanobis matrix from the transformation matrix.
-
-    .. math:: M = L^{\\top} L
-
-    Returns
-    -------
-    M : (d x d) matrix
-    """
-    L = self.transformer()
-    return L.T.dot(L)
-
   def transformer(self):
     """Computes the transformation matrix from the Mahalanobis matrix.
 
@@ -30,7 +18,7 @@ class BaseMetricLearner(BaseEstimator, TransformerMixin):
     -------
     L : upper triangular (d x d) matrix
     """
-    return cholesky(self.metric()).T
+    return cholesky(self.metric_).T
 
   def transform(self, X=None):
     """Applies the metric transformation.
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -29,7 +29,7 @@ class BaseMetricLearner(BaseEstimator):
     """
 
 
-class MetricTransformer():
+class MetricTransformer(object):
 
   @abstractmethod
   def transform(self, X):
@@ -63,8 +63,8 @@ class MahalanobisMixin(six.with_metaclass(ABCMeta, BaseMetricLearner,
 
   Attributes
   ----------
-  transformer_ : `np.ndarray`, shape=(num_dims, n_features)
-    The learned linear transformation ``L``.
+  transformer_ : `numpy.ndarray`, shape=(num_dims, n_features)
+      The learned linear transformation ``L``.
   """
 
   def score_pairs(self, pairs):
@@ -117,7 +117,7 @@ class MahalanobisMixin(six.with_metaclass(ABCMeta, BaseMetricLearner,
   def metric(self):
     return self.transformer_.T.dot(self.transformer_)
 
-  def _transformer_from_metric(self, metric):
+  def transformer_from_metric(self, metric):
     """Computes the transformation matrix from the Mahalanobis matrix.
 
     Since by definition the metric `M` is positive semi-definite (PSD), it
@@ -257,4 +257,4 @@ class _QuadrupletsClassifierMixin(BaseMetricLearner):
       The quadruplets score.
     """
     quadruplets = check_tuples(quadruplets)
-    return - np.mean(self.predict(quadruplets))
+    return -np.mean(self.predict(quadruplets))
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -11,6 +11,22 @@ class BaseMetricLearner(BaseEstimator):
   def __init__(self):
     raise NotImplementedError('BaseMetricLearner should not be instantiated')
 
+  @abstractmethod
+  def score_pairs(self, pairs):
+    """Returns the score between pairs
+    (can be a similarity, or a distance/metric depending on the algorithm)
+
+    Parameters
+    ----------
+    pairs : `numpy.ndarray`, shape=(n_samples, [2,] n_features)
+      3D array of pairs, or 2D array of one pair.
+
+    Returns
+    -------
+    scores: `numpy.ndarray` of shape=(n_pairs,) or scalar
+      The score of every pair.
+    """
+
 
 class MetricTransformer(TransformerMixin):
 
@@ -35,7 +51,7 @@ class MetricTransformer(TransformerMixin):
     return X.dot(L.T)
 
 
-class MahalanobisMixin(six.with_metaclass(ABCMeta)):
+class MahalanobisMixin(six.with_metaclass(ABCMeta, BaseMetricLearner)):
   """Mahalanobis metric learning algorithms.
 
   Algorithm that learns a Mahalanobis (pseudo) distance :math:`d_M(x, x')`,
@@ -50,9 +66,34 @@ class MahalanobisMixin(six.with_metaclass(ABCMeta)):
 
   Attributes
   ----------
-  transformer_: `np.ndarray`, shape=(n_features_out, n_features)
+  transformer_ : `np.ndarray`, shape=(n_features_out, n_features)
     The learned linear transformation ``L``.
   """
+
+  def score_pairs(self, pairs):
+    """Returns the learned Mahalanobis distance between pairs.
+
+    This distance is defined as: :math:`d_M(x, x') = \sqrt{(x-x')^T M (x-x')}`
+    where ``M`` is the learned Mahalanobis matrix, for every pair of points
+    ``x`` and ``x'``. This corresponds to the euclidean distance between
+    embeddings of the points in a new space, obtained through a linear
+    transformation. Indeed, we have also: :math:`d_M(x, x') = \sqrt{(x_e -
+    x_e')^T (x_e- x_e')}`, with :math:`x_e = L x` (See
+    :class:`MahalanobisMixin`).
+
+    Parameters
+    ----------
+    pairs : `numpy.ndarray`, shape=(n_samples, [2,] n_features)
+      3D array of pairs, or 2D array of one pair.
+
+    Returns
+    -------
+    scores: `numpy.ndarray` of shape=(n_pairs,) or scalar
+      The learned Mahalanobis distance for every pair.
+    """
+    pairwise_diffs = pairs[..., 1, :] - pairs[..., 0, :]
+    return np.sqrt(np.sum(pairwise_diffs.dot(self.metric()) * pairwise_diffs,
+                          axis=-1))
 
   def metric(self):
     return self.transformer_.T.dot(self.transformer_)
@@ -69,7 +110,7 @@ class MahalanobisMixin(six.with_metaclass(ABCMeta)):
     return cholesky(metric).T
 
 
-class _PairsClassifierMixin:
+class _PairsClassifierMixin(BaseMetricLearner):
 
   def predict(self, pairs):
     """Predicts the learned metric between input pairs.
@@ -87,9 +128,7 @@ class _PairsClassifierMixin:
     y_predicted : `numpy.ndarray` of floats, shape=(n_constraints,)
       The predicted learned metric value between samples in every pair.
     """
-    pairwise_diffs = pairs[:, 0, :] - pairs[:, 1, :]
-    return np.sqrt(np.sum(pairwise_diffs.dot(self.metric()) * pairwise_diffs,
-                          axis=1))
+    return self.score_pairs(pairs)
 
   def decision_function(self, pairs):
     return self.predict(pairs)
@@ -120,7 +159,7 @@ class _PairsClassifierMixin:
     return roc_auc_score(y, self.decision_function(pairs))
 
 
-class _QuadrupletsClassifierMixin:
+class _QuadrupletsClassifierMixin(BaseMetricLearner):
 
   def predict(self, quadruplets):
     """Predicts differences between sample distances in input quadruplets.
@@ -138,12 +177,8 @@ class _QuadrupletsClassifierMixin:
     prediction : `numpy.ndarray` of floats, shape=(n_constraints,)
       Metric differences.
     """
-    similar_diffs = quadruplets[:, 0, :] - quadruplets[:, 1, :]
-    dissimilar_diffs = quadruplets[:, 2, :] - quadruplets[:, 3, :]
-    return (np.sqrt(np.sum(similar_diffs.dot(self.metric()) *
-                           similar_diffs, axis=1)) -
-            np.sqrt(np.sum(dissimilar_diffs.dot(self.metric()) *
-                           dissimilar_diffs, axis=1)))
+    return (self.score_pairs(quadruplets[:, 0:1, :]) -
+            self.score_pairs(quadruplets[:, 2:3, :]))
 
   def decision_function(self, quadruplets):
     return self.predict(quadruplets)
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -197,8 +197,8 @@ class _QuadrupletsClassifierMixin(BaseMetricLearner):
     prediction : `numpy.ndarray` of floats, shape=(n_constraints,)
       Metric differences.
     """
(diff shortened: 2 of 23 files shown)


metric_learn/base_metric.py
```diff
@@ -28,7 +18,7 @@ def transformer(self):
     -------
     L : upper triangular (d x d) matrix
     """
-    return cholesky(self.metric()).T
+    return cholesky(self.metric_).T
```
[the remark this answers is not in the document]
wdevazelhes: You're right, it should not be there. I was waiting to implement ``ExplicitMixin`` so that I could move them there, and I left it here to pass tests but I should rather have copied it in every Metric Learner class to be coherent with inheritance. Anyway now that we've decided to implement everything in ``MahalanobisMixin`` for now (I updated the PR message),  I'll just need to move ``transformer`` as well as ``transform`` in ``MahalanobisMixin``.

metric_learn/base_metric.py
```diff
@@ -49,3 +39,23 @@ def transform(self, X=None):
...
+  :math:`d_M(x, x') = \sqrt{(x-x')^T M (x-x')}`, where :math:`M` is the
+  learned square matrix.
+
+  Attributes
+  ----------
+  metric_: `np.ndarray`, shape=(n_features, n_features)
+      The learned Mahalanobis matrix.
+  """
+
+  @property
```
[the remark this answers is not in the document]
wdevazelhes: I guess I saw the ``@property`` decorator as a way to enforce children to have a ``metric_`` attribute but I forgot that it would make ``metric_`` read-only. Indeed, I do not see any reason why it would be better to have it read-only, so maybe we could do it another way ? Like set a ``metric_`` argument to None in the ``__init__`` ? Or maybe after all there is no need to enforce it: we could just document it, for instance putting ``metric_`` in ``MahalanobisMixin``'s docstring under ``Attributes``, and doing it also for children Metric Learners, like for instance in scikit-learn ``BaseEnsemble`` documents an attribute ``estimators_`` that is also documented in ``RandomForestClassifier``(see  
https://github.com/scikit-learn/scikit-learn/blob/7a23b2493fbe094db2528114d75edc2e9739da17/sklearn/ensemble/base.py#L60-L85  and   https://github.com/scikit-learn/scikit-learn/blob/4d9a12d175a38f2bcb720389ad2213f71a3d7697/sklearn/ensemble/forest.py#L738-L862 )

wdevazelhes:
>It seems a lot of these implementations could be simplified by deferring to transformer() and having that method check if things have been fitted.

Yes I agree.
I was thinking on how to do it for ``metric()`` too (if we revert to a function the rather than a read-only attribute). And the thing is that in the case where we want to return initialisations (like ``A0``) it's hard to test if the initialisation is a transformer initialisation or a mahalanobis initialisation, except if we had a consistent naming for initialisation arguments (like ``A0`` for a Mahalanobis metric initialisation or ``L0`` for a Transformer initialisation for instance).

We would then have something like this (looking more like what is currently in master):

```python
class MahalanobisMixin:

  def metric(self):
    if hasattr(self, 'metric_'):
      return self.metric_  # in this case the estimator is fitted
    elif hasattr(self, 'A0') and _is_arraylike(self.A0):
      return check_array(self.A0)
    elif hasattr(self, 'L0') or hasattr(self, 'transformer_'):
      return self.metric_from_transformer(self.transformer())
    else:  # extracted from scikit-learn's check_is_fitted function
      msg = ("This %(name)s instance is not fitted yet, and is neither "
             "initialized with an explicit matrix. Call 'fit' with appropriate"
             " arguments before using this method, or initialize the metric_ "
             "with ``prior`` equals a matrix, not None.")
      raise NotFittedError(msg % {'name': type(self).__name__})

  def transformer(self):
    if hasattr(self, 'transformer_'):
      return self.transformer_  # in this case the estimator is fitted
    elif hasattr(self, 'L0') and _is_arraylike(self.L0):
      return check_array(self.L0)
    elif hasattr(self, 'A0') or hasattr(self, 'metric_'):
      return self.transformer_from_metric(self.metric()) 
    else:  # extracted from scikit-learn's check_is_fitted function
      msg = ("This %(name)s instance is not fitted yet, and is neither "
             "initialized with an explicit matrix. Call 'fit' with appropriate"
             " arguments before using this method, or initialize the metric_ "
             "with ``prior`` equals a matrix, not None.")
      raise NotFittedError(msg % {'name': type(self).__name__})

 def metric_from_transformer(self, transformer):
   return L.T.dot(L)

     
 def transformer_from_metric(self, metric):
   return cholesky(metric).T
```
	
This way children classes only need to have a ``metric_`` or ``transformer_`` attribute depending on what they learn, and can reimplement the ``transformer_from_metric`` or ``metric_from_transformer`` method if needed (for instance ``MMC`` does a diagonal decomposition instead of the default cholesky decomposition)

What do you think ?

perimosocordiae:
I like that idea. I think it covers a lot of common cases in a consistent way, with room to specialize where needed.

Instead of using an `A0`/`L0` convention, what do you think about having the priors/initializations stored as `metric_`/`transformer_` and keeping a boolean flag `self._is_fitted` to track the distinction? In my opinion, the less `hasattr` we need to use, the better.

wdevazelhes:
Yes, and it would simplify the above code by having to implement only the ``metric()`` method (or property) in the base class (this one never changes, it is always L.T.dot(L)). Then we could still provide to children a default ``transformer_from_metric`` method  (that could be overwritten for instance by ``MMC``) that would be called just at the end of ``fit`` for instance, to have the ``transformer_`` attribute for algorithms that learn the metric rather than the transformer.

bellet:
I don't think `transformer_from_metric` would need to be overwritten by any algorithm. I would recommend using eigenvalue decomposition in all cases rather than cholesky as this allows to check for negligible eigenvalues, and in this case make `L` rectangular so that it reduces the dimension (instead of having L always square)

wdevazelhes:
> Instead of using an A0/L0 convention, what do you think about having the priors/initializations stored as metric_/transformer_ 

The problem is that if we want to be scikit-learn compliant, (and therefore pass checks that are copied from scikit-learn ``check_estimator``, in particular ``check_no_attributes_set_in_init`` (called ``test_no_attributes_set_in_init`` in ``metric-learn``)), we cannot set any attribute at ``init`` with a name different than an argument name, so we cannot do:
```python
def __init__(self, L0):
  self.transformer_ = L0
```
Calling an init argument ``transformer_`` would be weird so I think that if we want to support returning ``L0`` when calling ``transformer_`` on an unfitted estimator, we must use properties, am I right ? Eventually we discussed that we also could use setters if we want ``transformer_`` to be writeable

wdevazelhes:
> I don't think transformer_from_metric would need to be overwritten by any algorithm. I would recommend using eigenvalue decomposition in all cases rather than cholesky as this allows to check for negligible eigenvalues, and in this case make L rectangular so that it reduces the dimension (instead of having L always square)

Yes, that's right
I guess the threshold/max_dimension would be set in the __init__ of the metric learner ? Maybe we could do it in a separate PR (and add it to the TODO of issue https://github.com/metric-learn/metric-learn/issues/91), since it would add also some API modifications ? 
In fact finally I would also put the feature to return the initialisation when calling ``metric_`` in another PR, since it is not crucial for the new API.

wdevazelhes:
I'm thinking maybe we could also allow the user to choose the decomposition type ? Because for large matrices (like 4096x4096 for instance, like the olivetti faces dataset), eigenvalues decomoposition is quite more expensive:
```python
In [32]: import numpy as np
    ...: M = np.random.randn(4096, 4096)
    ...: M = M.dot(M.T)
    ...: %timeit np.linalg.eigh(M)
    ...: %timeit np.linalg.cholesky(M)
    ...: 
26 s ± 7.13 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.05 s ± 31.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

bellet:
Good point. For now, maybe we should always use cholesky of `M` to get `L` (hence ignoring potential gains in dimensionality reduction)? And keep this issue as a future todo item as this is somewhat an auxiliary aspect

wdevazelhes:
Yes I agree, I just added it to the todo

wdevazelhes:
Hi, I just updated the PR, I think it should be ready to merge. 
These are some other things I added to the todo: 
- rename num_dims to n_components to be similar to scikit learn's naming: what do you think ? 
- check arrays/tuples before predicting/scoring/etc and check if fitted

metric_learn/base_metric.py
```diff
@@ -52,11 +47,90 @@ def transform(self, X=None):
...
+    -------
+    X_embedded : `numpy.ndarray`, shape=(n_samples, num_dims)
+      The embedded data points.
+    """
+    return X.dot(self.transformer_.T)
+
+  def metric(self):
+    return self.transformer_.T.dot(self.transformer_)
+
+  def transformer_from_metric(self, metric):
```
bellet: this should probably be a private method?

metric_learn/base_metric.py
```diff
@@ -107,7 +179,7 @@ def score(self, pairs, y):
     return roc_auc_score(y, self.decision_function(pairs))
 
 
-class _QuadrupletsClassifierMixin:
+class _QuadrupletsClassifierMixin(BaseMetricLearner):
 
   def predict(self, quadruplets):
```
bellet: it would be more logical if `predict` would compute accuracy on quadruplets (proportion of quadruplets correctly ordered) and `score` would compute the difference between distances

metric_learn/mmc.py
```diff
@@ -350,10 +355,7 @@ def _D_constraint(self, neg_pairs, w):
       sum_deri2 / sum_dist - np.outer(sum_deri1, sum_deri1) / (sum_dist * sum_dist)
     )
 
-  def metric(self):
-    return self.A_
-
-  def transformer(self):
+  def transformer_from_metric(self, metric):
```
bellet: I can't see why we need to override `transformer_from_metric` for MMC?

test/test_mahalanobis_mixin.py
```diff
@@ -0,0 +1,203 @@
...
+                  'nca',
+                  'rca',
+                  'sdml',
+                  ]
+
+
+@pytest.mark.parametrize('estimator, build_dataset', list_estimators,
+                         ids=ids_estimators)
+def test_score_pairs_pairwise(estimator, build_dataset):
+  # Computing pairwise scores should return an euclidean distance matrix.
```
bellet: an --> a

test/test_mahalanobis_mixin.py
```diff
@@ -0,0 +1,203 @@
...
+  X = X[:n_samples]
+  model = clone(estimator)
+  model.fit(inputs, labels)
+
+  pairwise = model.score_pairs(np.array(list(product(X, X))))\
+      .reshape(n_samples, n_samples)
+
+  check_is_distance_matrix(pairwise)
+
+  # a necessary condition for euclidean distances matrix: (see
```
bellet: euclidean distance matrices

test/test_mahalanobis_mixin.py
```diff
@@ -0,0 +1,203 @@
...
+  X, _ = load_iris(return_X_y=True)
+  n_samples = 20
+  X = X[:n_samples]
+  model = clone(estimator)
+  model.fit(inputs, labels)
+
+  pairwise = model.score_pairs(np.array(list(product(X, X))))\
+      .reshape(n_samples, n_samples)
+
+  check_is_distance_matrix(pairwise)
```
bellet: I think `score_pairs` returns squared distances, no? so this should be called on the element-wise square root of `pairwise`?

test/test_mahalanobis_mixin.py
```diff
@@ -0,0 +1,203 @@
...
+
+  # a necessary condition for euclidean distances matrix: (see
+  # https://en.wikipedia.org/wiki/Euclidean_distance_matrix)
+  assert np.linalg.matrix_rank(pairwise**2) <= min(X.shape) + 2
```
bellet: this should be square root instead of square?

test/test_mahalanobis_mixin.py
```diff
@@ -0,0 +1,203 @@
...
+
```
bellet: you could also text the following: get the transformer, transform data and then compute Euclidean distance with `scipy.spatial.distance.pdist`, then compare that it is the same as what you obtain

test/test_mahalanobis_mixin.py
```diff
@@ -0,0 +1,203 @@
...
+  assert model.score_pairs(tuples).shape == (tuples.shape[0],)
+  assert np.isscalar(model.score_pairs(tuples[1]))
+
+
+def check_is_distance_matrix(pairwise):
+  assert (pairwise >= 0).all()  # positivity
+  assert np.array_equal(pairwise, pairwise.T)  # symmetry
+  assert (pairwise.diagonal() == 0).all()  # identity
+  # triangular inequality
+  for i in range(pairwise.shape[1]):
```
bellet: not crucial but there must be a more efficient way to do this. at least, since you already checked that the matrix is symmetric you do not need to check all of them

bellet:
Also just an observation: Mahalanobis algorithms taking as input labeled data `X, y` (like NCA or LMNN) inherit both from `MahalanobisMixin` and `MetricTransformer`. Hence they have two methods that do the same thing (`embed` and `transform`). Not sure whether/how this redundancy can be removed (or whether it would be a good idea to remove it).

metric_learn/base_metric.py
wdevazelhes: Agreed

metric_learn/base_metric.py
wdevazelhes: I agree that having a `predict` and a `decision_function` doing the same thing is not ideal, and in the case of quadruplets this can be fixed easily since there is no threshold to fix like in pairs
But I did not really get what you mean ? 
Since `score` is the scikit-learn like function for scoring it should return one scalar, whereas `predict` should return a sample-wise output
I guess the most coherent with scikit-learn would be that `predict` would output a binary sign (or 0 or 1) depending on the ordering of pairs in the quadruplet, and `decision_function` would return the differences between distances, since it is a float-type score (like `predict_proba` but without being necessarily between 0 and 1)
`score` would return the accuracy of the predict function (proportion of quadruplets correctly ordered)

metric_learn/mmc.py
wdevazelhes: Agreed
I tried to remove it but `metric_learn_test.TestMMC` fails, throwing
```pytb
E       numpy.linalg.linalg.LinAlgError: Matrix is not positive definite
```
And indeed, looking into this test with the overwritten `_transformer_from_metric` with eigendecomposition, after doing the eigendecomposition, 3 out of 4 eigenvalues are very close to zero, so this would mean the metric is semi-definite (as stated in the code comment).
Cholesky decomposition is defined for semidefinite matrices, but looking it up it appears that the scipy's routine does not work in such a case (https://mail.scipy.org/pipermail/scipy-user/2011-July/029927.html). So I guess I could deal with semidefinite case inside the general `_transformer_from_metric` with maybe a little function to test if the matrix invertible with `numpy.linalg.det
` ? And in this case we would do the eigenvalues decomposition ?

test/test_mahalanobis_mixin.py
wdevazelhes: thanks, will do

test/test_mahalanobis_mixin.py
wdevazelhes: will do

test/test_mahalanobis_mixin.py
wdevazelhes: For now `score_pairs` returns not squared distances, as implemented in `MahalanobisMixin`
https://github.com/wdevazelhes/metric-learn/blob/eff278e007dded4a6f5ee2ead3ae92bb8c720eb6/metric_learn/base_metric.py#L97
Should it return squared distances though ? I will add the TODO to at least give the option because it would prevent some useless computations if the true distance is not needed

test/test_mahalanobis_mixin.py
wdevazelhes: same as above: `score_pairs` returns true euclidean distances, so I guess it is OK here too

test/test_mahalanobis_mixin.py
wdevazelhes: Agreed, will do

test/test_mahalanobis_mixin.py
wdevazelhes: Agreed, I'll make the `j` loop start from `i+1` instead of 0 (+1 because with positivity, identity and symmetry `d(x,x) <=d(x,y) + d(y,x)` is already tested), this should reduce the complexity from `n^3` to `n^3/2` I think. I could also prevent from testing if `k==i` or `k== j` since I already tested identity, but this would not be so significant, so maybe no need to include this `if` in the code
I don't know if it could be made even more efficient

metric_learn/base_metric.py
bellet: Yes, this is what I meant, sorry for the confusion

metric_learn/mmc.py
bellet: I think an easy solution is to apply Cholesky to `metric` plus a very small factor of the identity matrix

wdevazelhes:
I just updated the docstrings for the `transformer_` in each class, since Attributes are not inherited in sphinx (see issue https://github.com/sphinx-doc/sphinx/issues/741 ), and anyway I wanted to add some extra info for when `transformer_` is deduced from a learned metric with `transformer_from_metric` (I made the method public to show its docstring in the doc and be able to link to it). 

I think the PR is now ready to merge, which will allow to read more easily the diff  PR #117.

metric_learn/_util.py
```diff
@@ -9,4 +9,42 @@ def vector_norm(X):
...
+  tuples : object
+    The tuples to check.
+
+  Returns
+  -------
+  tuples_valid : object
+    The validated input.
+  """
+  # If input is scalar raise error
+  if len(tuples.shape) == 0:
```
perimosocordiae: [np.isscalar](https://docs.scipy.org/doc/numpy/reference/generated/numpy.isscalar.html)

metric_learn/base_metric.py
```diff
@@ -1,62 +1,149 @@
...
+      The score of every pair.
     """
-    L = self.transformer()
-    return L.T.dot(L)
 
-  def transformer(self):
-    """Computes the transformation matrix from the Mahalanobis matrix.
 
-    L = cholesky(M).T
+class MetricTransformer():
```
perimosocordiae: subclass from `object`?

metric_learn/base_metric.py
```diff
@@ -154,4 +255,5 @@ def score(self, quadruplets, y=None):
     score : float
       The quadruplets score.
     """
-    return - np.mean(np.sign(self.decision_function(quadruplets)))
+    quadruplets = check_tuples(quadruplets)
+    return - np.mean(self.predict(quadruplets))
```
perimosocordiae: The space between the `-` and `np.mean` is a little confusing.

test/test_mahalanobis_mixin.py
perimosocordiae: How about this:

```python
rows, cols = pairwise.shape
tol = 1e-3  # seems kinda large, but we can tune this
for i in range(rows):
  for j in range(i, cols):
    direct_dist = pairwise[i, j]
    triangle_dist = np.add.outer(pairwise[i], pairwise[:, j])
    assert (direct_dist <= triangle_dist + tol).all()

File: metric_learn/_util.py (hunk 12 above)
Comment (wdevazelhes) [reply metric_learn/_util.py]: That's right
Done

File: metric_learn/base_metric.py (hunk 13 above)
Comment (wdevazelhes) [reply metric_learn/base_metric.py]: Yes
Done

File: metric_learn/base_metric.py (hunk 14 above)
Comment (wdevazelhes) [reply metric_learn/base_metric.py]: Agreed
Done

**Comment by wdevazelhes** [comment]
>How about this:
rows, cols = pairwise.shape
tol = 1e-3  # seems kinda large, but we can tune this
for i in range(rows):
  for j in range(i, cols):
    direct_dist = pairwise[i, j]
    triangle_dist = np.add.outer(pairwise[i], pairwise[:, j])
    assert (direct_dist <= triangle_dist + tol).all()


That's better indeed

I didn't know about `np.add.outer`, would you say it is a better practice than broadcasting-like operations for instance (like `a + a[:, None]`)?

File: test/test_mahalanobis_mixin.py [hunk 15]
```diff
@@ -150,12 +150,13 @@ def check_is_distance_matrix(pairwise):
...
   assert (pairwise.diagonal() == 0).all()  # identity
   # triangular inequality
-  for i in range(pairwise.shape[1]):
-    # since we already checked symmetry we can start at i
-    for j in range(i, pairwise.shape[1]):
-      for k in range(pairwise.shape[1]):
-        assert (pairwise[i, j] - (pairwise[i, k] + pairwise[k, j]) <= 0 +
-                1e-3).all()
+  rows, cols = pairwise.shape
+  tol = np.finfo(float).eps
```
wdevazelhes: Maybe here we can start with the smaller value possible, and if some distance function fails to pass the test, we could investigate more and decide what is a reasonable value ?

wdevazelhes:
> How about this:
>     rows, cols = pairwise.shape
>     tol = 1e-3 # seems kinda large, but we can tune this
>     for i in range(rows):
>     for j in range(i, cols):
>     direct_dist = pairwise[i, j]
>     triangle_dist = np.add.outer(pairwise[i], pairwise[:, j])
>     assert (direct_dist <= triangle_dist + tol).all()
> 
> >That's better indeed
> I didn't know about np.add.outer, would you say it is a better practice than broadcasting-like operations for instance (like a + a[:, None])?

This change breaks the test, I think in fact it is not exactly what we want to test ? Because it will test the triangle inequality but also some inequalities like d(x1, x2) <= d(x1, x3) + d(x4, x2), which has no reason to be true isn't it ?

test/test_mahalanobis_mixin.py
```diff
@@ -150,13 +150,9 @@ def check_is_distance_matrix(pairwise):
...
-  rows, cols = pairwise.shape
-  tol = np.finfo(float).eps
-  for i in range(rows):
-    for j in range(i, cols):
-      direct_dist = pairwise[i, j]
-      triangle_dist = np.add.outer(pairwise[i], pairwise[:, j])
-      assert (direct_dist <= triangle_dist + tol).all()
+  tol = 1e-15
+  assert (pairwise <= pairwise[:, :, np.newaxis]
+          + pairwise[:, np.newaxis, :] + tol).all()
```
wdevazelhes: I think this would work, checking that each `pairwise[i, j] <= pairwise[i, :] + pairwise[:, j]` (the strict version of this would be with `pairwise.T[:, :, np.newaxis] + pairwise[:, np.newaxis, :]` but since pairwise is already checked to be symmetric we can avoid the transpose)
For the value `1e-15`, the test failed with 1e-16 for lsml, when sometimes running the tests, and sometimes not. I think 1e-15 is a reasonable value to set for now, but I will also make a commit to set all random_states (this made me realized they were not)

perimosocordiae:
Ah, you're right, we don't want the outer add there. I should have just suggested `pairwise[i,:] + pairwise[:,j]`.

(But in general, I do like using explicit "outer" rather than indexing, as it can sometimes be confusing what the intent is, especially for readers without a lot of experience with numpy.)

The full vectorized version is fine, as long as the total size of the matrix is relatively small.

perimosocordiae:
As this PR isn't going to affect the master branch, I think it's ready to merge now. Any objections?

bellet:
I can take a last look early next week if you want, I think I missed the last few commits

perimosocordiae:
Sounds good. I just didn't want this PR to stall if it's ready to go.

bellet:
Since for now we decided to use transform even for PairsClassifierMixin and such, shouldn't MahalanobisMixin be a TransformerMixin? This would avoid having to write explicitly that supervised algorithms inherit from both.
Otherwise LGTM

wdevazelhes:
> Since for now we decided to use transform even for PairsClassifierMixin and such, shouldn't MahalanobisMixin be a TransformerMixin? This would avoid having to write explicitly that supervised algorithms inherit from both.
> Otherwise LGTM

The problem is that `TransformerMixin` adds the `fit_transform` method (see https://github.com/scikit-learn/scikit-learn/blob/f0ab589f1541b1ca4570177d93fd7979613497e3/sklearn/base.py#L490-L520), 
and for now this function is undefined for weakly supervised learners, that's why only supervised ones should  inherit from `TransformerMixin`

`MetricTransformer` on the other hand, just enforces that children classes have a `transform` method (which `TransformerMixin` does not by the way, so maybe this class (`MetricTransformer`) is not necessary and just adds complexity, in addition to the fact that now we only have `MahalanobisMixin` that inherits from it ?)

wdevazelhes:
(sorry for the last commits https://github.com/metric-learn/metric-learn/pull/96/commits/779a93a2f236925044576f3905300b3f8f4ced2d and https://github.com/metric-learn/metric-learn/pull/96/commits/75d4ad231520796c658617fc8d81ca7945a3c397, I did some changes on a local repo that was not up to date, so I had to do a merge afterwards) the commit is basically just making the `transform` method of `MetricTransformer` abstract, and removes the part of the doc that was saying you could use X=None (the real changes are in https://github.com/metric-learn/metric-learn/pull/96/commits/779a93a2f236925044576f3905300b3f8f4ced2d)

bellet:
OK then I think this is ready to merge. Whether or not to keep `MetricTransformer` does not seem like a very important issue at this point

PR #99, opened 2018-06-29
[MRG] Add memory efficient implementation of NCA

### LINKED ISSUE #45: NCA - performance issues
The current NCA implementation creates a numpy matrix (tmp) of shape (n, n, d, d). It is problem even with medium size datasets(memory requirement n².d²). On my computer I have 64Go of memory, and I cannot run NCA on a dataset of 1021 instances in dimension 122.

A simple solution to deal with this memory problem is to compute tmp_i at each iteration:
`tmp_i = np.einsum('...i,...j->...ij', dX[i], dX[i])  # shape (n, d, d)`
This way, the memory requirement in only (n . d²), and it works on the dataset of dimension (1021, 122). However, it requires more than 12 hours to compute !!

This solution increases the execution time as tmp_i is recomputed at each iteration :

Shape of the input data     |    Previous execution time     |     New execution time
(44, 122)                            |    45s                                     |     52s
(198, 122)                          |    20min                                 |     25min
(1021, 122)                        |    Memory Error                     |     >12h

Even for small datasets the execution time is very long. Do you have any idea on how improve the performance of NCA ?
 - Can we use an optimization library such as scipy.optimize (used for the implementation of MLKR for instance) ? Do you think it would significantly reduce the execution time ?
 - Another solution which seems more complex is to re-implementing the optimization in cython.

wdevazelhes:
Fixes #45 
This PR adds a more memory efficient implementation of the gradient of NCA, avoiding creation of a (n, n, d, d) matrix. This is taken from a PR I opened in scikit-learn: https://github.com/scikit-learn/scikit-learn/pull/10058
Eventually, some others improvements of this PR (like verbose, checkers, choice of initializations) would be integrated in metric-learn, but I think for now just changing the gradient computation is fine to solve #45.

Here is a little snippet to show the performance gain: 

```python
from time import time
from sklearn.datasets import load_iris, load_digits
from metric_learn import NCA
for dataset in [load_iris(), load_digits()]:
	X, y = dataset.data, dataset.target
	nca = NCA()
	start = time()
	nca.fit(X, y)
	print(time() - start)
```
Prints for this PR: 
0.1164708137512207  (for iris)
21.722617626190186 (for digits)

Prints for master:
0.9377930164337158  (for iris)
Memory Error (for digits)

Note: with an identity initialization, it prints: 
0.10440230369567871  (for iris)
8.215924263000488 (for digits)
Therefore in this case the init is quite important to help the algorithm converge (identity seems a good init in this case)

### DIFF
--- a/metric_learn/nca.py
+++ b/metric_learn/nca.py
@@ -5,14 +5,11 @@ Ported to Python from https://github.com/vomjom/nca
 
 from __future__ import absolute_import
 
-import sys
-import time
 import warnings
 import numpy as np
 from scipy.optimize import minimize
-from sklearn.decomposition import PCA
 from sklearn.metrics import pairwise_distances
-from sklearn.utils.validation import check_X_y, check_random_state
+from sklearn.utils.validation import check_X_y
 
 try:  # scipy.misc.logsumexp is deprecated in scipy 1.0.0
     from scipy.special import logsumexp
@@ -26,11 +23,10 @@ EPS = np.finfo(float).eps
 
 class NCA(BaseMetricLearner):
   def __init__(self, num_dims=None, max_iter=100, learning_rate='deprecated',
-               random_state=0, tol=None):
+               tol=None):
     self.num_dims = num_dims
     self.max_iter = max_iter
     self.learning_rate = learning_rate  # TODO: remove in v.0.5.0
-    self.random_state = random_state
     self.tol = tol
 
   def transformer(self):
@@ -45,9 +41,6 @@ class NCA(BaseMetricLearner):
       warnings.warn('"learning_rate" parameter is not used.'
                     ' It has been deprecated in version 0.4 and will be'
                     'removed in 0.5', DeprecationWarning)
-
-    # Initialize the random generator
-    self.random_state_ = check_random_state(self.random_state)
 
     X, labels = check_X_y(X, y)
     n, d = X.shape
--- a/metric_learn/nca.py
+++ b/metric_learn/nca.py
@@ -61,7 +61,7 @@ class NCA(BaseMetricLearner):
     np.fill_diagonal(A, 1./(np.maximum(X.max(axis=0)-X.min(axis=0), EPS)))
 
     # Run NCA
-    mask = y[:, np.newaxis] == y[np.newaxis, :]
+    mask = labels[:, np.newaxis] == labels[np.newaxis, :]
     optimizer_params = {'method': 'L-BFGS-B',
                         'fun': self._loss_grad_lbfgs,
                         'args': (X, mask, -1.0),
--- a/metric_learn/nca.py
+++ b/metric_learn/nca.py
@@ -24,6 +24,27 @@ EPS = np.finfo(float).eps
 class NCA(BaseMetricLearner):
   def __init__(self, num_dims=None, max_iter=100, learning_rate='deprecated',
                tol=None):
+    """Neighborhood Components Analysis
+
+    Parameters
+    ----------
+    num_dims : int, optional (default=None)
+      Embedding dimensionality. If None, will be set to ``n_features``
+      (``d``) at fit time.
+
+    max_iter : int, optional (default=100)
+      Maximum number of iterations done by the optimization algorithm.
+
+    learning_rate : Not used
+
+      .. deprecated:: 0.4.0
+        `learning_rate` was deprecated in version 0.4.0 and will
+        be removed in 0.5.0. The current optimization algorithm does not need
+        to fix a learning rate.
+
+    tol : float, optional (default=None)
+        Convergence tolerance for the optimization.
+    """
     self.num_dims = num_dims
     self.max_iter = max_iter
     self.learning_rate = learning_rate  # TODO: remove in v.0.5.0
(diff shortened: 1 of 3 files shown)


metric_learn/nca.py
```diff
@@ -4,20 +4,35 @@
...
 
 from .base_metric import BaseMetricLearner
 
 EPS = np.finfo(float).eps
 
 
 class NCA(BaseMetricLearner):
-  def __init__(self, num_dims=None, max_iter=100, learning_rate=0.01):
+  def __init__(self, num_dims=None, max_iter=100, learning_rate='deprecated',
+               init='auto', random_state=0, tol=None):
```
bellet: remove init param

metric_learn/nca.py
```diff
@@ -4,20 +4,35 @@
...
     self.num_dims = num_dims
     self.max_iter = max_iter
-    self.learning_rate = learning_rate
+    self.learning_rate = learning_rate  # TODO: remove in v.0.5.0
+    self.init = init
```
bellet: again

wdevazelhes:
> LGTM as a quick fix. You should remove the init parameter which is not used, we can add more init schemes (factorized across several algorithms) in a later PR.

Thanks, I added it and removed it since it s a quick fix indeed, but forgot to remove these lines...

> Is there any documentation to update?

There was no documentation before, so I did not add one, but now that you say it, I think `tol`and `random_state` are less clear than `learning_rate` (and I should maybe warn again the fact that `learning_rate` is deprecated), so I think I will add some docstring to `__init__`

wdevazelhes:
I will also remove ``random_state`` since I had only put it for PCA initialization which is not done so it is not used

test/metric_learn_test.py
```diff
@@ -88,22 +89,65 @@ def test_iris(self):
...
     # Without dimension reduction
-    nca = NCA(max_iter=(100000//n), learning_rate=0.01)
+    nca = NCA(max_iter=(100000//n))
     nca.fit(self.iris_points, self.iris_labels)
-    # Result copied from Iris example at
-    # https://github.com/vomjom/nca/blob/master/README.mkd
-    expected = [[-0.09935, -0.2215,  0.3383,  0.443],
-                [+0.2532,   0.5835, -0.8461, -0.8915],
-                [-0.729,   -0.6386,  1.767,   1.832],
-                [-0.9405,  -0.8461,  2.281,   2.794]]
```
perimosocordiae: Have we verified that the new approach produces a correct result? Checking class separation is a very coarse approximation of correctness.

metric_learn/nca.py
```diff
@@ -27,33 +58,57 @@ def fit(self, X, y):
...
+    # Compute softmax distances
+    p_ij = pairwise_distances(X_embedded, squared=True)
+    np.fill_diagonal(p_ij, np.inf)
+    p_ij = np.exp(-p_ij - logsumexp(-p_ij, axis=1)[:, np.newaxis])
+    # (n_samples, n_samples)
+
+    # Compute loss
+    masked_p_ij = p_ij * mask
+    p = np.sum(masked_p_ij, axis=1, keepdims=True)  # (n_samples, 1)
+    loss = np.sum(p)
```
perimosocordiae: Stylistically, I prefer the method call `p.sum()` over the numpy function. But it's not a big deal if you like the other way better.

test/metric_learn_test.py
bellet: This is more a general comment but IMHO this sort of test is not very reliable: (i) the "expected" output comes either from running the code itself or running some other implementation which may or may not be reliable (in this case, the source does not seem especially reliable), and (ii) NCA being a nonconvex objective, depending on the initialization but also on the chosen optimization algorithm, one might converge to a different point, which does not imply that the algorithm is incorrect.

bellet: @wdevazelhes maybe some of the tests you designed for the sklearn version can be used instead?
https://github.com/wdevazelhes/scikit-learn/blob/cc072617e4670cafdec22df71b15021295e95b96/sklearn/neighbors/tests/test_nca.py

wdevazelhes: Also, the code in master uses stochastic gradient descent (updating L after a pass on each sample instead of the whole dataset). Therefore we will not be able to test this particular result (even if we tested without scipy's optimizer) since in this PR we compute the true full gradient at each iteration. 

> @wdevazelhes maybe some of the tests you designed for the sklearn version can be used instead?

Yes, I already added ``test_finite_differences`` (tests the gradient formula) and  ``test_simple_example`` (toy example, that checks on 4 points that 2 same labeled points become closer after fit_transform) . But now that you say it I will also add ``test_singleton_class`` and ``test_one_class``, which are also useful (they test edge cases on the dataset, and in some of these tests we know some analytical properties of the transformation: if only one class, or only singleton classes, then gradient is 0). However I don't think the other tests need to be included since they mostly test input formats, verbose, and other stuff that are necessary for inclusion in scikit-learn, but which I guess could be factored out for every algorithm in metric-learn in a later stage of development (and which were not enforced/tested for NCA before either so if we don't put them in this PR we don't regress).

metric_learn/nca.py
wdevazelhes: It's true that ``p.sum`` is more concise, I guess I like it more too
I'll make the change

GeorgePearse:
Hi @wdevazelhes, looks like you've done some great work here. What sort of relationship between dataset size and memory consumption does this method lead to?

PR #101, opened 2018-07-02
[MRG] FIX Fix LMNN rollback

### LINKED ISSUE #88: BUG. LMNN loops when a bad update is calculated.
There seems to be a relatively large bug in the code for LMNN. In lines 157-164 we have:
```
# update step size
  if delta_obj > 0:
    # we're getting worse... roll back!
    learn_rate /= 2.0
    df = df_old
    a1 = a1_old
    a2 = a2_old
    objective = objective_old
```
But in an iteration the objective is calculated as (line 149-150):
```
objective = total_active * (1-reg)
      objective += G.flatten().dot(L.T.dot(L).flatten())
```
Where none of total_active, reg, G, or L depend on the learn_rate. Since that is the only thing changed when the above error occurs, the result of the iteration following the roll back will be the same as teh one that caused the roll back. Hence as soon as lines 157 to 164 are executed, the algorithm will just keep halving the learning rate and re calculating the same values until the max iteration is hit (as can be seen below):
image
The solution most likely is to either make the calculation of the new objective dependent on the learning rate, or revert L to its previous value also (as in the value it had when the previous iteration that did not cause a roll back started).

wdevazelhes:
Fixes #88
 Stores ``L`` and ``G``  in addition to what was already stored (``df``, ``a1`` and ``a2``), at the last "good" point (meaning the last time a point had a better objective than all the previous points), and gets back to this point if the update worsens the objective.
Here is a little code to show the difference between this PR and master:

```python
from metric_learn import LMNN
lmnn = LMNN(verbose=True)
X, y = make_classification(random_state=0)
lmnn.fit(X, y)
```

This PR:
<details>

```
1 3719.8858456299436 -inf 419 1e-07
2 3719.3690067170437 -0.516838912899857 419 1.0099999999999999e-07
3 3718.8470839936936 -0.5219227233501442 419 1.0201e-07
4 3718.32002830601 -0.5270556876835144 419 1.030301e-07
5 3717.787790041948 -0.5322382640620162 419 1.0406040099999999e-07
/home/will/Code/metric-learn/metric_learn/lmnn.py:62: UserWarning: use_pca does nothing for the python_LMNN implementation
  warnings.warn('use_pca does nothing for the python_LMNN implementation')
6 3717.2503191273736 -0.5374709145744418 419 1.0510100500999999e-07
7 3716.707565022117 -0.5427541052567904 419 1.0615201506009999e-07
8 3716.1594767159886 -0.5480883061281929 419 1.0721353521070098e-07
9 3715.607125166057 -0.5523515499317 418 1.08285670562808e-07
10 3715.0504268415325 -0.5566983245244046 418 1.0936852726843608e-07
[...]
802 295.8600904596078 -0.006394422217169904 587 3.6168492926956634e-05
803 295.8625006630451 0.0024102034373072456 592 3.65301778562262e-05
804 295.8537018319783 -0.006388627629519306 592 1.844773981739423e-05
805 295.85064356510844 -0.003058266869857107 590 1.863221721556817e-05
806 295.84687892031536 -0.0037646447930796967 587 1.8818539387723852e-05
807 295.84384134091164 -0.0030375794037240667 595 1.900672478160109e-05
808 295.84056710633945 -0.0032742345721885613 584 1.9196792029417102e-05
809 295.83776376246647 -0.0028033438729835325 597 1.9388759949711275e-05
810 295.8366193762049 -0.0011443862615578837 582 1.9582647549208387e-05
811 295.8319662001932 -0.004653176011686355 595 1.977847402470047e-05
812 295.82935080014795 -0.002615400045272054 585 1.9976258764947475e-05
813 295.8262825137674 -0.0030682863805395755 596 2.0176021352596948e-05
814 295.8269467992082 0.0006642854407914456 589 2.0377781566122918e-05
LMNN converged with objective 295.8262825137674
```

</details>
<p></p>
Master:
<details>

```
1 3719.8858456299436 -inf 419 1e-07
2 3719.3690067170437 -0.516838912899857 419 1.0099999999999999e-07
3 3718.8470839936936 -0.5219227233501442 419 1.0201e-07
4 3718.32002830601 -0.5270556876835144 419 1.030301e-07
/home/will/Code/metric-learn/metric_learn/lmnn.py:62: UserWarning: use_pca does nothing for the python_LMNN implementation
  warnings.warn('use_pca does nothing for the python_LMNN implementation')
5 3717.787790041948 -0.5322382640620162 419 1.0406040099999999e-07
6 3717.2503191273736 -0.5374709145744418 419 1.0510100500999999e-07
7 3716.707565022117 -0.5427541052567904 419 1.0615201506009999e-07
8 3716.1594767159886 -0.5480883061281929 419 1.0721353521070098e-07
[...]
986 296.9272749579874 0.021928604062395607 593 8.894939764557867e-84
987 296.9272749579874 0.021928604062395607 593 4.4474698822789335e-84
988 296.9272749579874 0.021928604062395607 593 2.2237349411394667e-84
989 296.9272749579874 0.021928604062395607 593 1.1118674705697334e-84
990 296.9272749579874 0.021928604062395607 593 5.559337352848667e-85
991 296.9272749579874 0.021928604062395607 593 2.7796686764243334e-85
992 296.9272749579874 0.021928604062395607 593 1.3898343382121667e-85
993 296.9272749579874 0.021928604062395607 593 6.949171691060834e-86
994 296.9272749579874 0.021928604062395607 593 3.474585845530417e-86
995 296.9272749579874 0.021928604062395607 593 1.7372929227652084e-86
996 296.9272749579874 0.021928604062395607 593 8.686464613826042e-87
997 296.9272749579874 0.021928604062395607 593 4.343232306913021e-87
998 296.9272749579874 0.021928604062395607 593 2.1716161534565105e-87
999 296.9272749579874 0.021928604062395607 593 1.0858080767282552e-87
LMNN didn't converge in 1000 steps.
```

</details>


wdevazelhes:
I'll also add some non regression tests that fail in master and pass in this PR

test/metric_learn_test.py
```diff
@@ -70,6 +73,40 @@ def test_iris(self):
       csep = class_separation(lmnn.transform(), self.iris_labels)
       self.assertLess(csep, 0.25)
 
+  def test_convergence_simple_example(self):
+    # LMNN should converge on this simple example, which it did not with
+    # this issue: https://github.com/metric-learn/metric-learn/issues/88
+    X, y = make_classification(random_state=0)
+    old_stdout = sys.stdout
+    sys.stdout = StringIO()
+    lmnn = LMNN(verbose=True)
```
perimosocordiae: We should use `python_LMNN` here and the other test, to not fail when the shogun version is available.

test/metric_learn_test.py
```diff
@@ -70,6 +73,40 @@ def test_iris(self):
...
+    try:
+        lmnn.fit(X, y)
+    finally:
+        out = sys.stdout.getvalue()
+        sys.stdout.close()
+        sys.stdout = old_stdout
+    assert ("LMNN converged with objective" in out)
```
perimosocordiae: Style nit: the parens here aren't needed.

test/metric_learn_test.py
```diff
@@ -70,6 +73,40 @@ def test_iris(self):
...
+    X, y = make_classification(random_state=0)
+    old_stdout = sys.stdout
+    sys.stdout = StringIO()
+    lmnn = LMNN(verbose=True)
+    try:
+      lmnn.fit(X, y)
+    finally:
+      out = sys.stdout.getvalue()
+      sys.stdout.close()
+      sys.stdout = old_stdout
```
perimosocordiae: Might be nice to have this logic in a context manager. See https://eli.thegreenplace.net/2015/redirecting-all-kinds-of-stdout-in-python/

test/metric_learn_test.py
```diff
@@ -70,6 +73,40 @@ def test_iris(self):
...
+    lines = re.split("\n+", out)
+    objectives = [re.search("\d* (?:(\d*.\d*))[ | -]\d*.\d*", s)
```
perimosocordiae: Add a comment explaining this regular expression, with an example of what it should be matching.

metric_learn/lmnn.py
```diff
@@ -158,14 +162,25 @@ def fit(self, X, y):
...
+        objective_old = objective
+        L_old = L.copy()
+        G_old = G.copy()
+        df_old = df.copy()
+        a1_old = [a.copy() for a in a1]
+        a2_old = [a.copy() for a in a2]
+      # we update L and will see in the next iteration if it does indeed
+      # better
+      L -= learn_rate * 2 * L.dot(G)
+      learn_rate *= 1.01
```
perimosocordiae: This seems wrong to me, though it was also wrong before this PR.

Reading through the Shogun implementation [here](http://shogun-toolbox.org/api/latest/LMNN_8cpp_source.html) and [here](http://shogun-toolbox.org/api/latest/LMNNImpl_8cpp_source.html), they don't do any rolling back.
They do the `L` update unconditionally in `gradient_step`, then compute the objective value for the current iteration, then do the learning rate update based on the change in objective.

In one of the reference Matlab implementations [here](https://bitbucket.org/mlcircus/lmnn/src/5b49cafaeb9aacf19571f0b334178830cf712056/lmnn2/lmnn2.m?at=default&fileviewer=file-view-default#lmnn2.m-234) they do the `L` update first, then optionally roll back to a saved state when updating the step size.

--
So I think the correct fix would be to move the `L` update to the `# do the gradient update` section, after computing the new `G`, using the existing learning rate. Then, if the objective didn't improve we can halve the learning rate and roll back to the last good state (including `L` and `G`). Otherwise, we just grow the learning rate by 1% and carry on.

bellet: Inverting the order of things as suggested would also improve the readability of the code I think

test/metric_learn_test.py
wdevazelhes: done

test/metric_learn_test.py
wdevazelhes: done

test/metric_learn_test.py
wdevazelhes: Agreed. Also I just found there exist something in pytest that seem to do the job quite nicely: https://docs.pytest.org/en/3.2.1/capture.html
But it breaks a bit the structure of unittest classes... If it is important to keep the previous structure I'll use the context manager 
Tell me what you think

test/metric_learn_test.py
wdevazelhes: done

metric_learn/lmnn.py
wdevazelhes: I just submitted a new commit inverting things, I'll comment the lines to explain the change

test/metric_learn_test.py
```diff
@@ -72,6 +74,37 @@ def test_iris(self):
       self.assertLess(csep, 0.25)
 
 
+def test_convergence_simple_example(capsys):
```
perimosocordiae: Where does capsys get passed in from?

test/metric_learn_test.py
```diff
@@ -72,6 +74,37 @@ def test_iris(self):
...
+  # LMNN should converge on this simple example, which it did not with
+  # this issue: https://github.com/metric-learn/metric-learn/issues/88
+  X, y = make_classification(random_state=0)
+  lmnn = python_LMNN(verbose=True)
+  lmnn.fit(X, y)
+  out, _ = capsys.readouterr()
+  assert "LMNN converged with objective" in out
+
+
+def test_no_twice_same_objective(capsys):
```
perimosocordiae: Should these be methods of `TestLMNN`?

metric_learn/lmnn.py
```diff
@@ -90,83 +90,46 @@ def fit(self, X, y):
...
+    # first iteration
+    G, objective, total_active, df, a1, a2 = (
+        self._loss_grad(L, dfG, impostors, 1, k, reg, target_neighbors, df, a1,
+                        a2))
+    for it in xrange(2, self.max_iter):
+      # we first try to find a value of L that has better objective than the
+      #  previous L, following the gradient:
+
+      # we want to enter the loop for the first try, with the original
+      # learning rate (hence * 2 since it will be / 2)
```
perimosocordiae: This is a bit confusing to read. How about a `while True:` loop that breaks on the `delta_obj > 0` condition at the end?

test/metric_learn_test.py
wdevazelhes: I read here (https://github.com/pytest-dev/pytest/issues/2504#issuecomment-308828149) that pytest fixtures cannot be integrated with unittest classes, so I extracted the tests from the class hierarchy. But I agree that it is not ideal. They propose a workaround in the link, so maybe it would be better ? (adding these lines to `TestLMNN`, include the test in `TestLMNN`, and replace capsys by self.capsys in the test)
```python
@pytest.fixture(autouse=True)
def capsys(self, capsys):
  self.capsys = capsys
```

test/metric_learn_test.py
wdevazelhes: I think it should be automatically found by pytest (as one of the integrated fixtures), and runned when running `test_convergence_simple_example`. I verified doing -v (verbose) when doing pytest and these tests are correctly passing when they should (and failing when modifying the error message)

metric_learn/lmnn.py
wdevazelhes: I agree it would improve readability
Done

wdevazelhes:
@perimosocordiae I adressed your comments, except for the ones about pytest functions being outside of the unittest class architecture
I was not sure what to do, is it OK to let it this way, or should I go for the option mentioned in https://github.com/metric-learn/metric-learn/pull/101#discussion_r207859308 ? 
Otherwise I think the PR is ready to merge

wdevazelhes:
Oups, you're right for the .deb file, I don't know how it got there :p

metric_learn/lmnn.py
```diff
@@ -90,83 +90,49 @@ def fit(self, X, y):
...
+    G, objective, total_active, df, a1, a2 = (
+        self._loss_grad(L, dfG, impostors, 1, k, reg, target_neighbors, df, a1,
+                        a2))
+
+    for it in xrange(2, self.max_iter):
+      # then at each iteration, we try to find a value of L that has better
+      # objective than the previous L, following the gradient:
+      while True:
+        # the next point next_L to try out is found by a gradient step
+        L_next = L - 2 * learn_rate * G
```
wdevazelhes: Here it should be `2*learn_rate*L.dot(G)`, not `2*learn_rate*G`... (see #201)

metric_learn/lmnn.py
```diff
@@ -90,83 +90,49 @@ def fit(self, X, y):
...
-      if delta_obj > 0:
-        # we're getting worse... roll back!
-        learn_rate /= 2.0
-        df = df_old
-        a1 = a1_old
-        a2 = a2_old
-        objective = objective_old
-      else:
-        # update L
-        L -= learn_rate * 2 * L.dot(G)
```
wdevazelhes: see comment https://github.com/metric-learn/metric-learn/pull/101/files#r288573007 and #201

PR #103, opened 2018-07-31
[MRG] FIX: use addition for starting while loop and checks NaN

### LINKED ISSUE #74: MMC: w_previous referenced before assignment
While running MMC with diagonal set to True, I get the following error.

mmc.py, line 277 in fit_diag_ 
UnboundLocalError: local variable 'w_previous' referenced before assignment.

wdevazelhes:
Fixes #74: uses addition rather than multiplication for getting the while loop started, and throw error when NaN are found

### DIFF
--- a/metric_learn/mmc.py
+++ b/metric_learn/mmc.py
@@ -20,7 +20,7 @@ from __future__ import print_function, absolute_import, division
 import numpy as np
 from six.moves import xrange
 from sklearn.metrics import pairwise_distances
-from sklearn.utils.validation import check_array, check_X_y
+from sklearn.utils.validation import check_array, check_X_y, assert_all_finite
 
 from .base_metric import BaseMetricLearner
 from .constraints import Constraints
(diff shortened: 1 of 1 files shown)


metric_learn/mmc.py
```diff
@@ -261,9 +261,9 @@ def _fit_diag(self, X, constraints):
       # search over optimal lambda
       lambd = 1  # initial step-size
       w_tmp = np.maximum(0, w - lambd * step)
-
       obj = np.dot(s_sum, w_tmp) + self.diagonal_c * self._D_objective(X, c, d, w_tmp)
-      obj_previous = obj * 1.1  # just to get the while-loop started
+      assert_all_finite(obj)
+      obj_previous = obj + 1.1  # just to get the while-loop started
```
perimosocordiae: `obj + 1` may be less confusing to readers here.

wdevazelhes: Agreed, done

Issue #104, opened 2018-08-01
MLKR optimization algorithm does not work as expected regarding iterations


wdevazelhes:
`MLKR` does not have the expected behaviour regarding iterations. Indeed, it returns always 0 iterations. (Setting the `max_iter` does not seem to change this). 

### Minimal Example
```python
from sklearn.datasets import make_regression
from metric_learn import MLKR

X, y = make_regression()
mlkr = MLKR(max_iter=5)
mlkr.fit(X, y)
print(mlkr.n_iter_)
```

### Expected result: 

`5` (or between 1 and 5)

### Actual result: 

`0`

### Environment
Linux-4.4.0-130-generic-x86_64-with-debian-stretch-sid
Python 3.4.5 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:47:47) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]
NumPy 1.15.0
SciPy 1.1.0
Scikit-Learn 0.19.2
Metric-Learn 0.4.0 (master (last commit: https://github.com/metric-learn/metric-learn/commit/faa240fd7469176036a91430ae6a0a45e627c94a))

perimosocordiae:
Strange. We just forward it on to scipy's `minimize` function, so I'm not sure what's happening.

wdevazelhes:
Turns out in the previous example the optimization result said it converged at first iteration, but trying with this example the behaviour is as expected:

```python
In [1]: 
   ...: import numpy as np
   ...: from metric_learn import MLKR
   ...: np.random.seed(0)
   ...: 
   ...: X = np.random.randn(5, 2)
   ...: y = np.random.randn(5, 1)
   ...: 
   ...: mlkr = MLKR(max_iter=3)
   ...: mlkr.fit(X, y)
   ...: print(mlkr.n_iter_)
   ...: 

```

Returns: 
`3`

Converging at first iteration in the previous example is still weird but should be due to numerical problems (see PR #111 comment: https://github.com/metric-learn/metric-learn/pull/111/comment#issuecomment-412071625)

Closing, since the real problem is tackled in PR #111

PR #105, opened 2018-08-01
[WIP] Add verbose to NCA and MLKR

### LINKED ISSUE #67: Addition of verbosity setting for all of the algorithms
Currently, none of the algorithms have a verbosity setting while fitting. It is useful to see some details when the fitting process is going on.

wdevazelhes:
Fixes #67.

This PR adds a verbose argument for `NCA` and `MLKR`. Here is an example on a snippet. 

```python
from sklearn.datasets import load_digits
from metric_learn import NCA
dataset = load_digits()
X, y = dataset.data, dataset.target
nca = NCA(verbose=True, max_iter=5)
nca.fit(X, y)
```
Returns: 

```pytb
[NCA]
[NCA]  Iteration      Objective Value    Time(s)
[NCA] ------------------------------------------
[NCA]          0         1.436318e+03       0.23
[NCA]          1         1.770346e+03       0.24
[NCA]          2         1.777018e+03       0.26
[NCA]          3         1.779409e+03       0.25
[NCA]          4         1.781304e+03       0.26
[NCA]          5         1.782404e+03       0.25
/home/will/Code/metric-learn-bis/metric_learn/nca.py:110: ConvergenceWarning: [NCA] NCA did not converge: b'STOP: TOTAL NO. of ITERATIONS REACHED LIMIT'
  cls_name, opt_result.message), ConvergenceWarning)
[NCA] Training took 1533143523.76s.
```

The code mostly comes from scikit-learn's NCA PR (https://github.com/scikit-learn/scikit-learn/pull/10058), and this verbose code in particular was taken mostly from scikit-learn's LMNN PR (https://github.com/scikit-learn/scikit-learn/pull/8602)

### TODO: 
- [ ] Fix the convergence warning test for `MLKR` (related to #104)

### DIFF
--- a/metric_learn/mlkr.py
+++ b/metric_learn/mlkr.py
@@ -7,11 +7,15 @@ algorithm can also be viewed as a supervised variation of PCA and can be used
 for dimensionality reduction and high dimensional data visualization.
 """
 from __future__ import division, print_function
+import time
+import sys
+import warnings
 import numpy as np
 from scipy.optimize import minimize
 from scipy.spatial.distance import pdist, squareform
 from sklearn.decomposition import PCA
 from sklearn.utils.validation import check_X_y
+from sklearn.exceptions import ConvergenceWarning
 
 from .base_metric import BaseMetricLearner
 
@@ -21,7 +25,7 @@ EPS = np.finfo(float).eps
 class MLKR(BaseMetricLearner):
   """Metric Learning for Kernel Regression (MLKR)"""
   def __init__(self, num_dims=None, A0=None, epsilon=0.01, alpha=0.0001,
-               max_iter=1000):
+               max_iter=1000, verbose=False):
     """
     Initialize MLKR.
 
@@ -41,12 +45,16 @@ class MLKR(BaseMetricLearner):
 
     max_iter: int, optional
         Cap on number of congugate gradient iterations.
+
+    verbose : bool, optional (default=False)
+        Whether to print progress messages or not.
     """
     self.num_dims = num_dims
     self.A0 = A0
     self.epsilon = epsilon
     self.alpha = alpha
     self.max_iter = max_iter
+    self.verbose = verbose
 
   def _process_inputs(self, X, y):
       self.X_, y = check_X_y(X, y)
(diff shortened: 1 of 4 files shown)


metric_learn/mlkr.py
```diff
@@ -79,34 +87,68 @@ def fit(self, X, y):
...
-  np.fill_diagonal(K, 1)
-  W = 2 * K * (np.outer(ydiff, ydiff) / denom)
-  # note: this is the part that the matlab impl drops to C for
-  M = (dX.T * W.ravel()).dot(dX)
-  grad = 2 * A.dot(M)
-  return cost, grad.ravel()
+  def _loss(self, flatA, X, y, dX):
+
+    if self.n_iter_ == 0:
+      if self.verbose:
```
perimosocordiae: I'd combine these into a single condition.

metric_learn/mlkr.py
```diff
@@ -79,34 +87,68 @@ def fit(self, X, y):
...
+        header_fields = ['Iteration', 'Objective Value', 'Time(s)']
+        header_fmt = '{:>10} {:>20} {:>10}'
+        header = header_fmt.format(*header_fields)
+        cls_name = self.__class__.__name__
+        print('[{}]'.format(cls_name))
+        print('[{}] {}\n[{}] {}'.format(cls_name, header, cls_name,
+                                        '-' * len(header)))
```
perimosocordiae: Named format slots might make this easier to read. That is, like the "by name" examples here: https://docs.python.org/2/library/string.html#format-examples

metric_learn/mlkr.py
```diff
@@ -79,34 +87,68 @@ def fit(self, X, y):
...
+
+    t_funcall = time.time()
```
perimosocordiae: I'm not a big fan of the `t_` prefix for times. How about `start_time`?

metric_learn/mlkr.py
wdevazelhes: Agreed, done (except for the header field because there I thought keyword arguments would be more redundant than really helping reading , but tell me if you don't think so)

metric_learn/mlkr.py
wdevazelhes: I agree, done

wdevazelhes:
I just noticed that the Conjugate Gradient method (used in `MLKR`) has a different way to count iterations: unlike `L-BFGS-B`, one iteration is not equal to one function call, so the way to print iteration number, objective and gradient, as done in this PR does not work for `MLKR`. 

I was thinking I could maybe instead print the cost function for each function evaluation instead of each iteration, for `MLKR` ? 

Or we could use `L-BFGS-B` for `MLKR` too ? @perimosocordiae was there a particular reason to use Conjugate Gradient ? 
Looking into the documentation of  Conjugate Gradient https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin_cg.html, at `Notes` they say that it works better if the function has a global minima, which I don't think there is, right ? (any rotation of an optimal transformation `A` would be optimal too because conserving distances)
@bellet any thought on that ? 

Note: here is a thread that compares the two methods: https://scicomp.stackexchange.com/questions/507/bfgs-vs-conjugate-gradient-method

perimosocordiae:
The [MLKR paper](http://proceedings.mlr.press/v2/weinberger07a/weinberger07a.pdf) says this about the optimization problem, after briefly describing a gradient descent algorithm:

> Of course, other methods for minimizing (3) such as conjugate gradient, stochastic gradient or BFGS may also be used and might lead to faster convergence results.

The initial MLKR implementation by @dsquareindia used a hand-rolled gradient descent, and I later converted this to use `scipy.optimize.minimize` probably without thinking too hard about the choice of method.

So in summary: if you think L_BFGS-B will work better, feel free to try it out! :+1:

bellet:
L-BFGS-B should also work well (maybe slightly better) at the cost of slightly higher memory. Maybe you can indeed give it a try, this would make it more consistent with NCA

wdevazelhes:
Allright, we can go for L-BFGS-B for now (it can easily be changed in the future)
I just added a commit which also uses less features for the test of `make_regression`, making the task easier to train on hence making MLKR do really 2 iterations (and hitting max_iter in the test), and not stopping weirdly at the beginning (see https://github.com/metric-learn/metric-learn/issues/104)

perimosocordiae:
Changing the underlying optimizer is a bit of a compatibility break, too, so I'm fine with making it explicit by changing the keyword arguments.

wdevazelhes:
I just realized in fact `L-BFGS-B` also has an `epsilon` parameter, so maybe we want to keep it ? And also add it to `NCA` consequently ?

bellet:
From my understanding, `epsilon` is useless here because you already provide a function to compute the exact gradient (this is also true for NCA)

wdevazelhes:
Oh yes that's right
I'll provide a deprecation warning similar to the one in NCA and I think we'll be good to merge

wdevazelhes:
I just realized that `MLKR` is introduced in master (will be introduced in the next release), so I guess actually we don't need any deprecation ?

bellet:
Makes sense indeed :-) But need to fix the travis failure

wdevazelhes:
Done

PR #108, opened 2018-08-07
[MRG] Update __init__.py to allow printing version

wdevazelhes:
this code should print the version number:

```python
import metric_learn
print(metric_learn.__version__)
```

However it doesn't work. 

This modification allows the version number to be accessed like this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 from setuptools import setup
+import metric_learn
 
-version = "0.4.0"
+version = metric_learn.__version__
 setup(name='metric-learn',
       version=version,
       description='Python implementations of metric learning algorithms',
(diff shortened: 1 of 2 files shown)


wdevazelhes:
I agree that it would be better not to have it in several locations 
I changed the one in `__init__.py` to be found from the one in `setup.py`, like in scikit-learn: https://github.com/scikit-learn/scikit-learn/blob/76d763be80dcd572706ac91708f7661f96080af9/setup.py#L40-L42, I guess it works too ? I am fine with both though

wdevazelhes:
Great, thanks ! 
Merging

setup.py
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 from setuptools import setup
+import metric_learn
```
perimosocordiae: Wouldn't this cause trouble for users who clone the fresh repo and run `python setup.py install` ?

wdevazelhes:
Reopened in #110

PR #110, opened 2018-08-09
Add version number in __init__.py, new PR

wdevazelhes:
This is the new PR to add version number in __init__.py
I saw it from the umap package https://github.com/lmcinnes/umap/tree/master/umap
What do you think ? 
I tried and it seemed to solve the problem in the previous PR (#108)

### DIFF
--- a/metric_learn/__init__.py
+++ b/metric_learn/__init__.py
@@ -11,3 +11,7 @@ from .lfda import LFDA
 from .rca import RCA, RCA_Supervised
 from .mlkr import MLKR
 from .mmc import MMC, MMC_Supervised
+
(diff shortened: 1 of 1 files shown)


wdevazelhes:
Here is another proposal, which puts the version in a `_version.py` file.
Another possibility would be to put it in a non Python file, do the same by reading the file, and adding the file to a MANIFEST.in (see https://docs.python.org/2/distutils/sourcedist.html#the-manifest-in-template)

perimosocordiae:
This works well enough, so I merged it as-is. If we want something fancier later, we can change it.

PR #113, opened 2018-08-17
[MRG] Improve gradient computation in NCA and MLKR

wdevazelhes:
Uses [np.linalg.multi_dot](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.multi_dot.html) in gradient computations in `NCA` and `MLKR`. This should provide optimal performance, no matter the values of `n_components`, `n_features`, and `n_samples`.

### DIFF
--- a/metric_learn/mlkr.py
+++ b/metric_learn/mlkr.py
@@ -10,10 +10,9 @@ from __future__ import division, print_function
 import numpy as np
 from sklearn.utils.fixes import logsumexp
 from scipy.optimize import minimize
-from scipy.spatial.distance import pdist, squareform
 from sklearn.decomposition import PCA
+from sklearn.metrics import pairwise_distances
 from sklearn.utils.validation import check_X_y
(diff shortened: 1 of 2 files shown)


wdevazelhes:
I just realized that in fact in `NCA` one of the reason why we replaced A.dot(X.T) by X_embedded.T was that we already had it computed when computing the cost, so I'll update the PR to reuse it in NCA as it was before, and update MLKR in the same way

wdevazelhes:
Turns out in this case where we reuse `X_embedded` there is no need anymore to use `np.linalg.mutli_dot` since the optimal order of the products is to do first `X_embedded.T.dot(W_sym)` and then doing this dot `X`. (it makes a cost of <b>`n_components`</b>`* n_samples**2 + n_components * n_samples * n_features` instead of <b>`n_features`</b>`* n_samples **2 + n_components * n_samples * n_features`) 

I think the PR is ready to merge if travis passes

PR #114, opened 2018-08-17
[MRG] Add MLKR and MMC to doc index

wdevazelhes:
Add newly introduced MMC and MLKR to the documentation.


wdevazelhes:
If the current description of MMC and MLKR is good, I think we can merge.

PR #115, opened 2018-08-20
[MRG] Add documentation for supervised classes and add __init__ docstrings to doc

### LINKED ISSUE #22: Update docs for *_Supervised classes
PR #19 introduced supervised variants of some algorithms, along with a module for generating constraints from supervised labels. These will need to be documented before the next release.

wdevazelhes:
Fixes #22

Here is a small proposal for adding some documentation for supervised classes.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -47,7 +47,7 @@ particular form of input data.
    metric_learn.rca
    metric_learn.mmc
 
-Note that each semi-supervised algorithm has a supervised version of the form
+Note that each weakly-supervised algorithm has a supervised version of the form
 `*_Supervised` where similarity constraints are generated from
 the labels information and passed to the underlying algorithm.
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -15,9 +15,11 @@ metric learning algorithms.
 
 Supervised Algorithms
 ---------------------
-Supervised metric learning algorithms take as inputs points `X` and targets
-`y`, and learn to make points from the same class close to each other, and
-points from different classes far away from each other.
+Supervised metric learning algorithms take as inputs points `X` and target
+labels `y`, and learn to make points from the same class (for classification)
+or with close target value (for regression) close to each other, and points
+from different classes or with distant target values far away from each
+other.
 
 .. toctree::
    :maxdepth: 1
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -13,20 +13,40 @@ metrics.
 This package contains efficient Python implementations of several popular
 metric learning algorithms.
 
+Supervised Algorithms
+---------------------
+Supervised metric learning algorithms take as inputs points `X` and targets
+`y`, and learn to make points from the same class close to each other, and
+points from different classes far away from each other.
+
 .. toctree::
-   :caption: Algorithms
    :maxdepth: 1
 
    metric_learn.covariance
    metric_learn.lmnn
-   metric_learn.itml
-   metric_learn.sdml
-   metric_learn.lsml
    metric_learn.nca
    metric_learn.lfda
+
+Semi-Supervised Algorithms
+--------------------------
+Semi-supervised algorithms do the same, but work on a weaker information about
+similarities between data points, for instance pairs of similar and dissimilar
+points. See each of the following algorithm for its particular form of input
+data.
+
+.. toctree::
+   :maxdepth: 1
+
+   metric_learn.itml
+   metric_learn.lsml
+   metric_learn.sdml
    metric_learn.rca
 
-Each metric supports the following methods:
+Note that each semi-supervised algorithm has a supervised version of the form
+`*_Supervised` where information about similarities is generated from
+the labels information and passed to the underlying algorithm.
+
+Each metric learning algorithm supports the following methods:
 
 -  ``fit(...)``, which learns the model.
 -  ``transformer()``, which returns a transformation matrix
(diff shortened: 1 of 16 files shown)


wdevazelhes:
Docstrings in `__init__.py` of classes currently does not appear in the doc, I wonder if we should put them in the class docstring (below the class name), or rather add some directives to sphinx not to ignore them

doc/index.rst
```diff
@@ -13,22 +13,42 @@ metrics.
 This package contains efficient Python implementations of several popular
 metric learning algorithms.
 
+Supervised Algorithms
+---------------------
+Supervised metric learning algorithms take as inputs points `X` and targets
+`y`, and learn to make points from the same class close to each other, and
```
perimosocordiae: target class labels

perimosocordiae:
+1 for telling Sphinx to include the `__init__` docstrings.

doc/index.rst
wdevazelhes: maybe target labels only, to include `MLKR` ?

perimosocordiae: :+1:

wdevazelhes: Just realized the rest of the sentence ("[...]from the same class[...]") didn't include `MLKR` so I'll update this too

wdevazelhes:
> +1 for telling Sphinx to include the __init__ docstrings.

According to [this comment](https://stackoverflow.com/a/5599712/5635843) there are several ways to do that, I went for the one that includes the `special_members` because I found it easy to read and quite clean, but maybe option 1 (to include some code in conf.py), would be better, avoiding to add a line of code in every file ?

doc/index.rst
```diff
@@ -13,22 +13,44 @@ metrics.
...
    metric_learn.covariance
    metric_learn.lmnn
-   metric_learn.itml
-   metric_learn.sdml
-   metric_learn.lsml
    metric_learn.nca
    metric_learn.lfda
+   metric_learn.mlkr
+
+Semi-Supervised Algorithms
```
bellet: I would rather use something like "Weakly supervised algorithms" to avoid confusion with the classic semi-supervised setting (i.e., a subset of points are labeled and other are not).

doc/index.rst
```diff
@@ -13,22 +13,44 @@ metrics.
 This package contains efficient Python implementations of several popular
 metric learning algorithms.
 
+Supervised Algorithms
+---------------------
+Supervised metric learning algorithms take as inputs points `X` and target
+labels `y`, and learn to make points from the same class (for classification)
```
bellet: learn a distance metric that makes points...

doc/index.rst
```diff
@@ -13,22 +13,44 @@ metrics.
...
-   metric_learn.itml
-   metric_learn.sdml
-   metric_learn.lsml
    metric_learn.nca
    metric_learn.lfda
+   metric_learn.mlkr
+
+Semi-Supervised Algorithms
+--------------------------
+Semi-supervised algorithms do the same, but work on a weaker information about
```
bellet: maybe avoid "do the same" to have a more self consistent section
quick suggestion: "Weakly supervised algorithms work on weaker information about the data points than supervised algorithms. Rather than labeled points, they take as input similarity judgments on tuples of data points, for instance pairs of similar and dissimilar points. Refer to the documentation of each algorithm for its particular form of input data."

metric_learn/itml.py
```diff
@@ -145,16 +145,21 @@ class ITML_Supervised(ITML):
...
     Parameters
     ----------
     gamma : float, optional
         value for slack variables
     max_iter : int, optional
     convergence_threshold : float, optional
-    num_labeled : int, optional
-        number of labels to preserve for training
+    num_labeled : int, optional (default=np.inf)
+        number of labels to preserve for training. If np.inf (default),
```
bellet: what is meant by "preserve for training" needs to be clarified?

metric_learn/lsml.py
```diff
@@ -134,16 +134,22 @@ def _gradient(self, metric):
 class LSML_Supervised(LSML):
   def __init__(self, tol=1e-3, max_iter=1000, prior=None, num_labeled=np.inf,
                num_constraints=None, weights=None, verbose=False):
-    """Initialize the learner.
+    """Initialize the supervised version of `LSML`.
+
+    `LSML_Supervised` creates quadruplets from labeled samples by taking two
+    samples from the same class, and two samples from different classes, and
+    concatenating them. This way it builds quadruplets where the two first
```
bellet: ", and concatenating them" seems useless

doc/index.rst
wdevazelhes: I agree
Done

doc/index.rst
wdevazelhes: Yes I agree this is clearer
Done

metric_learn/itml.py
wdevazelhes: what about: "number of labeled points to keep for building pairs. Extra 
        labeled points will be considered unlabeled, and ignored as such.
        Use np.inf (default) to use all labeled points." ?

metric_learn/lsml.py
wdevazelhes: Agreed
Done

doc/index.rst
```diff
@@ -13,22 +13,45 @@ metrics.
...
+
+   metric_learn.itml
+   metric_learn.lsml
+   metric_learn.sdml
    metric_learn.rca
    metric_learn.mmc
-   metric_learn.mlkr
 
-Each metric supports the following methods:
+Note that each semi-supervised algorithm has a supervised version of the form
```
bellet: weakly

metric_learn/itml.py
bellet: In fact IMO this parameter is useless. I will open an issue

PR #117, opened 2018-08-21
[MRG] Add preprocessor option

wdevazelhes:
This PR adds an argument `preprocessor` to weakly supervised algorithms initialization: an option that allows them to accept two types of input: either 3D arrays of tuples of points as before, or 2D arrays of indices/identifiers of points. In the latter case, the preprocessor given as input to the Metric Learner will allow to retreive points from identifiers. The preprocessor is basically a callable that is called on a list of identifiers (or just on one identifier ? we'll have to decide), and that returns the points associated to this identifiers. 
If instead of a callable the user provides an array X, it will automatically create the "indexing" preprocessor (i.e. a preprocessor such that preprocessor(id) returns X[id]). We could also imagine other shortcuts like this (for instance the user could provide a string of a root folder containing images and then it would create a preprocessor such that preprocessor("img2.png") would return the vector associated to the image located at "rootfolder/img2.png") 

Note: This PR branched from MahalanobisMixin PR, so as soon as MahalanobisMixin is merged the diff should become more readable.

#### TODO: 

- [x] Add comments and more info to this PR
- [x] Add tests: in progress: add tests to check that the output format of `check_input` is as specified
-<strike> [ ] Add documentation </strike> -> probably in another PR
- [x] Make it simpler with unified `check_input` function
- [x] Refactor the `check_input` function and its tests to be cleaner
- [x] Write docstrings
- [x] Refactor tests to have the list of metric learners at one place
- [x] Fix the linalg bug


metric_learn/base_metric.py
```diff
@@ -1,31 +1,76 @@
...
 
     Parameters
     ----------
-    pairs : `numpy.ndarray`, shape=(n_samples, 2, n_features)
-      3D array of pairs.
+    preprocessor : array-like, shape=(n_samples, n_features) or callable
+        The preprocessor to call to get tuples from indices. If array-like,
+        tuples will be gotten like this: X[indices].
+    """
+    self.preprocessor = preprocessor
```
bellet: couldn't the `self.check_preprocessor()` be called here (or in the relevant meta classes) once and for all? Instead of calling it in various places in the downstream classes?

metric_learn/base_metric.py
```diff
@@ -1,31 +1,76 @@
...
+        The score of every pair.
+      """
+
+  def check_preprocessor(self):
+    if _is_arraylike(self.preprocessor):
+      self.preprocessor_ = SimplePreprocessor(self.preprocessor)
+    else:
+      self.preprocessor_ = self.preprocessor
+
+  def metric(self):
```
bellet: why are these specific `metric` and `transformer` methods added? this is specific to Mahalanobis metric learner

metric_learn/base_metric.py
```diff
@@ -1,31 +1,76 @@
...
+
+    L = cholesky(M).T
+
+    Returns
+    -------
+    L : upper triangular (d x d) matrix
     """
+    return cholesky(self.metric()).T
 
+  def format_input(self, input):
```
bellet: choose more concrete name for `input` and the function name (preprocess instead of format?)

metric_learn/base_metric.py
wdevazelhes: Yes I agree it is very redundant
Maybe we should create an abstract `fit` for a class `WeaklySupervisedMixin` and put it here, because to be compatible with scikit-learn `set_params` used in grid-searches etc, I think the `__init__` should not contain any processing of the input (like `if preprocessor [...]`  `else` etc), and they should all be done in `fit`. (Even if this was in order to use `set_params` to do hyper-parameter search in grid-search, and we shouldn't do grid search on a preprocessor too much (except maybe if we use the preprocessor as a feature extractor and we want to test different feature extractors))

metric_learn/base_metric.py
wdevazelhes: Yes indeed, I messed up with the merging, I'll remove them

wdevazelhes:
> A general comment: we should try not to "materialize" (i.e., form the feature representation) of all pairs at the same time, both when fitting and when scoring/predicting/etc. This would avoid memory errors. Instead, one should compute over batches instead when too big (so basically never call `format_input` on too big arrays). For this we can probably take inspiration from what is done in sklearn. 

Yes I agree, I should have put in in the TODO (done now: https://github.com/metric-learn/metric-learn/issues/91#issue-324012762)

> For clarity, this could be done in a separate PR though.

I agree, this can be for the next PR

metric_learn/base_metric.py
wdevazelhes: After thinking about it I am not sure what to do because checking the preprocessor in an abstract `_fit` in say `BaseMetricLearner` would mean calling `super()._fit` in `_BaseMMC._fit` and in `NCA.fit` for instance (btw the class call and super call would have different names for NCA: `fit` and `_fit`, respectively, but maybe this is not a problem), which would amount to the same lines of code added as before. But maybe this is a better practice and would pave the way if we have to mutualize other stuff in `fit` ? Or maybe for now it adds a little complexity with super calls, instead of an explicit `check_preprocessor` ? I'm OK with both strategies

metric_learn/_util.py
```diff
@@ -12,39 +14,145 @@ def vector_norm(X):
...
+  """Helper function to check that the number of points in each tuple is
+  equal to t (e.g. 2 for pairs), and raise a `ValueError` otherwise"""
+  if t is not None and tuples.shape[1] != t:
+    msg_t = (("Tuples of {} element(s) expected{}. Got tuples of {} "
+             "element(s) instead (shape={}):\ninput={}.\n")
+             .format(t, context, tuples.shape[1], tuples.shape, tuples))
+    raise ValueError(msg_t)
+
+
+class SimplePreprocessor():
```
bellet: `ArrayIndexer` maybe?

metric_learn/base_metric.py
bellet: In fact the right place for the call to `check_preprocessor` should probably be in a common `_prepare_inputs` function (see more detailed comment below).

bellet:
Looking at the code now, the call to `check_preprocessor` is sometimes at the beginning of `fit`, and sometimes in a custom function that checks/prepares the inputs (they come in different names, like `_process_inputs`,  `_process_pairs`, `_prepare_quadruplets`, `_process_data`, `_prepare_pairs`). I think this is what makes things redundant and quite messy.

There should probably be a common function that would make usual checks (e.g., for duplicates) for all input cases, and would call `check_preprocessor`. Again, this is probably better to do in a separate PR. I will open some issues relating to this so we do not forget.

Another point: if we want to have preprocessors for supervised algorithms (which I think we do, e.g. the use case of image data loaded from some directory), then the processor should actually be called to  should be used on the input data `X` - I do not think it is the case right now.

Finally, there was a debate as to where to place the preprocessor in `init` or `fit`. In `fit` it would be more natural, but there was a problem with sklearn API in this case, right?

metric_learn/_util.py
wdevazelhes: Yes indeed

wdevazelhes:
Tests are passing ! But I'll just do some more refactoring to simplify/clean the code (as I added in the TODO list at the top), pep8 corrections etc, before setting the PR to MRG for the final review and merge

bellet:
I guess this PR also addressed #123

bellet:
@wdevazelhes how are you doing on refactoring etc? Would be great to set this to MRG soon for final reviews

wdevazelhes:
I think it should be ready for review/merge now. 

To sum it up the PR allows all estimators to use a preprocessor, and the input will be converted at the beginning of every function (fit, predict, etc) into a 2D numpy array of samples or a 3D numpy array of tuples, using the preprocessor if needed, thanks to the `check_input` function, which is the main contribution of the PR. This function tries to reuse as much as possible scikit learn routines for checking inputs (check_array and check_X_y), and to raise the appropriate error messages if needed, while converting at the same time the input to the appropriate format.

Remark: I tried to fix the PEP8 as much as possible, but only in the diff of this PR, to make the review easier. So there remains some PEP8 errors outside of the diff, but they will be changed in other PRs. Sometimes there are PEP8 errors even in the diff but they should be for coherence with the rest (ex: wrong docstring syntax with missing spaces or newlines so as not to change the whole docstring).

bellet:
@perimosocordiae @terrytangyuan your feedback is very welcome on this!!!

metric_learn/lsml.py
```diff
@@ -31,18 +33,19 @@ def __init__(self, tol=1e-3, max_iter=1000, prior=None, verbose=False):
...
+    super(_BaseLSML, self).__init__(preprocessor)
 
   def _prepare_quadruplets(self, quadruplets, weights):
-    # for now we check_array and check_tuples but we should only
-    # check_tuples in the future (with enhanced check_tuples)
-    quadruplets = check_array(quadruplets, accept_sparse=False,
-                              ensure_2d=False, allow_nd=True)
-    quadruplets = check_tuples(quadruplets)
+    quadruplets = self._prepare_inputs(quadruplets,
+                                       type_of_inputs='tuples')
```
terrytangyuan: Maybe move `type_of_inputs` to a constant class otherwise it's easy to make mistakes

test/test_sklearn_compat.py
```diff
@@ -68,5 +81,343 @@ def test_mmc(self):
...
+                                              same_length=True,
+                                              random_state=RNG)
+  return X, pairs
+
+
+def build_classification(preprocessor):
+  # builds a toy classification problem
+  X, y = shuffle(*make_blobs(), random_state=RNG)
+  X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=RNG)
+  return (X, X, y, X_train, X_test, y_train, y_test, preprocessor)
```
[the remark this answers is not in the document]
terrytangyuan: Agreed. It's a bit unclear here.

metric_learn/lsml.py
terrytangyuan: Maybe move possible values of `type_of_inputs` to a constant class / enum.  Otherwise it's easy to make mistakes?

test/test_sklearn_compat.py
```diff
@@ -68,5 +81,343 @@ def test_mmc(self):
...
+    estimator.num_dims = 1
+  estimator.fit(tuples, y)
+  for method in ["predict", "decision_function", "predict_proba"]:
+    if hasattr(estimator, method):
+      dict_before = estimator.__dict__.copy()
+      getattr(estimator, method)(tuples)
+      assert estimator.__dict__ == dict_before, \
+          ("Estimator changes __dict__ during %s"
+           % method)
+  for method in ["transform"]:
```
terrytangyuan: If the loop is not needed now, we should probably remove it.

test/test_sklearn_compat.py
```diff
@@ -68,5 +81,343 @@ def test_mmc(self):
...
+    if hasattr(estimator, method):
+      dict_before = estimator.__dict__.copy()
+      # we transform only 2D arrays (dataset of points)
+      getattr(estimator, method)(X)
+      assert estimator.__dict__ == dict_before, \
+          ("Estimator changes __dict__ during %s"
+           % method)
```
terrytangyuan: This can be refactored as it appears above as well

test/test_sklearn_compat.py
```diff
@@ -68,5 +81,343 @@ def test_mmc(self):
...
+      " %s changed" % ', '.join(attrs_changed_by_fit))
+
+
+def _get_args(function, varargs=False):
+    """Helper to get function arguments"""
+
+    try:
+        params = signature(function).parameters
+    except ValueError:
+        # Error on builtin C function
```
terrytangyuan: Could you clarify this a bit more?

metric_learn/_util.py
```diff
@@ -12,39 +14,264 @@ def vector_norm(X):
...
   ----------
-  tuples : object
-    The tuples to check.
+  input : array-like
+    The input data array to check.
+
+  y : array-like
+    The input labels array to check.
+
+  preprocessor : callable (default=None)
```
terrytangyuan: Wrap `None`s with back ticks?

metric_learn/_util.py
```diff
@@ -12,39 +14,264 @@ def vector_norm(X):
     return np.linalg.norm(X, axis=1)
 
 
-def check_tuples(tuples):
-  """Check that the input is a valid 3D array representing a dataset of tuples.
-
-  Equivalent of `check_array` in scikit-learn.
+def check_input(input, y=None, preprocessor=None,
```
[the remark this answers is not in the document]
wdevazelhes: Agreed, what about `input_data` ?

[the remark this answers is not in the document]
bellet: Sure

metric_learn/_util.py
```diff
@@ -12,39 +14,264 @@ def vector_norm(X):
...
+    The type of inputs to check. If 'classic', the input should be
+    a 2D array-like of points or a 1D array like of indicators of points. If
+    'tuples', the input should be a 3D array-like of tuples or a 2D
+    array-like of indicators of tuples.
+
+  accept_sparse : `bool`
+    Set to true to allow sparse inputs (only works for sparse inputs with
+    dim < 3).
+
+  t : int
```
perimosocordiae: Can we give `t` a more descriptive name? How about `tuple_size`?

metric_learn/_util.py
```diff
@@ -12,39 +14,264 @@ def vector_norm(X):
...
+                        accept_sparse=True, dtype=None,
+                        ensure_min_features=0, ensure_min_samples=0)
+  else:
+    input, y = check_X_y(input, y, ensure_2d=False, allow_nd=True,
+                         copy=False, force_all_finite=False,
+                         accept_sparse=True, dtype=None,
+                         ensure_min_features=0, ensure_min_samples=0,
+                         multi_output=multi_output,
+                         y_numeric=y_numeric)
+    # we try to allow the more possible stuff here
```
perimosocordiae: What does this comment mean?

metric_learn/_util.py
```diff
@@ -12,39 +14,264 @@ def vector_norm(X):
...
+  err_msg = ('{expected_input} expected'
+             '{context}{additional_context}. Found {found_size}D array '
+             'instead:\ninput={input}. Reshape your data'
+             '{possible_preprocessor}.\n')
+  raise ValueError(err_msg.format(**err_args))
+
+
+def preprocess_tuples(tuples, preprocessor):
+  print("Preprocessing tuples...")
+  tuples = np.column_stack([preprocessor(tuples[:, i])[:, np.newaxis] for
```
perimosocordiae: I don't think you need to do the `np.newaxis` stuff here, as `np.column_stack` will accept 1d inputs.

metric_learn/base_metric.py
```diff
@@ -26,6 +37,51 @@ def score_pairs(self, pairs):
...
+
+    y: `numpy.ndarray` (optional)
+      The checked input labels array.
+    """
+    self.check_preprocessor()
+    return check_input(X, y,
+                       type_of_inputs=type_of_inputs,
+                       preprocessor=self.preprocessor_,
+                       estimator=self,
+                       t=self._t if hasattr(self, '_t') else None,
```
perimosocordiae: `t=getattr(self, '_t', None)`

metric_learn/base_metric.py
```diff
@@ -204,36 +274,29 @@ def predict(self, quadruplets):
...
-    -------
-    decision_function : `numpy.ndarray` of floats, shape=(n_constraints,)
-      Metric differences.
-    """
-    quadruplets = check_tuples(quadruplets)
-    return (self.score_pairs(quadruplets[:, :2, :]) -
-            self.score_pairs(quadruplets[:, 2:, :]))
+    # we broadcast with ... because here we allow quadruplets to be
+    # either a 3D array of points or 2D array of indices
+    return (self.score_pairs(quadruplets[:, :2, ...]) -
```
perimosocordiae: Putting an Ellipsis at the end of an indexing expression has no effect.

metric_learn/_util.py
```diff
@@ -12,39 +14,264 @@ def vector_norm(X):
...
+    input, y = check_X_y(input, y, ensure_2d=False, allow_nd=True,
+                         copy=False, force_all_finite=False,
+                         accept_sparse=True, dtype=None,
+                         ensure_min_features=0, ensure_min_samples=0,
+                         multi_output=multi_output,
+                         y_numeric=y_numeric)
+    # we try to allow the more possible stuff here
+  preprocessor_has_been_applied = False
+
+  if type_of_inputs == 'classic':
```
perimosocordiae: This function is pretty long. Maybe we could break up the `type_of_inputs == 'classic'` and `'tuples'` cases into helper functions? They could return a boolean indicating if the preprocessor was applied, which would also help avoid the manual toggling here.

metric_learn/_util.py
wdevazelhes: Right thanks, I'll note it down also for changing it in other parts of the code in some next PR.

metric_learn/_util.py
```diff
@@ -12,39 +14,264 @@ def vector_norm(X):
...
+
+  context = make_context(estimator)
+
+  args_for_sk_checks = dict(accept_sparse=accept_sparse,
+                            dtype=dtype, order=order,
+                            copy=copy, force_all_finite=force_all_finite,
+                            ensure_min_samples=ensure_min_samples,
+                            ensure_min_features=ensure_min_features,
+                            warn_on_dtype=warn_on_dtype, estimator=estimator)
+  if y is None:
```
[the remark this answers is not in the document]
wdevazelhes: Agreed, will do

metric_learn/_util.py
wdevazelhes: I'll replace it by a more detailed explanation of what is above (running check_array with fixed permissive arguments to convert input_data to numpy.ndarray to be able to do further checks)

metric_learn/_util.py
```diff
@@ -12,39 +14,264 @@ def vector_norm(X):
...
+        make_error_input(101, input, context)
+
+  elif type_of_inputs == 'tuples':
+    if input.ndim == 2:
+      if preprocessor is not None:
+        input = preprocess_tuples(input, preprocessor)
+        preprocessor_has_been_applied = True
+      else:
+        make_error_input(201, input, context)
+    elif input.ndim == 3:  # we should check_num_features which is not checked
```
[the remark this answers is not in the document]
wdevazelhes: Yes, in the following `if ensure_min_features >0:` statement. But I'll change the comment because I agree it tells the contrary :p

metric_learn/_util.py
wdevazelhes: Agreed, will do

metric_learn/_util.py
wdevazelhes: Actually the output of preprocessor(1d input of shape (n_constraints,)) is a 2d input (of shape (n_constraints, n_features). This is the way I found to concatenate t=`tuple_size` 2d arrays into a 3d array of shape (n_constraints, tuple_size, n_features), but there may be a better way

metric_learn/base_metric.py
wdevazelhes: Nice, thanks

metric_learn/base_metric.py
```diff
@@ -204,36 +274,29 @@ def predict(self, quadruplets):
...
     Returns
     -------
     prediction : `numpy.ndarray` of floats, shape=(n_constraints,)
       Predictions of the ordering of pairs, for each quadruplet.
     """
-    quadruplets = check_tuples(quadruplets)
+    quadruplets = check_input(quadruplets, type_of_inputs='tuples',
+                              preprocessor=self.preprocessor_,
+                              estimator=self, t=self._t)
+    # we broadcast with ... because here we allow quadruplets to be
```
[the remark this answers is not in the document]
wdevazelhes: Right, thanks

metric_learn/base_metric.py
```diff
@@ -204,36 +274,29 @@ def predict(self, quadruplets):
...
+    # either a 3D array of points or 2D array of indices
     return np.sign(self.decision_function(quadruplets))
 
   def decision_function(self, quadruplets):
-    """Predicts differences between sample distances in input quadruplets.
```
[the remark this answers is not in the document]
wdevazelhes: Thanks, my mistake, I must have deleted it when changing the function

metric_learn/base_metric.py
wdevazelhes: Thanks, I don't know why I added them :p

metric_learn/lmnn.py
```diff
@@ -252,7 +259,8 @@ class LMNN(_base_LMNN):
     """
 
     def fit(self, X, y):
-      self.X_, y = check_X_y(X, y, dtype=float)
+      self.X_, y = self._prepare_inputs(X, y, dtype=float)
+      self.X_ = self.X_, preprocessor=self.preprocessor_
```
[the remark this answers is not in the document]
wdevazelhes: Thanks, it was a typo, fixed

metric_learn/lmnn.py
```diff
@@ -43,13 +46,17 @@ def __init__(self, k=3, min_iter=50, max_iter=1000, learn_rate=1e-7,
     self.convergence_tol = convergence_tol
     self.use_pca = use_pca
     self.verbose = verbose
+    super(_base_LMNN, self).__init__(preprocessor)
 
 
 # slower Python version
 class python_LMNN(_base_LMNN):
 
   def _process_inputs(self, X, labels):
```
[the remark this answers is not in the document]
wdevazelhes: You mean for the python_LMNN inputs ? Like the specific error messages "Must have one label per point" or 'use_pca does nothing for the python_LMNN implementation' ? I agree that the name `_process_inputs` here does not tell the difference from `_prepare_inupts` explicitly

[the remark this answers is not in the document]
bellet: I guess I was a bit confused, but maybe add a comment in the function to point out that some checks are done to avoid that parameters from the Shogun version are passed to the Python version

In the near future we should improve the performance of our own implementation so that we can remove this dependency to Shogun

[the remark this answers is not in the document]
wdevazelhes: I agree, will do
I'll raise an issue for removing the Shogun dependency

metric_learn/rca.py
```diff
@@ -175,6 +183,7 @@ def fit(self, X, y, random_state=np.random):
     y : (n) data labels
     random_state : a random.seed object to fix the random_state if needed.
     """
+    X, y = self._prepare_inputs(X, y)
```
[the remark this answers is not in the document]
wdevazelhes: Yes, indeed it returns a LinalgError with one sample

test/test_sklearn_compat.py
wdevazelhes: Indeed, will do

test/test_sklearn_compat.py
wdevazelhes: Actually I copied this function from scikit-learn and forgot to add in the comment "Adapted from scikit learn"... But I don't know why I copied it, I'll import it instead.

metric_learn/lsml.py
wdevazelhes: I didn't think of it,  I agree it would be cleaner. But is there a way to do this without the need to import the class/enum every time we want to fill the argument ? Like if we create TYPE_OF_INPUTS = Enum('TYPE_OF_INPUTS', 'classic tuples') in _util, anyone who would like to fill the argument "type_of_inputs=" of `check_input` would have to import `TYPE_OF_INPUTS` from `metric_learn._util` right ? This could add a bit of complexity for users used to put strings in arguments like I think is the case in many scikit-learn estimators (like method='svd' etc)? And if as it is in the docstring it will help them to put the right argument. But in this case maybe we should raise an error if the given string is not in {'tuples', 'classic'}, reminding the allowed strings ?

metric_learn/sdml.py
```diff
@@ -151,7 +161,7 @@ def fit(self, X, y, random_state=np.random):
     self : object
         Returns the instance.
     """
-    y = check_array(y, ensure_2d=False)
+    X, y = self._prepare_inputs(X, y)
```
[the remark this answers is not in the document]
wdevazelhes: Thanks, done: I added it for all supervised algorithms

wdevazelhes:
Thanks @terrytangyuan !
I'll only add a bit more tests for cross-validation and we should be good to merge

test/test_sklearn_compat.py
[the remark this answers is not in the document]
wdevazelhes: Sure, will do

test/test_sklearn_compat.py
```diff
@@ -68,5 +81,343 @@ def test_mmc(self):
...
+    # if not, we build a 3D array of pairs of samples
+    pairs, y = wrap_pairs(X, indices)
+  pairs, y = shuffle(pairs, y, random_state=RNG)
+  (pairs_train, pairs_test, y_train,
+   y_test) = train_test_split(pairs, y, random_state=RNG)
+  return (X, pairs, y, pairs_train, pairs_test,
+          y_train, y_test, preprocessor)
+
+
+def build_quadruplets(preprocessor):
```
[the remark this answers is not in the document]
wdevazelhes: Will do

wdevazelhes:
Hi, 
I think the PR should be good to merge now !
I addressed most of the comments, except those remaining ones:

 - https://github.com/metric-learn/metric-learn/pull/117#pullrequestreview-175516040 : preprocessor can be a class, which is indeed a callable but not a function (fn) anymore, plus it can also be an X (array-like). But maybe in tensorflow too they are not necessarily functions ? @terrytang how would you give an input which is just a numpy array in `tf.Estimators` ? with `input_fn=X` or creating an input function from X with something like `input_fn=make_an_input_function(X)` ? 

 - https://github.com/metric-learn/metric-learn/pull/117#discussion_r233889408: I assume I fixed it by adding a comment on why to return two times X but I'm not sure it's what you refered to @bellet ? (Note that  I recently refactored `build_pairs` so also maybe the comment is outdated)

 - https://github.com/metric-learn/metric-learn/pull/117#discussion_r233976645 : I went for a detailed error message for now, but maybe in the future we could refactor with enums ?

terrytangyuan:
@wdevazelhes Basically `input_fn` is a function that returns a dictionary of features and targets. It is exposed to users so that they can transform their batches of data to avoid feeding the entire dataset at once and to speed up (asynchronous) training with (distributed) data pipeline. It's a function that gets called that provides batches of data to each fit step so that more flexible data types can be fed in, e.g. iterators or streaming input. For input that is not a function, there are utility methods such as [numpy_input_fn](https://www.tensorflow.org/api_docs/python/tf/estimator/inputs/numpy_input_fn) and [pandas_input_fn](https://www.tensorflow.org/api_docs/python/tf/estimator/inputs/pandas_input_fn) that can be used as `input_fn`, e.g. 
```
age = np.arange(4) * 1.0
height = np.arange(32, 36)
x = {'age': age, 'height': height}
y = np.arange(-32, -28)

with tf.Session() as session:
  input_fn = numpy_input_fn(
      x, y, batch_size=2, shuffle=False, num_epochs=1)
```
Basically `numpy_input_fn` facilitates the process of creating a function/callable from numpy arrays. 

Regarding "I went for a detailed error message for now, but maybe in the future we could refactor with enums ?" - sure, sounds good.

bellet:
I would say that this is ready to merge, unless @perimosocordiae has comments?

metric_learn/_util.py
```diff
@@ -12,39 +15,314 @@ def vector_norm(X):
     return np.linalg.norm(X, axis=1)
 
 
-def check_tuples(tuples):
-  """Check that the input is a valid 3D array representing a dataset of tuples.
-
-  Equivalent of `check_array` in scikit-learn.
+def check_input(input_data, y=None, preprocessor=None,
+                type_of_inputs='classic', tuple_size=None, accept_sparse=False,
+                dtype="numeric", order=None,
```
perimosocordiae: Super nitpick: use single quotes for `dtype='numeric'` to keep consistency with the rest of the file.

metric_learn/_util.py
```diff
@@ -12,39 +15,314 @@ def vector_norm(X):
...
+  if args_for_sk_checks['ensure_min_features'] > 0:
+    n_features = input_data.shape[2]
+    if n_features < args_for_sk_checks['ensure_min_features']:
+      raise ValueError("Found array with {} feature(s) (shape={}) while"
+                       " a minimum of {} is required{}."
+                       .format(n_features, input_data.shape,
+                               args_for_sk_checks['ensure_min_features'],
+                               context))
+  #  normally we don't need to check_tuple_size too because tuple_size
+  # should'nt be able to be modified by any preprocessor
```
perimosocordiae: typo: shouldn't

metric_learn/_util.py
```diff
@@ -12,39 +15,314 @@ def vector_norm(X):
...
+  if input_data.ndim != 3:  # we have to ensure this because check_array
+    # above does not
```
perimosocordiae: This is a weird comment line-break. I'd prefer moving the whole comment either above or below the `if input_data.ndim != 3:` line.

metric_learn/_util.py
```diff
@@ -12,39 +15,314 @@ def vector_norm(X):
...
+                              i in range(tuples.shape[1])])
+  except Exception as e:
+    raise PreprocessorError(e)
   return tuples
+
+
+def preprocess_points(points, preprocessor):
+  """form points if there is a preprocessor else keep them as such (assumes
+  that check_points has already been called)"""
+  print("Preprocessing points...")
```
perimosocordiae: Remove this print call.

metric_learn/lsml.py
perimosocordiae: It seems that `_prepare_inputs` isn't user-facing. So long as users won't need to (commonly) pass an enum as an argument in the public API, I'm +1 for using an enum.

In any case, we should definitely raise an error if an unexpected value is passed.

wdevazelhes:
Thanks @perimosocordiae ! I addressed your comments.

Regarding `_prepare_inputs`, indeed it is not user-centric for now because the arguments are not copied from `check_input` (there is `*args` and `**kwargs` instead, and no docstring consequently), and it is private. `check_input` is for now the more user centric function. But in the meantime, `_prepare_input` would be the main method to use for developers/users who want to create their own custom estimator. So I'm wondering if we shouldn't replace `_prepare_input` by `prepare_input`, a public method where the arguments of `check_input` and their docstring would be copied  ? And therefore keep using a string as a flag too ('classic'/'tuples') ?

metric_learn/_util.py
wdevazelhes: I agree, thanks 
Done

metric_learn/_util.py

metric_learn/_util.py
wdevazelhes: I agree, done

metric_learn/_util.py
wdevazelhes: I agree it is not very useful (and not well formatted)

bellet:
This may be a stupid question but why do we need 2 functions?

wdevazelhes:
I guess maybe we don't necessarily need 2 functions indeed. Originally I thought of `check_input` as supposedly similar to `check_array` and `check_X_y`, and these are functions that can be used (and tested), without any `Estimator` instantiation. `_prepare_input` on the contrary is a `BaseMetricLearner` method that at the same time does `self.check_preprocessor`, and calls `check_input` filled with all the parameters that can be infered from knowing we are in some `BaseMetricLearner` (by the way for now we still have to specify 'type_of_input' but a simple refactoring distinguishing `SupervisedMetricLearner` and `WeaklySupervisedMetricLearner` would allow to also fill this argument)
But after all maybe we don't really need `check_input` on its own (I am not sure users would really want to use it as such, and for the tests we can always create some fake Metric Learners) ?

Having only the method `_prepare_input` would simplify the code indeed (not having an intermediate layers of arguments to pass, and knowing that necessarily are inside a `BaseMetricLearner` would also simplify the error messages (and their testing) since for now we can pass a string or None into `check_estimator(estimator=...)`, and probably other simplifications). 

I guess the questions is: what is best: the "feature" (maybe not so useful) to be able to use `check_input` on its own without any `Estimator`,  or the simplification of the code ?

In any case this is quite a concentrated part of the code so it could be refactored at any time

bellet:
OK, maybe it makes sense to have two functions then.

perimosocordiae:
In it goes!

perimosocordiae:
@wdevazelhes this merged into the new_api_design branch. Can you make a PR to merge that branch back into master?

perimosocordiae:
Ah, I didn't notice that gh-133 was off of the new_api_design branch as well. Typically I prefer to do all PRs against master to avoid lots of conflict resolution down the road.

Re: compatibility breaks, users building from master should be ready for breaking changes from time to time. We'll need to make sure the docs are sufficient before cutting the next release, though.

wdevazelhes:
> OK, maybe it makes sense to have two functions then. check_input may indeed be used by users while _prepare_input is for developers so should probably remain private?
--

Yes that's true, let's keep this for now and see later how it goes maybe

wdevazelhes:
>Ah, I didn't notice that gh-133 was off of the new_api_design branch as well. Typically I prefer to do all PRs against master to avoid lots of conflict resolution down the road.
Re: compatibility breaks, users building from master should be ready for breaking changes from time to time. We'll need to make sure the docs are sufficient before cutting the next release, though.

There's also the fact that the API is quite changed so users would really need some doc informations on how to use the preprocessor etc
But maybe we can try to merge the doc soon even if it is not totally complete with all the examples ? This way we would benefit from users building from source that would try to use the new API and file bugs etc

perimosocordiae:
Yes, I'd prefer to get master up to date sooner than later. Docs are important, and we definitely need good examples before we cut a release, but I want to make sure we're stable on master for a reasonable amount of time before releasing a new version.

PR #119, opened 2018-08-31
[MRG] API: remove num_labeled parameter

### LINKED ISSUE #118: Do we need the num_labeled parameter for *_Supervised classes?
#### Description

The *_Supervised classes allow to use a weakly supervised algorithm on labeled data by creating constraints from the labeled points.

I do not really understand why these classes have a parameter `num_labeled`, which is used to ignore some labeled points when creating constraints. To reduce the training complexity, it is enough to limit the number of constraints by using the `num_constraints` parameter. Using all available points to create the desired number of constraints can only benefit the algorithm as it sees more different points in the training constraints (i.e., it reduces variance).

I think `num_labeled` could be removed so that all available points are used to create the desired number of constraints.

wdevazelhes:
Fixes #118

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist/
 .coverage
 htmlcov/
 .cache/
+doc/auto_examples/*
--- a/README.rst
+++ b/README.rst
@@ -34,27 +34,8 @@ package installed).
 
 **Usage**
 
-For full usage examples, see the `sphinx documentation`_.
-
-Each metric is a subclass of ``BaseMetricLearner``, which provides
-default implementations for the methods ``metric``, ``transformer``, and
-``transform``. Subclasses must provide an implementation for either
-``metric`` or ``transformer``.
-
-For an instance of a metric learner named ``foo`` learning from a set of
-``d``-dimensional points, ``foo.metric()`` returns a ``d x d``
-matrix ``M`` such that the distance between vectors ``x`` and ``y`` is
-expressed ``sqrt((x-y).dot(M).dot(x-y))``.
-Using scipy's ``pdist`` function, this would look like
-``pdist(X, metric='mahalanobis', VI=foo.metric())``.
-
-In the same scenario, ``foo.transformer()`` returns a ``d x d``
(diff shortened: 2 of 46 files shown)


terrytangyuan:
We should probably record these kind of breaking changes in a change log?

bellet:
I think the _Supervised classes are introduced in this next 0.4.0 release though?

bellet:
My mistake it looks like they were introduced in 0.3.0 (albeit without doc)...
Then, maybe we can release 0.4.0 without this and save this PR for 0.5.0 (along with major API changes)?

wdevazelhes:
Yes I think so, I tried to resolve the conflicts but my branch was not updated so I messed it up :p I will try again...

terrytangyuan:
Seems like rebase was done incorrectly/unsuccessfully? At least from a quick glance it seems like `setup.py`'s changes are unexpected.

wdevazelhes:
Yes indeed, I think I did a bad revert, I'll revert try revert the revert d6bd0d4

wdevazelhes:
OK now there's just some tests for NCA that I don't understand why they are here...

wdevazelhes:
Oups, it appears I deleted it in the big merge #139  (cf. https://github.com/metric-learn/metric-learn/commit/23d07466961fa7a72aa8692bc42d6d569b80c5c9#diff-1d435b5cb210c40bb6a7261667f5c5acL138)
I'll open an issue to put it back, so that it should disappear of the diff of this PR as soon as it is merged

wdevazelhes:
Here is the PR that we should merge (as soon as tests passes): #143

wdevazelhes:
there just remains one error of version to update and it should be OK

wdevazelhes:
I guess this PR should be ready to merge now, as soon as tests pass

bellet:
still 1 failure ;-)

wdevazelhes:
Turns out I used the function `Constraint.random_subset` in #139 so I'll just need change those lines

wdevazelhes:
It should be good to merge now :p

PR #126, opened 2018-10-09
Fix python2.7 compatibility in setup.py

terrytangyuan:
Got the following when running tests under python2.7:
```
Traceback (most recent call last):
  File "setup.py", line 11, in <module>
    with open('README.rst', encoding='utf-8') as f:
TypeError: 'encoding' is an invalid keyword argument for this function
```
Reference: https://stackoverflow.com/questions/25049962/is-encoding-is-an-invalid-keyword-error-inevitable-in-python-2-x

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,13 +2,14 @@
 # -*- coding: utf-8 -*-
 from setuptools import setup
 import os
+import io
 
 version = {}
-with open(os.path.join('metric_learn', '_version.py')) as fp:
+with io.open(os.path.join('metric_learn', '_version.py')) as fp:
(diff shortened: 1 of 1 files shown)


terrytangyuan:
It seems like tests are flaky and not related to this PR.

wdevazelhes:
Yes, indeed, master is now failing: I think it must be because of scikit-learn's recent update to 0.20. I'll raise an issue for that.
Running your PR with 0.19.2 version, it passes, so I guess we can merge it and fix the tests in another PR

perimosocordiae:
Merged.

Issue #129, opened 2018-10-11
We should avoid only testing hard values


wdevazelhes:
Some of the tests rely on testing hard values, like the ones that failed in #127 due to scikit-learn's iris dataset update. They could probably fail again if for instance we used another initialisation or another optimizer for some algorithms, while the algorithm would still be valid. Therefore I think these tests could  still be useful as benchmarking tasks, to ensure we keep making a good score for some basic tasks, but we should still probably rely more on testing toy examples, testing properties of the solution rather than hard values, that would work no matter the initialization or the optimisation procedure.

perimosocordiae:
Yes, I've wanted to fix this issue with the test suite since gh-51. It'll be difficult to get the right test coverage without over-specifying results, but any progress on this is welcome.

Issue #128, opened 2018-10-11
Choose the minimal conditions to allow fit


wdevazelhes:
In #127, we saw that algorithms should return a special error message if the number of samples is too small to be able to run. Therefore, we should choose the minimal number of samples on which we want to allow fit. If theoretically the algorithm would work with only `n` sample, we should ensure that the code indeed works with `n` samples as input. Therefore we should ideally also test these minimal conditions (minimally allowed conditions and forbidden conditions).

For instance: `MMC_Supervised` should theoretically work on two samples (either if they are from the same class, meaning MMC will put them into a point, or if they have a different class, in this case it can put them to any distance superior to 1)

Or also for instance maybe `LSML_Supervised` would at least need 4 samples to be able to form one quadruplet

wdevazelhes:
> (This should also be investigated for weakly supervised algorithms)

Yes indeed

Issue #127, opened 2018-10-10
tests failing due to scikit-learn update


wdevazelhes:
Some tests are failing in master, due to scikit-learn update v0.20.0. There are two types of tests that are failing: 

- Tests that uses the iris dataset: this is probably due to the fact that the iris dataset has changed in scikit learn v.0.20: cf http://scikit-learn.org/stable/whats_new.html#sklearn-datasets
For now maybe we can do a quickfix to pass the tests but maybe in the future we should replace these tests that test hard numerical values ? 

- Tests using sklearn's check_estimator: they fail for the check `check_fit2d_1sample` which was improved in v0.20 by checking the actual error message that is returned: see diff between lines 594-595 of v0.19.0 and lines 843-845 of v0.20 [here](https://github.com/scikit-learn/scikit-learn/compare/0.19.0...0.20.0) in Files changed, `sklearn/utils/estimator_checks.py` (the git diff didn't really worked here so these lines are not aligned). They can be fixed by setting `ensure_min_samples=2` in the checks that are done at fit, but we should probably think more about what are the minimum number of samples needed to make the algorithms work (since scikit-learn only tests this error is thrown for 1 sample, but maybe the algorithm won't work for 2 samples either and return the bad error message as before, which is not ideal...)

bellet:
I think this has been solved, and a separate issue has been created for the question of setting minimal conditions to allow fit (#128)

PR #130, opened 2018-10-11
[MRG] Quick fix of failed tests due to new scikit-learn version (0.20.0)

wdevazelhes:
Fixes #126 #127 

This is a quick fix that fixes #126 and #127, by setting `ensure_min_samples=2` in checks of the inputs to pass scikit-learn's check_estimator test (see #127), and by updating the scores with the new values for the iris dataset.

But I opened two issues that tackle solving these problems in the long term (cf #127)
- issue #128 to decide what is the minimal number of samples that we want to allow the "fit" method on 
- issue #129 to deal with tests that test some hard values

### DIFF
--- a/metric_learn/itml.py
+++ b/metric_learn/itml.py
@@ -191,7 +191,7 @@ class ITML_Supervised(ITML):
     random_state : numpy.random.RandomState, optional
         If provided, controls random number generation.
     """
-    X, y = check_X_y(X, y)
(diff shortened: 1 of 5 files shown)


bellet:
Some iris tests are still failing ;-)

wdevazelhes:
Oh yes indeed, I used the wrong scikit-learn version for testing locally :p

Issue #123, opened 2018-09-13
No unified function to check the inputs


bellet:
Right now, almost each algorithm has its own function to prepare/preprocess/check its inputs (verify that there is no duplicate, etc).

Having a single such function which can deal with all inputs and is used by all algorithms would make the code more readable and modular.

terrytangyuan:
Maybe it's also a good chance for us to think about how to handle batches of input or streaming data in case of larger dataset. Probably need to categorize the algorithms to see if it's doable algorithmically. Then we can do something similar to the concept of "input_fn" in [TensorFlow Estimators](https://arxiv.org/pdf/1708.02637.pdf).

wdevazelhes:
@terrytangyuan yes, that's one of the next steps indeed. Nice, I didn't know about `input_fn`, it looks interesting I'll check it out

bellet:
Addressed by #139

Issue #134, opened 2018-11-28
Avoid storing any unnecessary variable after fit


wdevazelhes:
Some algorithms currently store a reference to the data at fit time (having a `self.X_` attribute for instance). According to scikit-learn guidelines (https://scikit-learn.org/stable/developers/contributing.html#fitting), we should avoid this:
>Note that the model is fitted using X and y, but the object holds no reference to X and y. There are, however, some exceptions to this, as in the case of precomputed kernels where this data must be stored for use by the predict method.

If we really need all the data we could maybe store a copy of X but I don't think it is the case in metric-learn's algorithms

wdevazelhes:
Currently in the PR preprocessor #117 it was present in LFDA, LMNN, and MLKR, but I changed it in commit https://github.com/wdevazelhes/metric-learn/commit/f2d0cd7f989cc1c41e34cb41b69b2f6982fbe413 (and the following commit) so when the PR is merged this issue should be closed

bellet:
So once the API branch is merged we can close this as well?

wdevazelhes:
Yes !

wdevazelhes:
I just updated the title, see comment https://github.com/metric-learn/metric-learn/pull/139#issuecomment-450863654

wdevazelhes:
Here is a list of all variables that are stored at fit time. We should decide whether we keep them because they are useful for the user after the estimator is fitted (like `components_` in a PCA for instance), (and in this case we should add them in the docstring under the `Attributes` title), or make them temporary objects. For each, in comment I'll put a guess for the previous answer, but feel free to fill it if you can. 

List of attributes (except transformer)

- ITML: 
  - bounds_: ?
  - A_: should be discarded since we have transformer_

- LMNN: 
  - labels_inds: probably should be discarded but to check
  - n_iter_: should be kept and added to "Attributes"

- LSML:
  - vab_: ? 
  - vcd_: ?
  - w_: ? they are weights, todo: check in scikit-learn if they store it
  - prior_inv_: discard ? 
  - M_: discrad since there is already transformer_
  - n_iter_: keep and add to Attributes

- MLKR:
  - n_iter_: keep and add to Attributes

- MMC:
  - A_:  remove (already transformer_)
  - n_iter_: keep and advertise

- NCA:
 - n_iter_: idem

- SDML:
  - M_: remove (already transformer_)

bellet:
ITML: bounds_ can be kept

[...]

PR #139, opened 2018-12-21
[MRG] New api design

wdevazelhes:
This PR merges `new_api_design` into `master`, adding the new API to metric learn as well as a minimal documentation 

Achieves #91 
Also addresses #123  

TODO: 
- [x] Change `README` ? 
<strike>- [ ] fix pep8 erros comparing (in the diff with master)</strike>
<strike>- [ ]update the notebook `metric_plotting.ipynb` with the new API</strike>
<strike>- [ ] take a second look at methods that were modified to prevent using `self.X_` (and therefore have a new argument `X`) (ex: `RCA._check_dimension`) some of them could be static methods. It could be good to explicitly make them so (maybe just file an issue for that and solve after not to mix with this PR)</strike>


test/test_transformer_metric_conversion.py
```diff
@@ -19,60 +19,60 @@ def setUpClass(self):
...
     seed = np.random.RandomState(1234)
     sdml = SDML_Supervised(num_constraints=1500)
     sdml.fit(self.X, self.y, random_state=seed)
-    L = sdml.transformer()
+    L = sdml.transformer_
     assert_array_almost_equal(L.T.dot(L), sdml.metric())
 
   def test_nca(self):
     n = self.X.shape[0]
     nca = NCA(max_iter=(100000//n), learning_rate=0.01)
```
bellet: again here

doc/supervised.rst
```diff
@@ -0,0 +1,209 @@
...
+
+    import numpy as np
+    from metric_learn import NCA
+    from sklearn.datasets import load_iris
+
+    iris_data = load_iris()
+    X = iris_data['data']
+    Y = iris_data['target']
+
+    nca = NCA(max_iter=1000, learning_rate=0.01)
```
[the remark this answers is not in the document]
wdevazelhes: Indeed, thanks !
Done

test/test_transformer_metric_conversion.py
wdevazelhes: Yep, done

wdevazelhes:
I'll also update the `README.rst` here (for instance removing the paragraph about `transformer()`)

bellet:
Should we update the README, or do these instructions refer to the last release (install feom PyPi)?

wdevazelhes:
> Should we update the README, or do these instructions refer to the last release (install from PyPi)?

Ah yes that's right, they seem to refer to the last release
Should we keep instructions on using the package (algorithms etc) in the `README` though ? Because this seems to be already documented in the documentation (by the way in the doc we say the distance is ``(x−y)M(x−y)`` whereas in the README we say it's ``sqrt((x−y)M(x−y))``. I guess the real distance is with sqrt so we should modify in the docs right ? )
Because if we make changes in algorithms we'll have to always update this page, but do it just after the release, it seems a bit weird

wdevazelhes:
I just added a TODO in the description of the PR

bellet:
I agree that we could probably remove the detailed usage instructions on the README and only keep the link to the doc
And yes, the proper distance is with the square root and this is also what is returned by `score_pairs` so we should change the doc accordingly

perimosocordiae:
I'd prefer to keep this PR limited as much as possible to the merge, and then fix the remaining minor issues (pep8 stuff, jupyter notebook, static methods, etc) separately. The idea is that those will be small and easy to review, whereas this PR is pretty large already.

bellet:
Yes that's true, reviewing additional things here will be more difficult
@wdevazelhes could you perhaps simply update the README and move the rest to new issues that you can deal with separately? So we can merge this quickly and start making progress on small things and on improving the doc in separate PRs

wdevazelhes:
I agree, I just pushed the new README and updated the introduction with the square root

wdevazelhes:
> About `self.X_` / `self.X`: if you don't do it here, make sure you remove #134 from the list of issues addressed by this PR

I replaced `self.X_` by `self.X` here, but there still remains some variables that maybe do not need to be stored after fit, like for instance `self.labels_` in ``LMNN``, or `self.vab_` in `LSML`, so I'll remove it from listed issues, and change the title to a more general title (not just "avoid storing X" but "avoid storing unnecessary variables")

bellet:
Thanks, I will merge this

PR #143, opened 2019-01-02
[MRG] FIX put back nca's tests

wdevazelhes:
This PR puts back tests on NCA that were put in #99, but that I deleted unintentionnaly in PR #139


wdevazelhes:
It should be good to merge now. Basically it is just the same tests that were introduced in #99, but here I just also remove the learning rate deprecation test, and I change the tests to use `transformer_` instead of `A_`

bellet:
Thanks, let's merge

PR #145, opened 2019-01-03
[MRG] Documentation: introduction to metric learning

bellet:
This PR improves the documentation by adding a proper section introducing metric learning (problem setting, mahalanobis distance, use cases and general references). A small description of the package is also added to the index page for clarity.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -2,11 +2,12 @@ metric-learn: Metric Learning in Python
 =======================================
 |License| |PyPI version|
 
-This package contains efficient Python implementations of several popular
-supervised and weakly-supervised metric learning algorithms. The API of
-metric-learn is compatible with scikit-learn, allowing the use of all the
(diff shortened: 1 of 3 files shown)


bellet:
I still need to write a few lines about use-cases.

Note that I also commented out the part listing the class methods: I think it would be better to move this to the supervised and weakly supervised sections, within a proper 'API' subsection

bellet:
This should be ready to merge - feedback is welcome.

wdevazelhes:
> Note that I also commented out the part listing the class methods: I think it would be better to move this to the supervised and weakly supervised sections, within a proper 'API' subsection

Yes I agree, the current 'Package overview' subsection is currently not organized. We could add these comment lines, plus also organize the sections a bit (like separate weakly supervised algos from supervised etc). I'll raise an issue for that

perimosocordiae:
Merged, thanks! Moving the commented-out section should be handled by gh-149.

PR #151, opened 2019-01-07
[MRG] Move transformer_from_metric to util

### LINKED ISSUE #146: transformer_from_metric should be a helper function
#### Description
The function `transformer_from_metric(metric)` which decomposes a PSD matrix into a transformer is currently a method of the class `MahalanobisMixin`. I think the purpose of this function is to do the conversion at the end of the fit algorithm when the latter learns the metric as a PSD matrix instead of a linear transformation. Hence I think it should simply be a helper function.

bellet:
Fixes #146

### DIFF
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -345,7 +345,7 @@ def transformer_from_metric(metric):
   if np.allclose(metric, np.diag(np.diag(metric))):
     return np.sqrt(metric)
   elif not np.isclose(np.linalg.det(metric), 0):
-    return np.linlalg.cholesky(metric).T
+    return np.linalg.cholesky(metric).T
   else:
     w, V = np.linalg.eigh(metric)
     return V.T * np.sqrt(np.maximum(0, w[:, None]))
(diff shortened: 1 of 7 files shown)


perimosocordiae:
LGTM, merged.

perimosocordiae:
Side note: we may want to find a more performant way to check if a matrix is diagonal, other than `allclose(m, diag(diag(m)))`. I wonder if scikit-learn has something along those lines?

perimosocordiae:
Then again, the metric is a `d` by `d` matrix, so it's not a big deal in most cases.

wdevazelhes:
Yes, according to this [answer](https://stackoverflow.com/a/43884735/5635843) and this [answer](https://stackoverflow.com/a/23567941/5635843), maybe a good way in numpy would be to remove the diagonal and check if the matrix is zero, like: 
`np.fill_diag(m, 0); not m.any()` (but apparently `any` does not short circuit when finding a not zero element as noted in the Stackoverflow answer (but then this [answer](https://stackoverflow.com/a/53474543/5635843) uses numba to do a short-circuit function)
Plus I don't know if there is a way to test if numbers are **close** to 0 in numpy 
And yes maybe scikit-learn has some helper function to do that

Issue #131, opened 2018-11-13
Use a threshold for prediction for pairs


wdevazelhes:
A pairs predictor should be able to predict a binary result when given a pair, like a classifier, when we would call the `predict` function. This could be done by using a threshold and comparing the score of similarity for the pair to this threshold.

TODO: 
Also: 
- [ ] replace `score_pairs` by a better name like `compute_distances` (see comment https://github.com/metric-learn/metric-learn/pull/152#issuecomment-453554793)

bellet:
This threshold could be tuned (e.g. on the training set) so as to achieve a given level of precision

wdevazelhes:
1/ There's this recent PR in scikit-learn that is related: https://github.com/scikit-learn/scikit-learn/pull/10117

They use a meta-estimator taking as an argument the estimator we want to threshold. It seems to be pretty close to what we want to do, allowing to specify a precision level etc, allowing not to refit the model if needed...

However it does not have an option for setting a threshold for just optimizing the accuracy. They indeed wouldn't need it if all binary classifiers in scikit-learn optimized for a cost function that gave the higher accuracy when setting the threshold to 0.5 (which would mean that accuracy is the high-level metric optimized by default). I think this is the case for probabilistic classifiers, but I don't know if it is true for others...

I'll try to investigate that but maybe you already know the answer ? 

Because if we find a case in scikit-learn where we want to choose a threshold to optimize for accuracy then we could say it in the PR to add it there. But otherwise I think we would need to implement it in metric-learn.

What is more, maybe it would be good to have this (the accuracy maximizing thresholding) by default in pairs metric learners so that they directly have a predict function and we don't need a meta-estimator to have it ? But then for more sophisticated selections (like with precision, etc) they would use the scikit-learn metaestimator which should be compatible ? 



2/ Also related, there's this class (from the module sklearn.calibration) that allows to calibrate the predictions. It provides a  good `predict_proba` to estimators that don't have it or have a bad one (I don't think we need to put it in the code but maybe just say in the docs we can use this):

https://scikit-learn.org/stable/modules/generated/sklearn.calibration.CalibratedClassifierCV.html#sklearn.calibration.CalibratedClassifierCV

The only problem is I could only use it using a preprocessor. Indeed at fit time they have a check_array that doesn't accept 3D arrays.. I guess I could file an issue in scikit-learn about that, because it's a bit weird since GridSearchCV for instance doesn't do that and is still a metaestimator.

[...]

PR #152, opened 2019-01-09
[MRG] Refactor the metric() method 

### LINKED ISSUE #147: Output of metric() method
#### Description
Currently the `metric()` method of `MahalanobisMixin` returns the PSD matrix `self.transformer_.T.dot(self.transformer_)`.

I am not sure when this is really useful in practice?Shouldn't we rather return the metric function itself as a callable, or better a [`DistanceMetric`](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.DistanceMetric.html) that can be plugged easily into some sklearn estimators? This seems more useful.

Defined this way, `metric()` could be an abstract method of `BaseMetricLearner` (since every metric learner should be based on some kind of metric).

wdevazelhes:
Fixes #147 

TODO: 
- [x] Add some tests
- [x] Add references to the right parts of documentation (like `Mahalanobis Distances`) in the docstrings (if possible)
- [x] Emphasize a bit more the difference and links between this and `score_pairs` in the docstring
- [x] Be careful that it should work on 1D arrays  
- [x] Be careful that it should not return a float if given 2D arrays
- [x] Remove useless `np.atleast2d` (those in `transformer_from_metric` and those just before returning the `transformer_`)


wdevazelhes:
Maybe this PR would be the occasion for changing `score_pairs` into some more expressive name ? 
What do you think ? `compute_distances` maybe ?

metric_learn/base_metric.py
```diff
@@ -178,8 +214,57 @@ def transform(self, X):
...
+        The first point involved in the distances computation.
+      v : array-like, shape=(n_features,)
+        The second point involved in the distances computation.
+      Returns
+      -------
+      distance: float
+        The distance between u and v according to the new metric.
+      """
+      u = _validate_vector(u)
+      v = _validate_vector(v)
```
wdevazelhes: Here I use scipy's _validate_vector function (used in functions like `scipy.spatial.distance.euclidean`), to more easily mimic scipy's distances behaviour (regarding 1D arrays etc). But I see the underscore at the beginning, which would mean that this function is private. Does it mean I am not supposed to use it ?

wdevazelhes:
I guess if you agree with my comments this PR is ready to merge

metric_learn/base_metric.py
```diff
@@ -177,8 +214,57 @@ def transform(self, X):
                              accept_sparse=True)
     return X_checked.dot(self.transformer_.T)
 
-  def metric(self):
```
perimosocordiae: Maybe we should keep this for now but mark it as deprecated, and point to the new `get_mahalanobis_matrix()` method.

bellet:
> Maybe this PR would be the occasion for changing `score_pairs` into some more expressive name ?
> What do you think ? `compute_distances` maybe ?

I think #131 is probably more the place

test/test_mahalanobis_mixin.py
```diff
@@ -167,3 +167,47 @@ def test_embed_is_linear(estimator, build_dataset):
...
+  rng = np.random.RandomState(42)
+  input_data, labels, _, X = build_dataset()
+  model = clone(estimator)
+  set_random_state(model)
+  model.fit(input_data, labels)
+  metric = model.get_metric()
+  n_features = X.shape[1]
+  a, b = (rng.randn(n_features), rng.randn(n_features))
+  euc_dist = euclidean(model.transform(a[None]), model.transform(b[None]))
+  assert (euc_dist - metric(a, b)) / euc_dist < 1e-15
```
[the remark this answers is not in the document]
bellet: Or you could use `assert_almost_equal` with a custom tolerance. `1e-15` may be a bit brittle?

metric_learn/base_metric.py
```diff
@@ -85,6 +94,24 @@ def _prepare_inputs(self, X, y=None, type_of_inputs='classic',
                        tuple_size=getattr(self, '_tuple_size', None),
                        **kwargs)
 
+  @abstractmethod
+  def get_metric(self):
+    """Returns a function that returns the learned metric between two points.
```
bellet: Returns a function that takes as input two 1D arrays and outputs the learned metric score on these two points?

metric_learn/base_metric.py
```diff
@@ -85,6 +94,24 @@ def _prepare_inputs(self, X, y=None, type_of_inputs='classic',
...
+    This function will be independent from the metric learner that learned it
+    (it will not be modified if the initial metric learner is modified).
```
bellet: maybe add that the returned function can be directly plugged into the `metric` argument of sklearn's estimators

metric_learn/base_metric.py
```diff
@@ -85,6 +94,24 @@ def _prepare_inputs(self, X, y=None, type_of_inputs='classic',
...
+
+    Returns
+    -------
+    metric_fun : function
+      The function described above.
+
+    See Also
+    --------
+    score_pairs : a method that returns the metric between several pairs of
```
bellet: the metric score

metric_learn/base_metric.py
```diff
@@ -34,6 +36,13 @@ def score_pairs(self, pairs):
     -------
     scores: `numpy.ndarray` of shape=(n_pairs,)
       The score of every pair.
+
+    See Also
+    --------
+    get_metric : a method that returns a function to compute the metric between
+      two points. The difference is that it works on two 1D arrays and cannot
```
bellet: The difference with score_pairs

metric_learn/base_metric.py
```diff
@@ -85,6 +94,24 @@ def _prepare_inputs(self, X, y=None, type_of_inputs='classic',
...
+
+    Returns
+    -------
+    metric_fun : function
+      The function described above.
+
+    See Also
+    --------
+    score_pairs : a method that returns the metric between several pairs of
+      points. But this is a method of the metric learner and therefore can
```
bellet: -But +Unlike get_metric

metric_learn/base_metric.py
```diff
@@ -146,6 +173,16 @@ def score_pairs(self, pairs):
     -------
     scores: `numpy.ndarray` of shape=(n_pairs,)
       The learned Mahalanobis distance for every pair.
+
+    See Also
+    --------
+    get_metric : a method that returns a function to compute the metric between
```
bellet: same updates as above

metric_learn/base_metric.py
```diff
@@ -177,8 +214,57 @@ def transform(self, X):
...
+    (it will not be modified if the initial metric learner is modified).
+
+    Returns
+    -------
+    metric_fun : function
+      The function described above.
+
+    See Also
+    --------
+    score_pairs : a method that returns the metric between several pairs of
```

metric_learn/base_metric.py
```diff
@@ -177,8 +214,57 @@ def transform(self, X):
...
+    transformer_T = self.transformer_.T.copy()
+
+    def metric_fun(u, v):
+      """This function computes the metric between u and v, according to the
+      previously learned metric.
+
+      Parameters
+      ----------
+      u : array-like, shape=(n_features,)
+        The first point involved in the distances computation.
```
bellet: distance

metric_learn/base_metric.py
```diff
@@ -177,8 +214,57 @@ def transform(self, X):
...
+      v : array-like, shape=(n_features,)
+        The second point involved in the distances computation.
```
bellet: distance

test/test_mahalanobis_mixin.py
```diff
@@ -167,3 +167,47 @@ def test_embed_is_linear(estimator, build_dataset):
...
+  set_random_state(model)
+  model.fit(input_data, labels)
+  metric = model.get_metric()
+
+  n_features = X.shape[1]
+  a, b, c = (rng.randn(n_features) for _ in range(3))
+  assert metric(a, b) >= 0  # positivity
+  assert metric(a, b) == metric(b, a)  # symmetry
+  # one side of identity indiscernables: x == y => d(x, y) == 0. The other
+  # side is not always true for Mahalanobis distances.
```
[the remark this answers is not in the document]
bellet: Mahalanobis distances are only a "pseudo" metric because they do not satisfy the "identity of indiscernables": `metric(x, y)` can be `0` even if `x != y`

test/test_mahalanobis_mixin.py
```diff
@@ -167,3 +167,47 @@ def test_embed_is_linear(estimator, build_dataset):
...
+  """
+  rng = np.random.RandomState(42)
+  input_data, labels, _, X = build_dataset()
+  model = clone(estimator)
+  set_random_state(model)
+  model.fit(input_data, labels)
+  metric = model.get_metric()
+
+  n_features = X.shape[1]
+  a, b, c = (rng.randn(n_features) for _ in range(3))
```
bellet: perhaps it would be more convincing to test that these are true on a set of random triplets (a, b, c), instead of a single one

test/test_mahalanobis_mixin.py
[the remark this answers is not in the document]
wdevazelhes: Yes I agree it's better to use a built-in function. I just saw that `assert_almost_equal` gives an absolute error whereas `numpy.testing.assert_allclose` gives a relative error, I guess it's even better to put a relative one in this case ? For the 1e-15, I agree, it's just that I was thinking this way, if we change our implementation and the precision worsens for one reason or another, we could notice it with this test (and it could help solve bugs elsewhere for instance)

metric_learn/base_metric.py
wdevazelhes: Understood, I'll replace it by something else

wdevazelhes:
> It would be better to avoid using private function since it may get removed without any prior notices. Perhaps look into `_validate_vector`'s implementation to see if you can implement it using scipy's exposed APIs or utility functions.

Yes, indeed, I'll look at this

metric_learn/base_metric.py
wdevazelhes: Yes, I agree, I forgot about that

metric_learn/base_metric.py
wdevazelhes: Agreed it's clearer indeed

metric_learn/base_metric.py
wdevazelhes: Agreed, will do

metric_learn/base_metric.py
wdevazelhes: Agreed, will do

metric_learn/base_metric.py
wdevazelhes: Agreed, will do

metric_learn/base_metric.py
wdevazelhes: Agreed, will do

metric_learn/base_metric.py
wdevazelhes: yes, will do

metric_learn/base_metric.py
wdevazelhes: Thanks, will do

metric_learn/base_metric.py
wdevazelhes: will do

test/test_mahalanobis_mixin.py
wdevazelhes: Yes I agree, will do

test/test_mahalanobis_mixin.py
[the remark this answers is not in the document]
wdevazelhes: Yes, `x == y => d(x, y) == 0` but `d(x, y) == 0 !=> x == y`

wdevazelhes:
I adressed all comments. I just need to remove the metric_plotting that I included by mistake and we should be good to merge

wdevazelhes:
It should be good to merge now, as soon as tests pass

wdevazelhes:
I just realized I could have used the help of `scipy.spatial.distance.mahalanobis`. This would do the trick of checking the array without needing to duplicate scipy's code and tests !
I'll update the PR

test/test_mahalanobis_mixin.py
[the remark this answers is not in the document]
bellet: yes relative error is probably better

wdevazelhes:
Here is a little snippet comparing performances of different functions we could have, as suggested by @bellet : 

```python
In [9]: import numpy as np 
   ...: from metric_learn._util import validate_vector 
   ...: from scipy.spatial.distance import euclidean, mahalanobis 
   ...: from functools import partial 
   ...:  
   ...: a, b = np.random.randn(100), np.random.randn(100) 
   ...: M = np.random.randn(100, 100) 
   ...: L = np.random.randn(100, 3) 
   ...:  
   ...: def metric_fun_1(u, v): 
   ...:     # current implem 
   ...:     u = validate_vector(u) 
   ...:     v = validate_vector(v) 
   ...:     return euclidean(u.dot(L), v.dot(L)) 
   ...:  
   ...: def metric_fun_2(u, v): 
   ...:     # implem with replicated code from distance.mahalanobis 
   ...:     u = validate_vector(u) 
   ...:     v = validate_vector(v) 
   ...:     VI = np.atleast_2d(M) 
   ...:     delta = u - v 
   ...:     m = np.dot(np.dot(delta, VI), delta) 
   ...:     return np.sqrt(m) 
   ...:  
   ...: def metric_fun_3(u, v): 
   ...:     # implem calling mahalanobis distance 
   ...:     return mahalanobis(u, v, VI=M) 
   ...:  
   ...: metric_fun_4 = lambda u, v: mahalanobis(u, v, VI=M)  
   ...:  
   ...: metric_fun_5 = partial(mahalanobis, VI=M) 
   ...:  
   ...: def metric_fun_6(u, v): 
   ...:     # current implem 
   ...:     u = validate_vector(u) 
   ...:     v = validate_vector(v) 
   ...:     transformed_diff = (u - v).dot(L) 
   ...:     return transformed_diff.dot(transformed_diff.T) 
   ...:  
   ...: %timeit metric_fun_1(a, b) 
   ...: %timeit metric_fun_2(a, b) 
   ...: %timeit metric_fun_3(a, b) 
   ...: %timeit metric_fun_4(a, b) 
   ...: %timeit metric_fun_5(a, b) 
   ...: %timeit metric_fun_6(a, b)                                                                                                                                                                                  
18.3 µs ± 91.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
/home/will/anaconda3/envs/standard/bin/ipython:23: RuntimeWarning: invalid value encountered in sqrt
15.9 µs ± 1.68 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)
/home/will/anaconda3/envs/standard/lib/python3.7/site-packages/scipy/spatial/distance.py:1066: RuntimeWarning: invalid value encountered in sqrt
  return np.sqrt(m)
15.6 µs ± 548 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
16.5 µs ± 1.02 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)
15.8 µs ± 427 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
5.73 µs ± 231 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Note that I chose an L with a very small dimension to see if it could improve performance. What appears above is that:
- metric_func_1 is not so performant, even with the low rank of L. We could try see why (maybe because we transform the two vectors separately, and then there are a lot of checks done in euclidean ?)
- using Mahalanobis is quite performant compared to metric_func_1
- different methods for building the function (function, lambda, partial etc), at first sight does not seem to change the performance a lot
- the last computation is the most performant in this case: this can be explained by the following: 
  - `(u-v) L L^T (u - v)^T` has `k * nf + k` operations (with L of shape (nf, k))
  - `(u-v) M (u-v)^T` has `nf**2 + nf` operations (nf: n_features)
  - this would mean that since k is always less than nf, this last function would always be better
  - plus maybe also the fact that here we don't check atleast2D(L)

At first sight it seems that we should therefore go for the last function in the end ? What do you think ?

perimosocordiae:
Sure, I'm happy to keep our current implementation. Maybe with a comment saying it's equivalent to `mahalanobis(u, v, VI=M) `.

wdevazelhes:
Here is another benchmark. Note that in `metric_fun_2` I remove the `np.atleast2D` of L (which could be done outside the function returned so it is better like this). I also put smaller matrices to see more the effects of the different ways to build functions and I give L and M the same size (to check what would happen). What this benchmark shows is that indeed it is a bit faster not to do `np.atleast2D`, and that in the end the different ways to build functions still do not matter a lot. But there's something I don't understand is why `metric_fun_6` is faster than `metric_fun_2`: they should have the same number of operations... Or did I miss something ? 

```python
In [13]: import numpy as np 
    ...: from metric_learn._util import validate_vector 
    ...: from scipy.spatial.distance import euclidean, mahalanobis 
    ...: from functools import partial 
    ...:  
    ...: a, b = np.random.randn(4), np.random.randn(4) 
    ...: M = np.random.randn(4, 4) 
    ...: L = np.random.randn(4, 4) 
    ...:  
    ...: def metric_fun_1(u, v): 
    ...:     # current implem 
    ...:     u = validate_vector(u) 
    ...:     v = validate_vector(v) 
    ...:     return euclidean(u.dot(L), v.dot(L)) 
    ...:  
    ...: def metric_fun_2(u, v): 
    ...:     # implem with replicated code from distance.mahalanobis 
    ...:     u = validate_vector(u) 
    ...:     v = validate_vector(v) 
    ...:     delta = u - v 
    ...:     m = np.dot(np.dot(delta, M), delta) 
    ...:     return np.sqrt(m) 
    ...:  
    ...: def metric_fun_3(u, v): 
    ...:     # implem calling mahalanobis distance 
    ...:     return mahalanobis(u, v, VI=M) 
    ...:  
    ...: metric_fun_4 = lambda u, v: mahalanobis(u, v, VI=M)  
    ...:  
    ...: metric_fun_5 = partial(mahalanobis, VI=M) 
    ...:  
    ...: def metric_fun_6(u, v): 
    ...:     # current implem 
    ...:     u = validate_vector(u) 
    ...:     v = validate_vector(v) 
    ...:     transformed_diff = (u - v).dot(L) 
    ...:     return transformed_diff.dot(transformed_diff.T) 
    ...:  
    ...: %timeit metric_fun_1(a, b) 
    ...: %timeit metric_fun_2(a, b) 
    ...: %timeit metric_fun_3(a, b) 
    ...: %timeit metric_fun_4(a, b) 
    ...: %timeit metric_fun_5(a, b) 
    ...: %timeit metric_fun_6(a, b)                                                                                                                                                                                 
16.7 µs ± 74.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
6.44 µs ± 31.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
7.38 µs ± 21.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
7.4 µs ± 14.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
7.25 µs ± 16.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
4.89 µs ± 2.71 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

wdevazelhes:
> Sure, I'm happy to keep our current implementation. Maybe with a comment saying it's equivalent to `mahalanobis(u, v, VI=M) `.

What do you mean by our current implementation ? `metric_func_1` ? Tests seem to show that we should do `metric_func_6`
Sure, I agree, we should put this in comment, and also maybe use this function in the tests, it'll be clearer

wdevazelhes:
This made me realize that taking the square root can be quite expensive, so I think it could be cool to have an option like "squared=True/False" ? Maybe in a next PR ?

bellet:
We should go for `metric_fun_6`: it is faster than `metric_fun_2` when `L` is rectangular, and the same when `L` is square (same size as `M`). Note: you have put a comment "current implem" for `metric_fun_6`, hence @perimosocordiae 's comment I think ;-)

I agree that having an option for squaring is useful - in many cases one does not care about the square root and it makes distance computations faster. I think you should add it now

bellet:
> Sure, I agree, we should put this in comment, and also maybe use this function in the tests, it'll all be clearer
What do you mean use it in the tests? We should definitely test our actual method
Putting in the comments that it is equivalent to the call to Mahalanobis but sometimes faster is enough

metric_learn/base_metric.py
```diff
@@ -177,9 +240,87 @@ def transform(self, X):
...
+      v = validate_vector(v)
+      return euclidean(u.dot(transformer_T), v.dot(transformer_T))
+    return metric_fun
+
   def metric(self):
+    # TODO: remove this method in version 0.6.0
+    warnings.warn(("`metric` is deprecated since version 0.5.0 and will be "
+                   "removed in 0.6.0. Use `get_mahalanobis_matrix` instead."),
+                  DeprecationWarning)
     return self.transformer_.T.dot(self.transformer_)
```
perimosocordiae: Might as well call `self.get_mahalanobis_matrix()` here.

metric_learn/base_metric.py
```diff
@@ -177,9 +240,87 @@ def transform(self, X):
...
 
+  def get_mahalanobis_matrix(self):
+    """Returns a copy of the Mahalanobis matrix learned by the metric learner.
+
+    Returns
+    -------
+    M : `numpy.ndarray`, shape=(n_components, n_features)
+      The copy of the learned Mahalanobis matrix.
+    """
+    return self.transformer_.T.dot(self.transformer_).copy()
```
perimosocordiae: There's no need for a `.copy()` here, as the result of the dot product is a new array.

metric_learn/base_metric.py
```diff
@@ -177,9 +240,87 @@ def transform(self, X):
                              accept_sparse=True)
     return X_checked.dot(self.transformer_.T)
 
+  def get_metric(self):
+    """Returns a function that takes as input two 1D arrays and outputs the
```
perimosocordiae: It would be nice to only write this docstring once. Unfortunately Python doesn't do inherited docstrings, so we have to assign it manually:

```
def get_metric(self):
  transformer_T = ...  # code here as usual

get_metric.__doc__ = BaseMetricLearner.get_metric.__doc__
```

wdevazelhes:
>  Note: you have put a comment "current implem" for `metric_fun_6`, hence @perimosocordiae 's comment I think ;-)

Oh yes indeed, thanks, I corrected the snippet

wdevazelhes:
> What do you mean use it in the tests? We should definitely test our actual method

I agree, I just meant we could replace this test: 

```python
@pytest.mark.parametrize('estimator, build_dataset', metric_learners,
                         ids=ids_metric_learners)
def test_get_metric_equivalent_to_transform_and_euclidean(estimator,
                                                          build_dataset):
  """Tests that the get_metric method of mahalanobis metric learners is the
  euclidean distance in the transformed space
  """
  rng = np.random.RandomState(42)
  input_data, labels, _, X = build_dataset()
  model = clone(estimator)
  set_random_state(model)
  model.fit(input_data, labels)
  metric = model.get_metric()
  n_features = X.shape[1]
  a, b = (rng.randn(n_features), rng.randn(n_features))
  euc_dist = euclidean(model.transform(a[None]), model.transform(b[None]))
  assert_allclose(metric(a, b), euc_dist, rtol=1e-15)
```

By something like this:

```python
@pytest.mark.parametrize('estimator, build_dataset', metric_learners,
                         ids=ids_metric_learners)
def test_get_metric_equivalent_to_explicit_mahalanobis(estimator,
                                                       build_dataset):
  """Tests that using the get_metric method of mahalanobis metric learners is 
  equivalent to explicitely calling scipy's mahalanobis metric
  """
  rng = np.random.RandomState(42)
  input_data, labels, _, X = build_dataset()
  model = clone(estimator)
  set_random_state(model)
  model.fit(input_data, labels)
  metric = model.get_metric()
  n_features = X.shape[1]
  a, b = (rng.randn(n_features), rng.randn(n_features))
  expected_dist = mahalanobis(a[None], b[None],
                              VI=model.get_mahalanobis_matrix())
  assert_allclose(metric(a, b), expected_dist, rtol=1e-15)
```

The test would even show more explicitely what the get_metric function does 
But it's not very important

wdevazelhes:
> I agree that having an option for squaring is useful - in many cases one does not care about the square root and it makes distance computations faster. I think you should add it now

I agree, let's do it while we're at it :)

metric_learn/base_metric.py
wdevazelhes: That's right, thanks

metric_learn/base_metric.py
wdevazelhes: That's right thanks, I don't know why I left the copy there...

metric_learn/base_metric.py
wdevazelhes: Cool thanks ! I always wondered how to avoid copying docstrings

wdevazelhes:
I just added the squared option, and addressed all the comments, so I think as soon as it is green we should be good to go

wdevazelhes:
I just added a commit that should fix the previous error, which does the following: 

- ensure that the transformer_ fitted is always 2D:
  - in the result returned from transformer_from_metric
  - in the code of metric learners, for metric learners that don't call transformer_from_metric
- for metric learners that cannot work on 1 feature, ensure it when checking the input
- add a test to check this behaviour

wdevazelhes:
> I am not sure I follow. I think a transformer should always be 2D. Even assuming the data is 1D, I would want the shape to be `(1, 1)` for consistency

Yes, that's the case: even without forcing with `np.atleast2d` as I did in my previous commit, they in fact passed the test checking that with scalar inputs, the data should be of shape (1, 1) (I checked it by going back to an old commit and copy-pasting this test)

wdevazelhes:
> I also do not see why algorithms wouldn't work on data with a single feature? (it is useless, sure, but should probably work)

I agree, `SDML` and `RCA` returned an error because they called respectively `pinvh` and `_inv_sqrtm` on a non 2d array. I'll fix them by calling `np.atleast2d` on the arguments of these methods, that's better than forbidding that the input be scalar...

bellet:
> I agree, `SDML` and `RCA` returned an error because they called respectively `pinvh` and `_inv_sqrtm` on a non 2d array. I'll fix them by calling `np.atleast2d` on the arguments of these methods, that's better than forbidding that the input be scalar...

Why was the input scalar then? I think we have many tests to make sure the input data has the right shape. So even with only 1 feature, we should always be working with 2D arrays. Am I missing something?

wdevazelhes:
I just removed the useless `np.atleast2d` (see comments https://github.com/metric-learn/metric-learn/pull/152#issuecomment-457125468 and https://github.com/metric-learn/metric-learn/pull/152#issuecomment-458179822), so I think we should be good to go

wdevazelhes:
Ah yes, I forgot about the docs, I udpated the commented part not to forget to add it when in other PRs I'll update the doc.

PR #159, opened 2019-01-23
[MRG] MAINT: remove variables not needed to store

### LINKED ISSUE #134: Avoid storing any unnecessary variable after fit

wdevazelhes:
Fixes #134

### DIFF
--- a/metric_learn/itml.py
+++ b/metric_learn/itml.py
@@ -73,9 +73,9 @@ class _BaseITML(MahalanobisMixin):
     self.bounds_[self.bounds_==0] = 1e-9
     # init metric
     if self.A0 is None:
-      self.A_ = np.identity(pairs.shape[2])
+      A = np.identity(pairs.shape[2])
     else:
-      self.A_ = check_array(self.A0)
+      A = check_array(self.A0, copy=True)
     gamma = self.gamma
     pos_pairs, neg_pairs = pairs[y == 1], pairs[y == -1]
     num_pos = len(pos_pairs)
@@ -87,7 +87,6 @@ class _BaseITML(MahalanobisMixin):
     neg_bhat = np.zeros(num_neg) + self.bounds_[1]
     pos_vv = pos_pairs[:, 0, :] - pos_pairs[:, 1, :]
     neg_vv = neg_pairs[:, 0, :] - neg_pairs[:, 1, :]
-    A = self.A_
 
     for it in xrange(self.max_iter):
       # update positives
@@ -125,7 +124,7 @@ class _BaseITML(MahalanobisMixin):
       print('itml converged at iter: %d, conv = %f' % (it, conv))
     self.n_iter_ = it
 
-    self.transformer_ = transformer_from_metric(self.A_)
+    self.transformer_ = transformer_from_metric(A)
     return self
 
 
@@ -134,6 +133,17 @@ class ITML(_BaseITML, _PairsClassifierMixin):
 
   Attributes
   ----------
+  bounds_ : `list` of two numbers
+      Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+      all given pairs of similar points ``a`` and ``b``, and
+      ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
+      ``d``, with ``bounds=[ pos, neg]``, and ``d`` the learned distance. If
+      not provided at initialization, these are the ones derived at train
+      time.
+
+  n_iter_ : `int`
+      The number of iterations the solver has ran.
+
   transformer_ : `numpy.ndarray`, shape=(num_dims, n_features)
       The linear transformation ``L`` deduced from the learned Mahalanobis
       metric (See function `transformer_from_metric`.)
@@ -151,8 +161,12 @@ class ITML(_BaseITML, _PairsClassifierMixin):
         preprocessor.
     y: array-like, of shape (n_constraints,)
         Labels of constraints. Should be -1 for dissimilar pair, 1 for similar.
-    bounds : list (pos,neg) pairs, optional
-        bounds on similarity, s.t. d(X[a],X[b]) < pos and d(X[c],X[d]) > neg
+    bounds : `list` of two numbers
+        Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+        all given pairs of similar points ``a`` and ``b``, and
+        ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
+        ``d``, with ``bounds=[pos, neg]``, and ``d`` the learned distance.
+        If not provided at initialization, these will be derived at train time.
 
     Returns
     -------
@@ -167,6 +181,17 @@ class ITML_Supervised(_BaseITML, TransformerMixin):
 
   Attributes
   ----------
+  bounds_ : `list` of two numbers
+      Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+      all given pairs of similar points ``a`` and ``b``, and
+      ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
+      ``d``, with ``bounds=[pos, neg]``, and ``d`` the learned distance. If
+      not provided at initialization, these are the ones derived at train
+      time.
+
+  n_iter_ : `int`
+      The number of iterations the solver has ran.
+
   transformer_ : `numpy.ndarray`, shape=(num_dims, n_features)
       The linear transformation ``L`` deduced from the learned Mahalanobis
       metric (See function `transformer_from_metric`.)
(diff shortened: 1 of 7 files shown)


metric_learn/itml.py
```diff
@@ -167,6 +181,17 @@ class ITML_Supervised(_BaseITML, TransformerMixin):
 
   Attributes
   ----------
+  bounds_ : `list` of two numbers
+      Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+      all given pairs of similar points ``a`` and ``b``, and
+      ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
+      ``d``, with ``bounds=[pos, neg]``, and ``d`` the learned distance. If
+      not provided at initialization, these are the ones derived at train
+      time.
```
wdevazelhes: I am not familiar with `ITML` but the above is based on what I understood of it. Feel free to tell me if I'm wrong

metric_learn/itml.py
```diff
@@ -151,8 +161,12 @@ def fit(self, pairs, y, bounds=None):
...
     y: array-like, of shape (n_constraints,)
         Labels of constraints. Should be -1 for dissimilar pair, 1 for similar.
-    bounds : list (pos,neg) pairs, optional
-        bounds on similarity, s.t. d(X[a],X[b]) < pos and d(X[c],X[d]) > neg
+    bounds : `list` of two numbers
+        Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+        all given pairs of similar points ``a`` and ``b``, and
+        ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
+        ``d``, with ``bounds=[pos, neg]``, and ``d`` the learned distance.
+        If not provided at initialization, these will be derived at train time.
```
wdevazelhes: I also updated the description of the argument in `fit` (`bounds`), to match the one of `self.bounds_`, and to be more detailed. Same comment: feel free to tell me if I'm wrong

metric_learn/itml.py
perimosocordiae: As currently implemented, `bounds_` should be an ndarray-like of shape (2,).

We should probably update the code to accept list or tuple types for the bounds as well.

perimosocordiae: "these are the ones" -> "these are"

metric_learn/itml.py
```diff
@@ -167,6 +181,17 @@ class ITML_Supervised(_BaseITML, TransformerMixin):
...
+  bounds_ : `list` of two numbers
+      Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+      all given pairs of similar points ``a`` and ``b``, and
+      ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
+      ``d``, with ``bounds=[pos, neg]``, and ``d`` the learned distance. If
+      not provided at initialization, these are the ones derived at train
+      time.
+
+  n_iter_ : `int`
+      The number of iterations the solver has ran.
```
perimosocordiae: "has ran" -> "has run"

wdevazelhes: >As currently implemented, bounds_ should be an ndarray-like of shape (2,).
>
>We should probably update the code to accept list or tuple types for the bounds as well.

That's right, I forgot about that. For now I'll just change the docstring, but I'll open an issue to get this thing straight

> "these are the ones" -> "these are"

Thanks, will do 


> "has ran" -> "has run"

That's right, thanks

metric_learn/itml.py
```diff
@@ -181,16 +181,16 @@ class ITML_Supervised(_BaseITML, TransformerMixin):
 
   Attributes
   ----------
-  bounds_ : `list` of two numbers
+  bounds_ : array-like, shape=(2,)
       Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
       all given pairs of similar points ``a`` and ``b``, and
       ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
       ``d``, with ``bounds=[pos, neg]``, and ``d`` the learned distance. If
```
wdevazelhes: I keep saying ``bounds=[pos, neg]`` here because it can be still be a list here, and I think this is more general than ``bounds=array([pos, neg])`` which would make think more of a numpy array only

metric_learn/itml.py
```diff
@@ -134,6 +133,17 @@ class ITML(_BaseITML, _PairsClassifierMixin):
 
   Attributes
   ----------
+  bounds_ : array-like, shape=(2,)
+      Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+      all given pairs of similar points ``a`` and ``b``, and
+      ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
+      ``d``, with ``bounds=[ pos, neg]``, and ``d`` the learned distance. If
+      not provided at initialization, these are derived at train
```
bellet: maybe it would be nice to explain how it is done then: bounds_[0] and bounds_[1] are set to the 5th and 95th percentile of the pairwise distances among all points available at training?
for supervised version, of all points in the training data `X` ; for weakly supervised, of all points present in `pairs`

metric_learn/itml.py
bellet: again

metric_learn/itml.py
```diff
@@ -167,6 +181,17 @@ class ITML_Supervised(_BaseITML, TransformerMixin):
 
   Attributes
   ----------
+  bounds_ : array-like, shape=(2,)
+      Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+      all given pairs of similar points ``a`` and ``b``, and
+      ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
+      ``d``, with ``bounds=[pos, neg]``, and ``d`` the learned distance. If
+      not provided at initialization, these are derived at train
+      time.
```
bellet: again

metric_learn/itml.py
```diff
@@ -193,8 +218,12 @@ def __init__(self, gamma=1., max_iter=1000, convergence_threshold=1e-3,
...
     num_constraints: int, optional
         number of constraints to generate
-    bounds : list (pos,neg) pairs, optional
-        bounds on similarity, s.t. d(X[a],X[b]) < pos and d(X[c],X[d]) > neg
+    bounds : `list` of two numbers
+      Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+      all given pairs of similar points ``a`` and ``b``, and ``d(c, d) > neg``
+      for all given pairs of dissimilar points ``c`` and ``d``, with
+      ``bounds=[pos, neg]``, and ``d`` the learned distance. If not provided at
+      initialization, these will be derived at train time.
```
bellet: again

metric_learn/itml.py
wdevazelhes: Yes I agree it's more precise

metric_learn/itml.py
wdevazelhes: will do

metric_learn/itml.py
wdevazelhes: will do

metric_learn/itml.py
wdevazelhes: will do

wdevazelhes:
Regarding `bounds`, I also noticed that for the supervised version it has to be given at initalization, and for the weakly supervised version it has to be put at fit. I think we should uniformize that. Since it is a data dependent parameter (indeed the default setting is deduced from the data), I think we should put them in fit. Since this is out of the scope of this PR and it should have a deprecation warning (which adds again code in this PR), I'll open another PR for that, and for this one I'll let the parameter where they are.

wdevazelhes:
I adressed all the comments, so we should be good to merge this one

metric_learn/itml.py
```diff
@@ -134,6 +133,18 @@ class ITML(_BaseITML, _PairsClassifierMixin):
 
   Attributes
   ----------
+  bounds_ : array-like, shape=(2,)
+      Bounds on similarity, aside slack variables, s.t. ``d(a, b) < pos`` for
+      all given pairs of similar points ``a`` and ``b``, and
+      ``d(c, d) > neg`` for all given pairs of dissimilar points ``c`` and
+      ``d``, with ``bounds=[ pos, neg]``, and ``d`` the learned distance. If
```
bellet: My suggestion was to get rid of bounds=[ pos, neg] altogether and replace pos and neg in the inequalities by bounds_[0] and bounds_[1]

wdevazelhes: Ah yes I agree it's even clearer, done

Issue #138, opened 2018-12-20
SDML does not work in sandwich example


bellet:
#### Description
SDML does not seem to do anything in the sandwich example from the doc (see figure below). We should make sure it is due to the default hyperparameters (and fix them in the example is that's the case).

![sphx_glr_plot_sandwich_001](https://user-images.githubusercontent.com/7905665/50301250-1de2a700-0487-11e9-9525-4a6cd9706fd1.png)

wdevazelhes:
In a previous version of metric-learn, scipy, and numpy, the example seemed to work. For instance, I have this result for theses versions: 
Linux-4.4.0-141-generic-x86_64-with-debian-stretch-sid
('Python', '2.7.15 |Anaconda, Inc.| (default, Dec 14 2018, 19:04:19) \n[GCC 7.3.0]')
('NumPy', '1.11.3')
('SciPy', '0.17.1')
('Scikit-Learn', '0.17.1')
metric-learn: from commit ddfac991c0d32f3931831d649133dfbd15c61eff

![figure_1-2](https://user-images.githubusercontent.com/31916524/51022270-1729ea00-1584-11e9-9bf5-fcc4ec62159a.png)

I guess there is some problem coming from new versions of either metric-learn, scikit-learn, scipy or numpy...

wdevazelhes:
Related to #154

Issue #154, opened 2019-01-11
SDML raises SDP error on iris


wdevazelhes:
#### Description
Trying to update the `metric_plotting.ipynb` notebook, it didn't work for SDML, so here is a minimal example to reproduce. 
 
#### Steps/Code to Reproduce

```python
import numpy as np
from sklearn.datasets import load_iris
from metric_learn import SDML_Supervised
dataset = load_iris()
X, y = dataset.data, dataset.target
sdml = SDML_Supervised(num_constraints=200)
sdml.fit(X, y, random_state = np.random.RandomState(1234))
```

#### Expected Results
No error is thrown

#### Actual Results

```pytb
/home/will/anaconda3/envs/standard/lib/python3.7/site-packages/sklearn/utils/validation.py:761: DataConversionWarning: A column-vector y was passed when a 1d array was expected. Please change the shape of y to (n
_samples, ), for example using ravel().
  y = column_or_1d(y, warn=True)
/home/will/anaconda3/envs/standard/lib/python3.7/site-packages/sklearn/utils/deprecation.py:77: DeprecationWarning: Function pinvh is deprecated; sklearn.utils.extmath.pinvh was deprecated in version 0.19 and wil
l be removed in 0.21. Use scipy.linalg.pinvh instead.
  warnings.warn(msg, category=DeprecationWarning)
/home/will/anaconda3/envs/standard/lib/python3.7/site-packages/sklearn/utils/deprecation.py:77: DeprecationWarning: Function pinvh is deprecated; sklearn.utils.extmath.pinvh was deprecated in version 0.19 and wil
l be removed in 0.21. Use scipy.linalg.pinvh instead.
  warnings.warn(msg, category=DeprecationWarning)
/home/will/anaconda3/envs/standard/lib/python3.7/site-packages/sklearn/utils/deprecation.py:77: DeprecationWarning: Function graph_lasso is deprecated; The 'graph_lasso' was renamed to 'graphical_lasso' in versio
n 0.20 and will be removed in 0.22.
  warnings.warn(msg, category=DeprecationWarning)
---------------------------------------------------------------------------
FloatingPointError                        Traceback (most recent call last)
<ipython-input-2-caf37592524f> in <module>
      3 X, y = load_iris(return_X_y=True)
      4 sdml = SDML_Supervised(num_constraints=200)
----> 5 sdml.fit(X, y, random_state = np.random.RandomState(1234))

[...]

PR #162, opened 2019-01-28
[MRG] FIX: make proposal for sdml formulation

wdevazelhes:
Digging into SDML's code and paper, I don't understand some parts of the implementation. I think it should be as proposed in this PR. Tell me if I'm wrong

Looking at these this paper about SDML https://icml.cc/Conferences/2009/papers/46.pdf, line , and this paper about Graphical Lasso http://statweb.stanford.edu/~tibs/ftp/graph.pdf: it seems that in SDML, we want to optimize equation 8, which can indeed be done with Graphical Lasso according to the paper on Graphical Lasso. In fact, equation 8 in SDML's paper is the same as equation 1 in Graphical Lasso's paper (up to a minus sign). The following variables are equivalent:

|SDML's paper | Graphical Lasso's paper|
|-------------------|-------------------------------|
|M|theta|
|P|S|

where in SDML's paper, `P = M_0^-1 + etha * X.L.X^T`.

But note that in SDML's paper, M_0^-1 is the inverse of the a priori Mahalanobis matrix, which can be indeed initialized to the inverse of the covariance matrix. So then M_0^-1 will be the inverse of the inverse of the covariance matrix hence the covariance matrix itself.

So we should just compute `P = emp. covariance matrix + self.balance_param * loss_matrix` and do graphical lasso on this (and not: inverse the covariance matrix, do `P = this_inverse + self.balance_param * loss_matrix`, inverse the result, and compute Graphical Lasso on this as it is done currently)

And in both cases, we want to evaluate the sparse inverse of M/theta, so things are OK for that

Also, I didn't get the hack to ensure positive semidefinite, doesn't it change the result ?

This PR's modification does not fix the issues we had (like `plot_sandwich.py` does not have better results with this). So maybe let's not merge it until we have a whole fix for SDML.

There are other things that could explain why SDML doesn't work, like choosing the optimization parameter alpha, and also that `graph_lasso` seems to sometimes work badly, see https://github.com/scikit-learn/scikit-learn/issues/6887 and https://github.com/scikit-learn/scikit-learn/issues/11417)

So first tell me if you agree with this modification (not on merging it or not, just whether it's the right formula or not), so that we can look elsewhere to see how to fix SDML's error.

TODO: 
- [x] Put a message just before merge, in the release draft, to announce that skggm is needed for `SDML`
- [x] replace sklearn's pinvh (deprecated) by scipy's pinvh
- [x] deal with the 1D input case for sdml
- [x] Add a small test on the non SPD case that skggm can solve
- [x] Make travis indeed install skggm if the version allows it


wdevazelhes:
> I think this is correct except for the initialization: M_0 should be either identity or the empirical covariance matrix (see SDML paper, beginning of page 7, where these two possible initializations are mentioned).

Ah that's right, I didn't see that. But isn't it weird to use the covariance matrix as a prior ? I thought we would usually take the inverse covariance matrix (Mahalanobis Matrix)? Like for instance if we take the balance parameter to zero the learned metric would be a sparse covariance matrix, wouldn't we want a sparse precision matrix (mahalanobis matrix) instead ? 

 

> Or more generally (this is probably how things should be written): `P = pinvh(prior) + self.balance_param * loss_matrix`
> where `prior` is either the identity or the covariance matrix

Yes I agree I should have written things like that

wdevazelhes:
> Once this has been updated, if it does not fix the problem, here are some next things to try:
> 
>     * switch to identity as the prior
> 
>     * add a tiny bit of identity to the empirical covariance to make sure it is PSD (it might not due to some small numerical errors)
> 
>     * try varying the parameters `balance_param` and `sparsity_param` (which corresponds to parameter `alpha` of graph_lasso) to see if it allows to obtain satisfying results on the buggy cases of sandwich and iris
> 
>     * try also to increase a bit the parameter `eps` of graph_lasso when getting warnings/errors about ill conditioned system

I agree, I'll test this and report the result here

bellet:
> > I think this is correct except for the initialization: M_0 should be either identity or the empirical covariance matrix (see SDML paper, beginning of page 7, where these two possible initializations are mentioned).
> 
> Ah that's right, I didn't see that. But isn't it weird to use the covariance matrix as a prior ? I thought we would usually take the inverse covariance matrix (Mahalanobis Matrix)? Like for instance if we take the balance parameter to zero the learned metric would be a sparse covariance matrix, wouldn't we want a sparse precision matrix (mahalanobis matrix) instead ?

Yes in fact you are right, it must be a typo. M_0 should be the **inverse** covariance matrix so that both terms of `P` are some sort of covariance matrices

wdevazelhes:
An update on this topic: using the package http://github.com/skggm/skggm, replacing `graph_lasso` by their function `quic`, it works, as long as we specify an SPD initialization.

bellet:
Until the sklearn version is made more robust (as intended in https://github.com/scikit-learn/scikit-learn/issues/12228), I would vote for temporarily switching to skggm so that we have a working implementation

mnarayan:
One of the skggm developers here. Do let us know how using skggm works out for you.

test/metric_learn_test.py
```diff
@@ -141,7 +141,7 @@ def test_iris(self):
     sdml = SDML_Supervised(num_constraints=1500)
     sdml.fit(self.iris_points, self.iris_labels, random_state=rs)
     csep = class_separation(sdml.transform(self.iris_points), self.iris_labels)
-    self.assertLess(csep, 0.25)
+    self.assertLess(csep, 0.20)
```
wdevazelhes: The score that I had when running the example was 0.1933487414312185 so I thought let's be a bit more conservative that 0.25 so that if it fails we would be warned

wdevazelhes:
This is failing because of https://github.com/skggm/skggm/issues/119

perimosocordiae:
We shouldn't need to specify Cython as a dependency unless we have Cython code in metric-learn itself.

wdevazelhes:
> We shouldn't need to specify Cython as a dependency unless we have Cython code in metric-learn itself.


Ah yes that's right, I could install it without explicitly installing cython indeed, even if skggm installs cython itself (among other packages)

wdevazelhes:
Oh but it worked locally, with python 3.7. With python 3.4 (like travis), is fails indeed. Maybe that's because skggm say they are compatible for python 3.6.x ? Maybe that's the occasion to move to python 3.6.x for metric learn ? According to pip messages I have when installing on python 3.4: "Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429)"

metric_learn/sdml.py
```diff
@@ -64,17 +64,25 @@ def _fit(self, pairs, y):
...
       prior = np.identity(pairs.shape[2])
     diff = pairs[:, 0] - pairs[:, 1]
     loss_matrix = (diff.T * y).dot(diff)
     emp_cov = pinvh(prior) + self.balance_param * loss_matrix
-    M, _, _, _, _, _ = quic(emp_cov, lam=self.sparsity_param, msg=self.verbose,
-                            Theta0=np.eye(pairs.shape[2]),
-                            Sigma0=np.eye(pairs.shape[2]))
 
-    self.transformer_ = transformer_from_metric(M)
+    # our initialization will be the matrix with emp_cov's eigenvalues,
```
wdevazelhes: This is an init that we talked about with @bellet, that I found worked better (allowed tests to pass when with identity I had a lot of Linalg Error)

test/metric_learn_test.py
```diff
@@ -166,8 +166,7 @@ def test_deprecation_num_labeled(self):
       # test that a deprecation message is thrown if num_labeled is set at
       # initialization
       # TODO: remove in v.0.6
-      X = np.array([[0, 0], [0, 1], [2, 0], [2, 1]])
-      y = np.array([1, 0, 1, 0])
+      X, y = make_classification()
```
wdevazelhes: The test was failing before so I changed it, I think this is reasonable since what was before was not realistic

test/test_fit_transform.py
```diff
@@ -63,6 +64,8 @@ def test_lmnn(self):
     assert_array_almost_equal(res_1, res_2)
 
   if has_installed_skggm():
+    # TODO: remove the mark when SDML has become deterministic
```
wdevazelhes: SDML is not deterministic (see skggm issue https://github.com/skggm/skggm/issues/121), so for now I exclude it from this test, until we find a solution (I don't think it should prevent merging this PR though)

test/test_mahalanobis_mixin.py
```diff
@@ -96,7 +96,7 @@ def check_is_distance_matrix(pairwise):
   assert np.array_equal(pairwise, pairwise.T)  # symmetry
   assert (pairwise.diagonal() == 0).all()  # identity
   # triangular inequality
-  tol = 1e-15
+  tol = 1e-12
```
wdevazelhes: SDML was failing due to the harsh tolerance so I changed it but I think it's still reasonable

test/test_mahalanobis_mixin.py
```diff
@@ -272,14 +272,17 @@ def test_get_squared_metric(estimator, build_dataset):
...
   """Tests that the transformer of metric learners is 2D"""
-  input_data, labels, _, X = build_dataset()
-  model = clone(estimator)
-  set_random_state(model)
-  # test that it works for X.shape[1] features
-  model.fit(input_data, labels)
-  assert model.transformer_.shape == (X.shape[1], X.shape[1])
+  # TODO: remove this check when SDML has become robust to 1D elements,
+  #  or when the 1D case is dealt with separately
+  if not str(estimator).startswith('SDML'):
```
wdevazelhes: SDML does not deal well with 1D inputs (it returned nan). We could investigate in another PR, or here 
(one quick solution that would work would simply be to deal with the 1D case since the covariance matrix would be only one element so maybe we can decide what to do then)

test/test_sklearn_compat.py
```diff
@@ -72,7 +72,7 @@ def test_itml(self):
   def test_mmc(self):
     check_estimator(dMMC)
 
-  # This fails due to a FloatingPointError
+  # TODO: uncomment this when SDML has become deterministic
```
wdevazelhes: I would have liked to uncomment this since there is not FloatingPointError anymore, but there is still the undeterminacy of SDML that makes it fail

test/test_sklearn_compat.py
```diff
@@ -115,37 +115,39 @@ def test_cross_validation_manual_vs_scikit(estimator, build_dataset,
...
-    scores, predictions = [], np.zeros(input_data.shape[0])
-    for fold_size in fold_sizes:
-      start, stop = current, current + fold_size
-      current = stop
-      test_slice = slice(start, stop)
-      train_mask = np.ones(input_data.shape[0], bool)
-      train_mask[test_slice] = False
-      y_train, y_test = labels[train_mask], labels[test_slice]
-      estimator.fit(input_data[train_mask], y_train)
+  # TODO: remove this check when SDML has become deterministic
```
wdevazelhes: Same reason here

metric_learn/sdml.py
```diff
@@ -81,7 +81,6 @@ def _fit(self, pairs, y):
     M, _, _, _, _, _ = quic(emp_cov, lam=self.sparsity_param,
                             msg=self.verbose,
                             Theta0=theta0, Sigma0=sigma0)
-    M = M + 1e-10 * np.eye(M.shape[0])  # to ensure M is positive semi-definite
```
wdevazelhes: This is something I had added to pass tests but then I realized with the good initialization (see lines above, in the previous commit), I don't need this anymore so I delete it

wdevazelhes:
I realized there is a pathological case where the optimization problem diverges (has no solution theoretically): 
if the pseudo-covariance that we want to reconstruct (i.e. the input of ``quic``) is diagonal with at least one negative diagonal value, then the problem becomes: 
``argmin theta_i: sum(d_i * theta_i - log (theta_i))``, with ``d_i``diagonal coefficient `i` (because in the diagonal case the determinant is the product thus we can develop the log, and also because the L1 penalty is only applied to non-diagonal coefficients hence not here)
This sum can be minimized by minimizing for each i since the terms are independent, and if `d_i >0` the optimal `theta_i` is `1/d_i`. But if `d_i<0`, then as long as we increase `theta_i` the term is smaller, so it diverges.

see for instance: 

```python
import numpy as np
from inverse_covariance import quic
quic(np.array([[-1., 0], [0, 2.]]), lam=0.01)
```
Returns:
```python
(array([[nan,  0.],
        [ 0., nan]]), array([[nan, nan],
        [nan, nan]]), -1.9323349832290746e+244, 4.1e-05, 12, -1.9323349832290746e+244)
```

In terms of real life inputs, this could happen for instance in the following example:
```python
In [1]: import numpy as np
   ...: from metric_learn import SDML
   ...: 
   ...: pairs = np.array([[[-10., 0.], [10., 0.]], [[0., 50.], [0., -60]]])
   ...: y = [1, -1]
   ...: sdml = SDML()
   ...: sdml.fit(pairs, y)
   ...: sdml.get_mahalanobis_matrix()
   ...: 
/home/will/Code/metric-learn/metric_learn/sdml.py:66: FutureWarning: arrays to stack must be passed as a "sequence" type such as list or tuple. Support for non-sequence iterables such as generators is deprecated as of NumPy 1.16 and will raise an error in the future.
  X = np.vstack({tuple(row) for row in pairs.reshape(-1, pairs.shape[2])})

/home/will/anaconda3/envs/test/lib/python3.6/site-packages/numpy/linalg/linalg.py:2093: RuntimeWarning: invalid value encountered in det
  r = _umath_linalg.det(a, signature=signature)
Out[1]: 
array([[nan, nan],
       [nan, nan]])
```
Returns:
```python
/home/will/Code/metric-learn/metric_learn/sdml.py:66: FutureWarning: arrays to stack must be passed as a "sequence" type such as list or tuple. Support for non-sequence iterables such as generators is deprecated as of NumPy 1.16 and will raise an error in the future.
  X = np.vstack({tuple(row) for row in pairs.reshape(-1, pairs.shape[2])})
/home/will/anaconda3/envs/test/lib/python3.6/site-packages/numpy/linalg/linalg.py:2093: RuntimeWarning: invalid value encountered in det
  r = _umath_linalg.det(a, signature=signature)
Out[14]: 
array([[nan, nan],
       [nan, nan]])
```

(Remark: Here there could be just the problem that the initialization has a 0 eigenvalue that could be solved by just adding an epsilon, but anyway there should still be a problem since the algorithm cannot converge)


I didn't manage to see whether this is a particular case of a more general problem than diagonal matrices, but I think it is significant and realistic enough to need to adress the problem

Here are some possible options (some might be impossible, I didn't check them deeply yet):

- Add an L2 regularization to ensure the coefficients don't grow to much ? (which is btw what is proposed in this issue in scikit-learn: https://github.com/scikit-learn/scikit-learn/issues/12228)
- Use the L1 regularization also for diagonal coefficients ? (which is one option of quic)
- Try to formulate an "equivalent" problem that would be solvable and still achieve the result we want (put similar points together or dissimilar points together)
- Maybe say that we should avoid these situations with a right balance parameter, i.e. warn the user to decrease the balance parameter if we are in these situations (indeed, we can interpret SDML like this: If we don't account for the regularization term (like it's the case in our problematic case), there is a force (f1) that makes the transformation matrix grow (but with a logarithmic intensity) (the log det term), and there is another force that at the same time wants to bring similar points together, dissimilar points together, AND keep every points close to the mean (f2) (because we can see the covariance term inside the "pseudo-covariance" as additional constraints for every points to be closer to the mean). Normally f2 should be an attractive force so that it will find an equilibrium with f1, but if say the balance parameter `etha` is very high and there are a lot of negative constraints, f2 can become a repulsive force (in some direction(s)), and hence in some directions there would be only repulsive force hence it would diverge. Hence maybe there should be a sort of limit balance parameter that would ensure the system wouldn't diverge

Note: It is possible that divergence problems could arise also for other algorithms, if they do not have a regularisation, or a gradient that vanishes itself when the solution is "good enough" (like NCA): indeed the example used above shows a case where we have four points, but the constraints are "perpendicular", which means that we can do a good result by collapsing one dimension so that the two points that should be similar are infinitely close to each other, and extenting  the other perpendicular dimension so that points that should be dissimilar are infinitely far away from each other.

mnarayan:
The inverse covariance estimation has only been formulated for positive
definite or semidefinite cases. Variances are never negative,  so a
negative diagonal value should be fixed. Adding a large enough constant `c`
so that all the diagonal values are increased by `c` is an option.

wdevazelhes:
> The inverse covariance estimation has only been formulated for positive definite or semidefinite cases. Variances are never negative, so a negative diagonal value should be fixed. 

The problem is that in our algorithm, we have a pseudo-covariance matrix (not a covariance matrix from real samples, but a matrix that take into account user specified constraints), that can happen to have negative eigenvalues,
I don't know if that would make the problem diverge in any case though (I just know it would in the diagonal case, I don't know for any non SPD matrix)

> Adding a large enough constant `c` so that all the diagonal values are increased by `c` is an option.

Yes, that's right, thanks, this would also solve the problem, I added it to the list above

bellet:
Thanks for the detailed report. Given the formulation of SDML I think it makes sense that it can fail in some extreme cases if `loss_matrix` is too far from being PSD and the importance of the prior (which is PSD) is too small (i.e. `balance_param` is too large). `balance_param` is an hyperparameter that should definitely be tuned.

So maybe we could simply check, before calling the graph lasso solver, whether `emp_cov` is PSD. If it is not, print a warning about it to say that it may lead to degenerate solutions, and that `balance_param` can be reduced to avoid this.

Note that the trick of adding a constant large enough to the diagonal corresponds to using the identity as prior and picking `balance_param` small enough

wdevazelhes:
@bellet that sounds good, I'll add this warning
Now that you solved graph lasso's bug in https://github.com/scikit-learn/scikit-learn/pull/13276,  we can use scikit-learn's graph lasso ! (if scikit-learn's release is made before the next metric-learn's release) 
We should still have skggm's support for python 2.7 users though, since next scikit-learn's release won't support python 2.7 anymore. But for the release after we should only need scikit-learn :)

test/test_mahalanobis_mixin.py
```diff
@@ -284,5 +284,14 @@ def test_transformer_is_2D(estimator, build_dataset):
 
     # test that it works for 1 feature
     trunc_data = input_data[..., :1]
+    # we drop duplicates that might have been formed, i.e. of the form
```
wdevazelhes: I did this to make tests pass, they didn't because of #177 and #176 
The dataset created here is reasonable: we just drop duplicates which what a good user would do in practice. It's not like if we created a very weird dataset to prevent the tests from failing, so it's ok to do that. #177 and #176 need to be fixed and tested separately in other PRs

test/test_mahalanobis_mixin.py
```diff
@@ -272,26 +272,25 @@ def test_get_squared_metric(estimator, build_dataset):
...
-      for slice_idx in slices[trunc_data.shape[1]]:
-        _, indices = np.unique(trunc_data[:, slice_idx, :], axis=2,
-                               return_index=True)
-        trunc_data = trunc_data[indices]
-        labels = labels[indices]
-    model.fit(trunc_data, labels)
-    assert model.transformer_.shape == (1, 1)  # the transformer must be 2D
+  input_data, labels, _, X = build_dataset()
+  model = clone(estimator)
+  if model.__class__.__name__.startswith('SDML'):
```
wdevazelhes: this is a special check that in the case where we have SDML, we ensure we are in the case where the matrix to reconstruct is PSD. This is a specific check but it's done sometimes in scikit-learn, (e.g. transforming the data into a square matrix for algorithms with kernel='precomputed', and I could find more). It's ok, because the purpose of this test is just to test the form (2D) of the output in some standard case. Corner cases behaviours shoudl be (and are) dealt with separately (in this PR we decide to raise a warning in the case where the matrix to reconstruct is not PSD). 
We could also ensure to have one good form of the input good for all estimators, but first it's a bit more complicated (we would need to find a reasonable (to work with all alrogithms) dataset with 1 feature such as the matrix is PSD), and if we add an algo that fails we have to rethink the dataset etc... So it's better (and simpler) to have a reasonable dataset (extracted from iris), and to tune the algorithms not to fail on it, since we just want to check they return a 2d metric in the case of 1d input here

test/test_mahalanobis_mixin.py
```diff
@@ -272,26 +272,25 @@ def test_get_squared_metric(estimator, build_dataset):
                          ids=ids_metric_learners)
 def test_transformer_is_2D(estimator, build_dataset):
   """Tests that the transformer of metric learners is 2D"""
-  # TODO: remove this check when SDML has become robust to 1D elements,
```
wdevazelhes: in fact the pb was not robustness to 1D elements, but the fact that in 1D if the metric is [[negative number]], the algo will diverge (see above, since it's diagonal with negative coeff). This has been dealt with in the rest of the PR

wdevazelhes:
See my edit of https://github.com/metric-learn/metric-learn/pull/162#issuecomment-469673749: we should be able to only use scikit-learn's graph lasso now, I'll just need to check it works and use it (and remove skggm dependencies)

wdevazelhes:
I just removed the use of skggm to use scikit-learn's graphical lasso. Note that a lot of tests were failing. They were having a non SPD problem. I changed the balance parameter and used the identity instead of covariance matrix in the formula to make it work. But this means that skggm solver was returning a result where now scikit-learn returns an error (although it's in a non SPD case. SPD cases seem to work fine with scikit-learn now).

I can look into scikit-learn on why it fails on these examples, but anyway I think changes in scikit-learn that would solve the pb would not be integrated until v21, which would mean not available for python 2.

So I think for this release we could go for the below alternative solution: 
- We first test if the matrix is SPD or not. If it is not we raise a warning as we do now. If it is, sklearn's method works, so let's now consider the case where it's not.
- We try to run scikit-learn's graphical lasso. If it doesn't work, we check if the users have installed skggm. 
- If they have, we say to them we'll retry with skggm. Otherwise we return an error like "sklearn's graphical lasso didn't converge. Maybe install skggm and try again (skggm's solver will then automatically be selected if sklearn's method fails.)"  (note that users would already have had the warning about the fact that they want to reconstruct a non SPD matrix, which would give them also alternative ways to solve their pb)

wdevazelhes:
After discussing with @bellet and @nvauquie , we decided to solve the problem in the following way: 
- if the user has installed skggm, SDML will use it, and print a message saying it uses it.
- if not, SDML will use scikit-learn's `graphical_lasso`, and print a warning saying it uses it, and that it can diverge in some non SPD cases where skggm converges, and it advises the user to install skggm.

For the testing, on travis:
- python2.7 and python3.6 version will test with skggm installed, so they will test that "using skggm's solver" is printed when fitting SDML
- python3.4 will test not having installed skggm, so they will test that the warning is returned advising to install skggm
For all the tests, I force SDML with a low balance param and use_cov=False so that both skggm's quic and scikit-learn's graphical_lasso can solve them. 

I will also just put a test with a non PSD case that skggm can solve without forcing the balance param and use_cov (and that graphical_lasso cannot solve), and test it only if skggm is installed (to ensure skggm works well in this case).

What do you think ?

metric_learn/constraints.py
```diff
@@ -96,6 +96,6 @@ def wrap_pairs(X, constraints):
   c = np.array(constraints[2])
   d = np.array(constraints[3])
   constraints = np.vstack((np.column_stack((a, b)), np.column_stack((c, d))))
-  y = np.vstack([np.ones((len(a), 1)), - np.ones((len(c), 1))])
+  y = np.hstack([np.ones((len(a),)), - np.ones((len(c),))])
```
wdevazelhes: In fact we should never return a column vector but a line vector (these are the ones scikit-learn likes to work on) 
Otherwise scikit-learn's checks called by our checks will return a warning

metric_learn/sdml.py
```diff
@@ -52,24 +55,50 @@ def __init__(self, balance_param=0.5, sparsity_param=0.01, use_cov=True,
...
+    # with a constant added so that they are all positive (plus an epsilon
+    # to ensure definiteness). This is empirical.
+    w, V = np.linalg.eigh(emp_cov)
+    if any(w < 0.):
+      warnings.warn("Warning, the input matrix of graphical lasso is not "
+                    "positive semi-definite (PSD). The algorithm may diverge, "
+                    "and lead to degenerate solutions. "
+                    "To prevent that, try to decrease the balance parameter "
+                    "`balance_param` and/or to set use_covariance=False.",
+                    ConvergenceWarning)
```
wdevazelhes: is `ConvergenceWarning` OK ? It's the one that seemed the more appropriate but here we raise it before even running the graphical lasso so maybe it's a bit weird... It would be better a `PossibleConvergenceWarning` kind of warning maybe ?

wdevazelhes:
Everything in the TODO is done, and the PR seems OK to me, it's ready for review !
Setting to MRG

metric_learn/constraints.py
perimosocordiae: maybe be simpler to do:

    y = np.concatenate([np.ones_like(a), -np.ones_like(c)])

metric_learn/_util.py
```diff
@@ -15,6 +15,14 @@ def vector_norm(X):
     return np.linalg.norm(X, axis=1)
 
 
+def has_installed_skggm():
```
perimosocordiae: I don't think this needs to be a function, as the answer isn't going to change during execution.

metric_learn/sdml.py
```diff
@@ -12,12 +12,15 @@
...
-from sklearn.utils.extmath import pinvh
+from scipy.linalg import pinvh
+from sklearn.covariance import graphical_lasso
+from sklearn.exceptions import ConvergenceWarning
 
 from .base_metric import MahalanobisMixin, _PairsClassifierMixin
 from .constraints import Constraints, wrap_pairs
-from ._util import transformer_from_metric
+from ._util import transformer_from_metric, has_installed_skggm
+if has_installed_skggm():
```
perimosocordiae: I'd prefer a simpler conditional import here:

```python
try:
  from inverse_covariance import quic
except ImportError:
  HAS_SKGGM = False
else:
  HAS_SKGGM = True
```

metric_learn/sdml.py
```diff
@@ -52,24 +55,50 @@ def __init__(self, balance_param=0.5, sparsity_param=0.01, use_cov=True,
...
   def _fit(self, pairs, y):
+    if not has_installed_skggm():
+      msg = ("Warning, skggm is not installed, so SDML will use "
+             "scikit-learn's graphical_lasso method. It can fail to converge"
+             "on some non SPD matrices where skggm would converge. If so, "
+             "try to install skggm. (see the README.md for the right "
+             "version.)")
+      warnings.warn(msg)
+    else:
+      print("SDML will use skggm's solver.")
```
perimosocordiae: We should only print this if `self.verbose` is true.

metric_learn/sdml.py
```diff
@@ -52,24 +55,50 @@ def __init__(self, balance_param=0.5, sparsity_param=0.01, use_cov=True,
     super(_BaseSDML, self).__init__(preprocessor)
 
   def _fit(self, pairs, y):
+    if not has_installed_skggm():
+      msg = ("Warning, skggm is not installed, so SDML will use "
+             "scikit-learn's graphical_lasso method. It can fail to converge"
+             "on some non SPD matrices where skggm would converge. If so, "
+             "try to install skggm. (see the README.md for the right "
+             "version.)")
```
perimosocordiae: Perhaps we can catch the case where scikit-learn's version fails and emit the warning then?

metric_learn/sdml.py
```diff
@@ -52,24 +55,50 @@ def __init__(self, balance_param=0.5, sparsity_param=0.01, use_cov=True,
...
     diff = pairs[:, 0] - pairs[:, 1]
     loss_matrix = (diff.T * y).dot(diff)
-    P = M + self.balance_param * loss_matrix
-    emp_cov = pinvh(P)
-    # hack: ensure positive semidefinite
-    emp_cov = emp_cov.T.dot(emp_cov)
-    _, M = graph_lasso(emp_cov, self.sparsity_param, verbose=self.verbose)
-
-    self.transformer_ = transformer_from_metric(M)
+    emp_cov = pinvh(prior) + self.balance_param * loss_matrix
```
perimosocordiae: Seems odd to round-trip the covariance matrix through two `pinvh` calls in the `self.use_cov` case.

metric_learn/sdml.py
perimosocordiae: I'm fine with it.

metric_learn/sdml.py
```diff
@@ -52,24 +55,50 @@ def __init__(self, balance_param=0.5, sparsity_param=0.01, use_cov=True,
...
+    # to ensure definiteness). This is empirical.
+    w, V = np.linalg.eigh(emp_cov)
+    if any(w < 0.):
+      warnings.warn("Warning, the input matrix of graphical lasso is not "
+                    "positive semi-definite (PSD). The algorithm may diverge, "
+                    "and lead to degenerate solutions. "
+                    "To prevent that, try to decrease the balance parameter "
+                    "`balance_param` and/or to set use_covariance=False.",
+                    ConvergenceWarning)
+    sigma0 = (V * (w - min(0, np.min(w)) + 1e-10)).dot(V.T)
```
perimosocordiae: Maybe simpler:

```python
min_eigval = w.min()
if min_eigval < 0:
  warnings.warn(...)
  min_eigval = 0

w += 1e-10 - min_eigval
sigma0 = (V * w).dot(V.T)
```

README.rst
```diff
@@ -20,7 +20,7 @@ Metric Learning algorithms in Python.
 **Dependencies**
 
 -  Python 2.7+, 3.4+
--  numpy, scipy, scikit-learn
+-  numpy, scipy, scikit-learn, and skggm (commit `a0ed406 <https://github.com/skggm/skggm/commit/a0ed406586c4364ea3297a658f415e13b5cbdaf8>`_) for `SDML`
```
perimosocordiae: I'd call out skggm separately as an optional dependency.

metric_learn/sdml.py
perimosocordiae: We'd need to duplicate the logic in the tests, but I'm fine with that.

setup.py
```diff
@@ -38,6 +38,7 @@
       extras_require=dict(
           docs=['sphinx', 'shinx_rtd_theme', 'numpydoc'],
           demo=['matplotlib'],
+          sdml=['skggm']
```
perimosocordiae: Can we specify a commit hash here? Or maybe since their latest release is 0.2.8, we could specify `skggm>=0.2.9`.

doc/getting_started.rst
```diff
@@ -15,7 +15,7 @@ Alternately, download the source repository and run:
 **Dependencies**
 
 -  Python 2.7+, 3.4+
--  numpy, scipy, scikit-learn
+-  numpy, scipy, scikit-learn, and skggm (commit `a0ed406 <https://github.com/skggm/skggm/commit/a0ed406586c4364ea3297a658f415e13b5cbdaf8>`_) for `SDML`
```
bellet: same comment here as https://github.com/metric-learn/metric-learn/pull/162/files#r264683385

metric_learn/sdml.py
bellet: I agree

metric_learn/sdml.py
bellet: And maybe clarify: skggm's graphical lasso solver

test/test_mahalanobis_mixin.py
```diff
@@ -281,5 +281,15 @@ def test_transformer_is_2D(estimator, build_dataset):
...
   # test that it works for 1 feature
   trunc_data = input_data[..., :1]
+  # we drop duplicates that might have been formed, i.e. of the form
+  # aabc or abcc or aabb for quadruplets, and aa for pairs.
+  slices = {4: [slice(0, 2), slice(2, 4)], 2: [slice(0, 2)]}
+  if trunc_data.ndim == 3:
+    for slice_idx in slices[trunc_data.shape[1]]:
+      pairs = trunc_data[:, slice_idx, :]
+      diffs = pairs[:, 1, :] - pairs[:, 0, :]
+      to_keep = np.nonzero(diffs.ravel())
```
bellet: This is a bit difficult to parse. Why do we need these slices? I am a bit lazy to check but maybe it should be made more clear even if it is less efficient (this is a small 1D dataset anyway so we don't care)

also maybe removing things that are very close to being the same (as opposed to exactly the same) would be more robust

metric_learn/sdml.py
bellet: and maybe print something similar when sklearn solver is used

metric_learn/constraints.py
wdevazelhes: Indeed, thanks

metric_learn/_util.py
wdevazelhes: That's right thanks

metric_learn/sdml.py
wdevazelhes: Agreed

metric_learn/sdml.py
wdevazelhes: Agreed

metric_learn/sdml.py
wdevazelhes: I agree, here is the new version I'll commit. When using scikit-learn's graphical lasso, we try it and if an error is returned or the result is not finite, we raise a warning that will be printed before the error (if there is an error), or before returning M (if there is no error but there are NaNs) : Tell me what you think 
```python
    if HAS_SKGGM:
      theta0 = pinvh(sigma0)
      M, _, _, _, _, _ = quic(emp_cov, lam=self.sparsity_param,
                              msg=self.verbose,
                              Theta0=theta0, Sigma0=sigma0)
    else:
      try:
        _, M = graphical_lasso(emp_cov, alpha=self.sparsity_param,
                               verbose=self.verbose,
                               cov_init=sigma0)
        error = None
      except FloatingPointError as e:
        error = e
      if not np.isfinite(M).all() or error is not None:
        msg = ("Scikit-learn's graphical lasso has failed to converge. "
               "Package skggm's graphical lasso can sometimes converge on "
               "non SPD cases where scikit-learn's graphical lasso fails to "
               "converge. Try to install skggm and rerun the algorithm. (see "
               "the README.md for the right version.)")
        warnings.warn(msg)
        if error is not None:
          raise(error)  
```

wdevazelhes: Actually, trying to come up with the right tests, I realized the following: 
in pathological diagonal cases (see above), scikit-learn's graphical lasso can return no error and  return a result that is not SPD (in the below example, a diagonal matrix with a negative coefficient on the diagonal). But then a NAN will appear when taking the `transformer_from_metric`. So maybe if using scikit-learn, I should additionally check that the result is indeed SPD and if not raise the error "SDML has failed to converge to an SPD matrix using scikit-learn's graphical lasso"
I don't know if skggm guarantees to return an SPD matrix, so maybe I should do the check there too ? (and then later if we dig deeper into that we could remove this if the test is not/[no more] needed)

Example (go in debug mode or put a print statement in SDML to see the result of the graphical lasso)



```python
    from metric_learn import SDML
    import numpy as np

    pairs = np.array([[[-10., 0.], [10., 0.]], [[0., 50.], [0., -60]]])
    y_pairs = [1, -1]

    sdml = SDML(use_cov=False, balance_param=100,verbose=True)

    diff = pairs[:, 0] - pairs[:, 1]
    emp_cov = np.identity(pairs.shape[2]) + 100 * (diff.T * y_pairs).dot(diff)

    print(emp_cov)

    sdml.fit(pairs, y_pairs)
    print(sdml.get_mahalanobis_matrix())
```

Returns: 
```
[[   40001.        0.]
 [       0. -1209999.]]
SDML will use scikit-learn's graphical lasso solver.
/home/will/Code/metric-learn/metric_learn/sdml.py:92: ConvergenceWarning: Warning, the input matrix of graphical lasso is not positive semi-definite (PSD). The algorithm may diverge, and lead to degenerate solutions. To prevent that, try to decrease the balance parameter `balance_param` and/or to set use_covariance=False.
  ConvergenceWarning)
[graphical_lasso] Iteration   0, cost  inf, dual gap 0.000e+00
/home/will/Code/metric-learn/metric_learn/_util.py:346: RuntimeWarning: invalid value encountered in sqrt
  return np.sqrt(metric)
[[2.4999375e-05           nan]
 [          nan           nan]]
```

And if we print the result of graphical lasso (note that it's the inverse of the initial matrix): 

```
[[ 2.49993750e-05  0.00000000e+00]
 [ 0.00000000e+00 -8.26446964e-07]]
```

bellet: How about this:
1- first check if the matrix given as input is PSD. If not, show the warning that the algorithm may diverge etc
2- regardless of the solver: if an error is thrown by the solver, or the solution is not a PSD matrix, show error and abort (as there is no result or the result does not correspond to a valid distance)
3- if we fall in case 2- and we use sklearn solver, say that the user could try to install skggm as its graphical lasso solver is more stable

wdevazelhes: I agree, I'll go for something like this: 

```python
    try:
      if HAS_SKGGM:
        theta0 = pinvh(sigma0)
        M, _, _, _, _, _ = quic(emp_cov, lam=self.sparsity_param,
                                msg=self.verbose,
                                Theta0=theta0, Sigma0=sigma0)
      else:
        _, M = graphical_lasso(emp_cov, alpha=self.sparsity_param,
                               verbose=self.verbose,
                               cov_init=sigma0)
      raised_error = None
      
    except Exception as e:
      raised_error = e
    w, V = np.linalg.eigh(M)
    if raised_error is not None or any(w < 0.):
      msg = ("There was a problem in SDML when using {}'s graphical "
             "lasso.").format("skggm" if HAS_SKGGM else "scikit-learn")
      if not HAS_SKGGM:
        skggm_advice = ("skggm's graphical lasso can sometimes converge "
                        "on non SPD cases where scikit-learn's graphical "
                        "lasso fails to converge. Try to install skggm and "
                        "rerun the algorithm. (See the README.md for the " 
                        "right version of skggm.)")
        msg += skggm_advice
      if raised_error is not None:
        msg += "The following error message was thrown: {}.".format(
          raised_error)
      raise(msg)
```

wdevazelhes: I'll add infinite values that can be returned by skggm as a failure case too

wdevazelhes: Done, in commit https://github.com/metric-learn/metric-learn/pull/162/commits/eb9571953a0001d54c2f4a1df47ab1519347f809

metric_learn/sdml.py
wdevazelhes: Agreed, done

metric_learn/sdml.py
wdevazelhes: In the heuristic if `min_eigval > 0` we would want to keep the matrix untouched (with just an epsilon added), (so I don't think with your solution it would be the case since we would substract the min eigval) but I agree something simpler would be better, so something like:

```python
min_eigval = w.min()
if min_eigval < 0:
  warnings.warn(...)
  w -= min_eigval   # we translate the eigenvalues to make them all positive

w += 1e-10  # we add a small offset to avoid definiteness problems
sigma0 = (V * w).dot(V.T)
```

wdevazelhes: done

README.rst
wdevazelhes: Thanks, I agree, I forgot to change that after making skggm optional

setup.py
wdevazelhes: I tried indeed but didn't manage to make it work. But good idea, skggm>=0.2.9 seems is better than nothing here

wdevazelhes:
> Have you verified that it fixes #138 and #154 ? If so please indicate this in the PR text

@bellet I just checked and it does not directly fixes them, since the two issues in fact are in one case where the matrix is not SPD and there are convergence problems. So I think to solve them we should change the balance parameter. I'll open a different PR for that

test/test_mahalanobis_mixin.py
wdevazelhes: I agree it's difficult to parse, I'll change the test copying/pasting for the quadruplets/pairs case.

wdevazelhes: Done, in commit https://github.com/metric-learn/metric-learn/pull/162/commits/31072d326620dce295669a45108f336d138459a0
What do you think ?

wdevazelhes:
Travis is failing because of a weird bug, I think related to pytest: when I run `pytest -k test_sdml_raises_warning_non_psd`, it passes, but it doesn't passes when I run `pytest -k metric_learn_test` (i.e. the whole test file). This is fixed in pytest 4, but pytest 4 raises errors on fixtures, so I would like for now to stay on pytest 3... I'll try to find something to make it work...

wdevazelhes:
This will be able to be fixed as soon as https://github.com/metric-learn/metric-learn/pull/186 is merged

wdevazelhes:
I think it should be ready to merge now ! :)

Issue #164, opened 2019-01-31
Ensure labels for positive/negative pairs are as expected


wdevazelhes:
We should ensure in `check_input` that when using Weakly Supervised algorithms, the users puts the right labels, which are +1 for positive pairs, and -1 for negative pairs.

PR #168, opened 2019-02-04
[MRG+1] Threshold for pairs learners

### LINKED ISSUE #131: Use a threshold for prediction for pairs

### LINKED ISSUE #165: Scoring functions
Besides the default `score` method of our estimators, we should properly test and document the fact that one can use [various metrics](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.metrics) to evaluate our supervised and weakly-supervised metric learners when using sklearn's tools such as `model_selection.GridSearchCV` and `model_selection.cross_val_score` through their `scoring` parameter.

This should be done after dealing with #131, which will introduce a predict method for weakly supervised pairwise metric learners.

wdevazelhes:
This PR fits a threshold for tuples learners to allow a predict (and scoring) on pairs

Fixes #131 
Fixes #165  

Finally, we decided that it would be good to have at least  a  minimal implementation of threshold calibration inside metric-learn, so that `_PairsClassifiers` can have a threshold hence a predict directly out of the box, without the need for a MetaEstimator. A MetaEstimator could however be used outside the algorithm for more precise threshold calibration (with cross-validation). 

The following features should be implemented: 
- [x] We should have two methods for `_PairsClassifiers`: `set_threshold()` and <strike>`calibrate_threshold(validation_set, method='max_tpr', args={'min_tnr': 0.1})`. `set_threshold` will set the threshold to a hard value, and `calibrate_threshold` will take a validation set and a method ('accuracy', 'f1'...) and will find the threshold which optimizes the metric in `method` on the validation set.</strike> We went for the same syntax that scikit-learn's PR.
- [ ] At fit time, we should either have a simple rule to set a threhold (for instance median of distances, or mean between positive pairs distances mean and negative pairs distances mean), or we should return `calibrate_threshold(trainset)`, and in this case also raise a warning at the end that says that the threshold has been fitted on the trainset, so we should check scikit-learn's utilities for calibration to have a calibration less prone to overfitting. Also in this case we could allow to put arguments in `fit` like `fit(pairs, y, threshold_method='max_tpr', threshold_args={'min_tnr': 0.1})
- [x] The following scores should be implemented: 
  - 'accuracy'
  - ‘f_beta’
  - ‘max_tpr’
  - 'max_tnr'
See scikit-learn's calibration PR (https://github.com/scikit-learn/scikit-learn/pull/10117) for more details, and the documentation of it here https://35753-843222-gh.circle-artifacts.com/0/doc/modules/calibration.html
- [x] For some estimators for which a natural threshold exist (like `ITML`: the mean between the lower threshold and the higher threshold), we should put this threshold

- [x] Decide what to do by default, rule of thumb scoring or calibration on trainset ? 

## Questions: 

- Should we do the same thing for QuadrupletsLearners ? (the natural threshold is 0, so I don't think it would make as much sense here, and users would maybe rather use the future meta estimator from scikit-learn's PR https://github.com/scikit-learn/scikit-learn/pull/10117), but since we will already have it implemented for pairs, and maybe for coherence, we could have it also for Quadruplets Learners

## TODO: 
- [x] Implement tests that check that we can use custom scores in cross val (cf #165). This needs to have a predict hence is related to this PR. 
- [x] Implement API tests (that the behaviour is as expected)
- [x] Implement numerical tests (on examples where we know the f_1 score etc
- [x] Implement the actual method (cf. features above)
- [x] Add this in the doc: also talk in the doc about the `CalibratedClassifierCV`
- [x] <strike>Think about which scoring make sense with quadruplets and what impact is has on the  code </strike>
- [x] <strike>Use/Adapt `CalibratedClassifierCV` or an equivalent for quadruplets</strike>
- [x] <strike> Maybe test CalibratedClassifierCV that it returns a coherent value (like all that have predict_proba = 0.8 have indeed 80% success) </strike> CalibratedClassifier's behaviour should be tested in scikit-learn: in metric learn we should just test the API (but on a small draft example I tested it for ITML and it worked more or less)
- [ ] Add a case (and a test case) where the best accuracy is when we need to reject all points (so threshold  = best score + 1). See if this applies too to other strategies


wdevazelhes:
I just added two commits: 

In the first commit, I just added a minimal implementation of threshold, which already fixes #165. (Since it gives a `predict` function to all pairs metric learners)

In the second commit, I also modified LSML to be able to `cross_val_score`. To do that I add an optional argument `y`. This is to allow using things like `cross_val_score(lsml, quadruplets, y, scoring='f1')` which uses scikit learn scoring function hence need an explicit `y`. I know that this adds an extra argument that will almost never be used (`y`), but since we deal well with the default case, and document it, I think it's worth it since it allows to use scikit-learn functions out of the box. (The alternative solution would be including `cross_val_score` in metric-learn's codebase specially for `LSML`, to fill y with one before doing the actual scoring)
(Note that there is no need to refactor other parts of the code because `LSML` was always used with a custom dataset that put the second argument at `np.ones(n_samples)` (see `build_quadruplets` in `test_utils.py`). Before it therefore put `weights=np.ones(n_samples)`, but now it puts `y=np.ones(n_samples)`, which is the same (and desired) behaviour)

Note that I also realized that the score was not computing the accuracy because in fact outputs were either -1 or 1, not 0 or 1, so I changed it, including the newly introduced `y`. Note that we use accuracy instead of `roc_auc` like it's the case for `_PairsClassifier`. We could use `roc_auc` as a default score, what do you think ? I don't know what would be the right default score.

metric_learn/base_metric.py
```diff
@@ -443,11 +444,22 @@ def score(self, quadruplets, y=None):
...
+      first two points to be closer than the last two points in each
+      quadruplet.
 
     Returns
     -------
     score : float
       The quadruplets score.
     """
-    return -np.mean(self.predict(quadruplets))
+    quadruplets = check_input(quadruplets, y, type_of_inputs='tuples',
```
wdevazelhes: note that here quadruplets will be checked twice (once here, once in predict). This is because when I do `y = np.ones(quadruplets.shape[0])`, I want to be sure that I can do `quadruplets.shape[0]`, and otherwise an error message would be returned before (by the `check_input` method). I think it's fine isn't it ? I don't see any other solution to do so. Note that I also check at the same time `y` because I like the fact that `column_or_1d` check will be called on y (since it is not done in `accuracy_score`).

wdevazelhes:
I'll solve errors in tests, but one of them is that 'roc_auc' is not defined if there is only one class (fails for quadruplet I think because the true labels are `np.ones(n_samples)`). This makes sense, 'roc_auc' should indeed not be defined for quadruplets, since a task to classify a quadruplet can be arbitrarily defined with a label +1 and some ordering of points in quadruplets, or  a label of -1 and the opposite ordering. Therefore there is no notion of "positive" or "negative" like in usual classification I guess. I think still this code is useful to use `cross_val_score(..., 'accuracy')`, but I'll double check every score since I guess some other useful scoring for classification will be problematic too for quadruplets.

wdevazelhes:
I was about to make a test for being compatible with scikit-learn's `CalibratedClassifierCV`, and realized I also need to check whether it would do the right thing for quadruplets.
Even if it doesn't, I think some probability calibration should still be useful for quadruplets, like a function that would return a proba between 0 and 1 such that quadruplets which have a proba of 0.2 have 80% chance to be in the wrong order and 20% chance to be in the right order. This would also mean that swaping the pairs in the quadruplets should return the opposite proba. I still need to understand better `CalibratedClassifierCV` to see how it could be used/adapted for quadruplets.

metric_learn/base_metric.py
```diff
@@ -296,6 +296,7 @@ def get_mahalanobis_matrix(self):
 
 class _PairsClassifierMixin(BaseMetricLearner):
 
+  classes_ = [-1, 1]
```
wdevazelhes: We need estimators to have a `classes_` attribute for `CalibratedClassifierCV` to work. Typically, scikit-learn estimators deduce it at fit time, but in our case we will always have these ones (see also issue #164). That's why I put it as a class attribute, to be more explicit.

Classes have to be in this order (`[-1, 1]`)(i.e. sorted), otherwise there is a bug: because here https://github.com/scikit-learn/scikit-learn/blob/1a850eb5b601f3bf0f88a43090f83c51b3d8c593/sklearn/calibration.py#L312-L313:
`self.label_encoder_.classes_` is a sorted list of classes (so it'll be `[-1, 1]` in every cases). So when it transforms `self.base_estimator.classes_` it will return `[0, 1]` if our estimator has `[-1, 1]` as classes, and `[1, 0]` if it has `[1, -1]`. In the latter case, it will return `IndexError` because here (https://github.com/scikit-learn/scikit-learn/blob/1a850eb5b601f3bf0f88a43090f83c51b3d8c593/sklearn/calibration.py#L357) it will try to reach the column n°1 (and not 0) of `Y` (which does not exist).

Also, we have to put -1 and 1 in this order so that for `label_binarizer` (which is called by `CalibratedClassifierCV`), -1 (first element of `[-1, 1]`) will be considered as the negative label, and 1 will be considered as the positive label. (That's how `label_binarizer` work, see the example below)

Example: (warning: pos_label and neg_label are not to say what input classes are pos or neg, as one could think, but rather how we want them to appear in the output)
```python
In [1]: from sklearn.preprocessing import label_binarize 
   ...: print(label_binarize([-1, 1, -1], [-1, 1], pos_label=2019, neg_label=2018))                                                                                                                                                                                                        
[[2018]
 [2019]
 [2018]]
```
And:

```python
In [1]: from sklearn.preprocessing import label_binarize 
   ...: print(label_binarize([-1, 1, -1], [1, -1], pos_label=2019, neg_label=2018))                                                                                                                                               
[[2019]
 [2018]
 [2019]]
```

metric_learn/base_metric.py
```diff
@@ -458,9 +460,14 @@ def score(self, quadruplets, y=None):
...
     """
-    quadruplets = check_input(quadruplets, y, type_of_inputs='tuples',
-                              preprocessor=self.preprocessor_,
-                              estimator=self, tuple_size=self._tuple_size)
+    checked_input = check_input(quadruplets, y, type_of_inputs='tuples',
+                                preprocessor=self.preprocessor_,
+                                estimator=self, tuple_size=self._tuple_size)
+    # checked_input will be of the form `(checked_quadruplets, checked_y)` if
+    # `y` is not None, or just `checked_quadruplets` if `y` is None
+    quadruplets = checked_input if y is None else checked_input[0]
```
wdevazelhes: I find this a bit ugly, but I couldn't find another way to do it. Maybe refactor `check_input` with two functions like `check_X_y` and `check_array` ? Or we can just keep it as it is for now

test/test_quadruplets_classifiers.py
```diff
@@ -25,35 +25,11 @@ def test_predict_only_one_or_minus_one(estimator, build_dataset,
   assert np.isin(predictions, [-1, 1]).all()
 
 
-@pytest.mark.parametrize('with_preprocessor', [True, False])
```
wdevazelhes: I removed this test because it makes no sense for quadruplets since the threshold should be always 0

metric_learn/base_metric.py
```diff
@@ -296,6 +296,7 @@ def get_mahalanobis_matrix(self):
 
 class _PairsClassifierMixin(BaseMetricLearner):
 
+  classes_ = [-1, 1]
   _tuple_size = 2  # number of points in a tuple, 2 for pairs
```
bellet: shouldn't we document the attribute `threshold_` here in the general class rather than in the downstream classes?

wdevazelhes:
> 1/ I think we have a problem with `decision_function`: this is interpreted as `the larger, the more likely it corresponds to the positive class`. However as it is a distance it is the other way around for us, which I think will make the computation of AUC wrong. I think we should return minus the distance instead. To return the actual distances, we already have score_pairs.

Yes, absolutely, I just changed it

metric_learn/base_metric.py
wdevazelhes: I guess I should document it for both abstract and downstream classes, but I don't think I can document it only here: indeed, we can inherit docstrings, but I don't think we can inherit docstrings paragraph-wise (e.g. `_PairsClassifierMixin` has an attribute `threshold_` that we want to transmit to `MMC`. But the problem is that `MMC` has also a title, and also other attributes like `transformer_`). I didn't find an option to do this in sphinx and this impossibility make some sense because if we write two docstrings paragraphs (below the title), would the result in the downstream class be the concatenation of both descriptions ? It could give weird results (e.g. : "This is the abstract class", "This is is the MMC class", would give: "This is the abstract class. This is the MMC class") However there seem to exist some packages that could allow to do something like this. We could use them but maybe later on in the development if things get too replicated. https://github.com/Chilipp/docrep https://github.com/meowklaski/custom_inherit
Then there's also the solution not to care about inheriting the docstrings, but I think it's better for a user to know what are all the attributes in his object without needing to click on the doc of all mother classes (that's what is done in scikit-learn for instance I think)
Finally there's also another option which is to declare attributes as properties so that we can have a more granular inheritance, but I think this adds to much complexity just for docstrings

wdevazelhes: (done: I just added them to the base PairsClassifier class too)

wdevazelhes:
> 2/ For quadruplet metric learner: I think the only score that makes sense is the accuracy, because each quadruplet contains both a positive and a negative pair. So while the trick of adding an argument `y` works to call the accuracy score function of sklearn, an easy way to avoid that users try to use other scoring functions would be not to add the `y` and simply compute the accuracy ourselves. This way it will raise an error when trying to use other sklearn scoring functions. What do you think? I am not sure how we could raise a specific error to explain that only accuracy makes sense for quadruplets. In any case, we could deal with the quadruplet thing in a separate PR (see also point 3/ below) and keep this PR only about pairwise metric learners.

I agree, however I think functions that use `predict_proba` could still make sense for quadruplets ? (like scikit-learn's brier loss or neg log loss) (assuming that we have a predict_proba function for quadruplets)? Indeed they don't have a notion of positive or negative class. But it seems more important to prevent using a lot of wrong common scoring functions rather than allowing to use two rare scoring functions.

Also, all scikit-learn estimators have a `y` argument even if it is not used, for compatibility with pipelines. 
 But however I don't think pipelines with `LSML` would be wanted by users. 

If we wanted to still implement a way to raise an error if the user used scikit-learn's scoring function, we could maybe do something with the `inspect` module (inspecting which function has called `predict` and raise an error if it is a `_score` that is not the accuracy or brier or neg log loss or something like that). 

But to sum up this would mean doing something hacky for in the end allowing functionnalities that are not mainstream (using the probabilistic losses, and calling the scoring with `cross_val_score(scoring= 'accuracy')` instead of `cross_val_score(scoring=None)`), so you're right for now I'll go with the simpler thing (removing the y), and indeed, doing so in another PR seems better

wdevazelhes:
> 3/ For quadruplet metric learners but also the metric transformers, in terms of use case it would still be useful to allow people to easily get a pair classifier out of the learned distance. For this we could for instance introduce a new " instance of pair metric learner which would simply take a predefined metric as input instead of fitting on data (users can then easily set/calibrate the threshold using existing methods). There may be a better way to do this. We should discuss this separately (will create an issue).

I agree, let's deal with this in the new issue

wdevazelhes:
> can't we rather keep the threshold positive (i.e., which works with actual distances and not negative distances), and change the predict accordingly? It feels weird to have a negative threshold

Yes I agree, I was hesitating between both, but I guess a negative threshold is weird indeed

metric_learn/base_metric.py
```diff
@@ -333,7 +332,7 @@ def predict(self, pairs):
       The predicted learned metric value between samples in every pair.
     """
     check_is_fitted(self, ['threshold_', 'transformer_'])
-    return 2 * (self.decision_function(pairs) > - self.threshold_) - 1
+    return 2 * (self.decision_function(pairs) >= - self.threshold_) - 1
```
wdevazelhes: This is the rule used in scikit-learn for thresholds, see for instance here: https://github.com/scikit-learn/scikit-learn/blob/62d205980446a1abc1065f4332fd74eee57fcf73/sklearn/metrics/ranking.py#L377-L381

metric_learn/base_metric.py
```diff
@@ -396,6 +395,150 @@ def _set_default_threshold(self, pairs, y):
...
+      y_ordered = y_valid[scores_sorted_idces]
+      # finds the threshold that maximizes the accuracy:
+      cum_tp = stable_cumsum(y_ordered == 1)  # cumulative number of true
+      # positives
+      cum_tn_inverted = stable_cumsum(y_ordered[::-1] == -1)
+      cum_tn = np.concatenate([[0], cum_tn_inverted[:-1]])[::-1]
+      cum_accuracy = (cum_tp + cum_tn) / n_samples
+      max_i = np.argmax(cum_accuracy)
+      # note: we want a positive threshold (distance), so we take - threshold
+      self.threshold_ = - scores_sorted[max_i]
```
wdevazelhes: For the threshold, I use a threshold which is taken among the values of the train set scores (i.e. some/one of the train set scores value(s) on the trainset will be equal to the threshold). But I thought it could also make sense to set a threshold that is  the mean between the value of the lowest value of the train set scores that we want to keep and the highest value of the train set scores that we want to reject, doing sort of a margin for classification. Would that be a better strategy ? Is there other srategies that are better ?

metric_learn/base_metric.py
```diff
@@ -396,6 +395,150 @@ def _set_default_threshold(self, pairs, y):
...
+          y_valid, self.decision_function(pairs_valid), pos_label=1)
+      with np.errstate(divide='ignore', invalid='ignore'):
+        f_beta = ((1 + beta**2) * (precision * recall) /
+                  (beta**2 * precision + recall))
+      f_beta[np.isnan(f_beta)] = 0.
+      imax = np.argmax(f_beta)
+      self.threshold_ = - thresholds[imax]
+      return self
+
+    fpr, tpr, thresholds = roc_curve(y_valid,
```
wdevazelhes: Note that there is a particular corner case that is not explicitely coded here but comes from scikit-learn: we should decide if that's what we want or not: 
In `roc_curve`, an additional point is added in (0, 0) (0 false positives and 0 true positives (so 0 positives)) if needed (i.e. if there is not already such a point), with an artificial threshold which is thresholds[0] + 1 (see  https://github.com/scikit-learn/scikit-learn/blob/62d205980446a1abc1065f4332fd74eee57fcf73/sklearn/metrics/ranking.py#L637-L642). (this points represent a case where we want to reject all samples, the +1 makes sense since it will reject all samples from the training set)
Is that what we want ? It kind of makes sense since it works if we want to reject everything but the +1 seems a bit arbitrary

metric_learn/base_metric.py
```diff
@@ -396,6 +395,150 @@ def _set_default_threshold(self, pairs, y):
...
+      cum_accuracy = (cum_tp + cum_tn) / n_samples
+      max_i = np.argmax(cum_accuracy)
+      # note: we want a positive threshold (distance), so we take - threshold
+      self.threshold_ = - scores_sorted[max_i]
+      return self
+
+    if strategy == 'f_beta':
+      precision, recall, thresholds = precision_recall_curve(
+          y_valid, self.decision_function(pairs_valid), pos_label=1)
+      with np.errstate(divide='ignore', invalid='ignore'):
```
wdevazelhes: We ignore the warnings here, in the same taste as https://github.com/scikit-learn/scikit-learn/blob/62d205980446a1abc1065f4332fd74eee57fcf73/sklearn/metrics/classification.py#L1284

test/test_pairs_classifiers.py
```diff
@@ -85,3 +127,194 @@ def test_set_default_threshold_toy_example():
...
+  n_samples = 50
+  pairs, y = rng.randn(n_samples, 2, 5), rng.choice([-1, 1], size=n_samples)
+  pairs_learner = IdentityPairsClassifier()
+  pairs_learner.fit(pairs, y)
+  pairs_learner.calibrate_threshold(pairs, y, **kwargs)
+  best_score = scoring(y, pairs_learner.predict(pairs))
+  scores = []
+  i = 0
+  predicted_scores = pairs_learner.decision_function(pairs)
+  predicted_scores = np.hstack([[np.min(predicted_scores) - 1],
```
wdevazelhes: We also test that it works for thresholds higher or lower than the extreme predicted scores.

metric_learn/base_metric.py
```diff
@@ -396,6 +395,150 @@ def _set_default_threshold(self, pairs, y):
...
+        f_beta = ((1 + beta**2) * (precision * recall) /
+                  (beta**2 * precision + recall))
+      f_beta[np.isnan(f_beta)] = 0.
```
wdevazelhes: We need to set nans to zero otherwise they will be considered higher than the others (also discussed in https://github.com/scikit-learn/scikit-learn/pull/10117/files#r262115773)

metric_learn/base_metric.py
```diff
@@ -396,6 +395,150 @@ def _set_default_threshold(self, pairs, y):
...
+      cum_tn_inverted = stable_cumsum(y_ordered[::-1] == -1)
+      cum_tn = np.concatenate([[0], cum_tn_inverted[:-1]])[::-1]
+      cum_accuracy = (cum_tp + cum_tn) / n_samples
+      max_i = np.argmax(cum_accuracy)
+      # note: we want a positive threshold (distance), so we take - threshold
+      self.threshold_ = - scores_sorted[max_i]
+      return self
+
+    if strategy == 'f_beta':
+      precision, recall, thresholds = precision_recall_curve(
```
wdevazelhes: Note that `precision_recall_curve` adds a virtual point, but it is harmless for the next computations (see https://github.com/scikit-learn/scikit-learn/pull/10117/files#r262115773)

metric_learn/base_metric.py
wdevazelhes: Note also that some points can be deleted in this curve (see these lines of code, and PR #5237). But it will not change the result as it just removes suboptimal points (and keeps optimal points i.e. points with the highest tpr for the same tnr and points with the highest tnr for the same tpr). We could have avoided this deletion by setting `drop_intermediate=True` but we don't need to do that since the deletion is harmless.

wdevazelhes:
Hi, I just pushed a version of the `calibrate_threshold` method, let me know what you think. I reused most of the code from https://github.com/scikit-learn/scikit-learn/pull/10117

Finally, I also need to add some documentation about the threshold
(see the TODO in the PR message)

Also, there is still to decide what we do by default ? The "rule of thumb" threshold (mean between two mean of distances), or accuracy ? or another score ? I guess accuracy is the more natural, but maybe sometimes it can be a little prone to overfitting (since when I say by default it means we set on the trainset) ? But this is maybe not so big (we could verify it on a quick benchmark on some datasets ?)

wdevazelhes:
I just added some documentation for the calibration, we just need to decide what to do for the default threshold (the heuristic of the mean of means, or a calibration on the trainset ? ) And it should be ready to merge

metric_learn/base_metric.py
bellet: In principle what you suggest is a good idea. Not sure what they do in the sklearn PR?

metric_learn/base_metric.py
bellet: Maybe `- self.decision_function(pairs) <= self.threshold_` is more intuitive?

metric_learn/base_metric.py
bellet: I would say that it is fine this way? It does not seem very important anyway so going with the sklearn approach is fine

metric_learn/base_metric.py
bellet: maybe it is good to add a comment in the code to explain

metric_learn/base_metric.py
bellet: again, probably good to mention it in a comment in the code

metric_learn/base_metric.py
bellet: again?

metric_learn/base_metric.py
```diff
@@ -317,13 +331,16 @@ def predict(self, pairs):
...
+    return 2 * (self.decision_function(pairs) >= - self.threshold_) - 1
 
   def decision_function(self, pairs):
-    """Returns the learned metric between input pairs.
+    """Returns the decision function used to classify the pairs.
 
-    Returns the learned metric value between samples in every pair. It should
-    ideally be low for similar samples and high for dissimilar samples.
+    Returns the opposite of the learned metric value between samples in every
+    pair. Hence it should ideally be low for dissimilar samples and high for
```
bellet: maybe add "to be consistent with scikit-learn convention"?

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+    See more in the :ref:`User Guide <calibration>`.
+    """
+    similar_threshold = np.mean(self.score_pairs(
+        pairs[(y == 1).ravel()]))
+    dissimilar_threshold = np.mean(self.score_pairs(
+        pairs[(y == -1).ravel()]))
+    self.threshold_ = np.mean([similar_threshold, dissimilar_threshold])
+
+  def set_threshold(self, threshold):
+    """Sets the threshold of the metric learner to the given value `threshold
```
bellet: missing `

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+
+    See more in the :ref:`User Guide <calibration>`.
+
+    Parameters
+    ----------
+    threshold : float
+      The threshold value we want to set. It's a distance metric with
```
bellet: there seems to be some typo here

maybe say
It is the value to which the predicted distance for test pairs will be compared.

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
     """
     return roc_auc_score(y, self.decision_function(pairs))
 
+  def set_default_threshold(self, pairs, y):
```
bellet: to remove if indeed we choose the accuracy-calibrated threshold as default

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+    -------
+    self : `_PairsClassifier`
+      The pairs classifier with the new threshold set.
+    """
+    self.threshold_ = threshold
+    return self
+
+  def calibrate_threshold(self, pairs_valid, y_valid, strategy='accuracy',
+                          threshold=None, beta=None):
+    """Decision threshold calibration for binary classification
```
bellet: pairwise binary classification instead of binary classification

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+    `predict`. The methods for picking cutoff points make use of traditional
+    binary classification evaluation statistics such as the true positive and
+    true negative rates and F-scores. The threshold will be found to maximize
+    the chosen score on the validation set `(pairs_valid, y_valid)`.
+
+    See more in the :ref:`User Guide <calibration>`.
+
+    Parameters
+    ----------
+    strategy : str, optional (default='roc')
```
bellet: 'roc' is not a valid option

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+      The strategy to use for choosing the cutoff point
```
bellet: better to always use "threshold"

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+
+      'accuracy'
+          selects a decision threshold that maximizes the accuracy
+      'f_beta'
+          selects a decision threshold that maximizes the f_beta score
+      'max_tpr'
+          selects the point that yields the highest true positive rate with
```
bellet: a decision threshold

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+    Returns
+    -------
+    self : `_PairsClassifier`
+      The pairs classifier with the new threshold set.
+    """
+    self.threshold_ = threshold
+    return self
+
+  def calibrate_threshold(self, pairs_valid, y_valid, strategy='accuracy',
+                          threshold=None, beta=None):
```
bellet: threshold is not a good name! we already use threshold for the actual threshold

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+      'accuracy'
+          selects a decision threshold that maximizes the accuracy
+      'f_beta'
+          selects a decision threshold that maximizes the f_beta score
+      'max_tpr'
+          selects the point that yields the highest true positive rate with
+          true negative rate at least equal to the value of the parameter
+          threshold
+      'max_tnr'
+          selects the point that yields the highest true negative rate with
```

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+      The strategy to use for choosing the cutoff point
+
+      'accuracy'
+          selects a decision threshold that maximizes the accuracy
+      'f_beta'
+          selects a decision threshold that maximizes the f_beta score
+      'max_tpr'
+          selects the point that yields the highest true positive rate with
+          true negative rate at least equal to the value of the parameter
+          threshold
```
bellet: change name

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+      'max_tnr'
+          selects the point that yields the highest true negative rate with
+          true positive rate at least equal to the value of the parameter
+          threshold
```
bellet: change name

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+
+    Parameters
+    ----------
+    strategy : str, optional (default='roc')
+      The strategy to use for choosing the cutoff point
+
+      'accuracy'
+          selects a decision threshold that maximizes the accuracy
+      'f_beta'
+          selects a decision threshold that maximizes the f_beta score
```
bellet: maybe add that beta is given by the value of parameter beta?

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+          threshold
+      'max_tnr'
+          selects the point that yields the highest true negative rate with
+          true positive rate at least equal to the value of the parameter
+          threshold
+
+    beta : float in [0, 1], optional (default=None)
+      beta value to be used in case strategy == 'f_beta'
+
+    threshold : float in [0, 1] or None, (default=None)
```
bellet: change name

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+      In case strategy is 'max_tpr' or 'max_tnr' this parameter must be set
+      to specify the threshold for the true negative rate or true positive
```
bellet: value instead of threshold

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+      rate respectively that needs to be achieved
+
+    pairs_valid : array-like, shape=(n_pairs_valid, 2, n_features)
+      The validation set of pairs to use to set the threshold.
+
+    y_valid : array-like, shape=(n_pairs_valid,)
+      The labels of the pairs of the validation set to use to set the
```
bellet: these labels must be -1 and 1

metric_learn/base_metric.py
bellet: we could use 1 as default value for beta, as f1 score is very standard

metric_learn/base_metric.py
```diff
@@ -369,6 +386,168 @@ def score(self, pairs, y):
...
+
+    if strategy == 'f_beta':
+      precision, recall, thresholds = precision_recall_curve(
+          y_valid, self.decision_function(pairs_valid), pos_label=1)
+      with np.errstate(divide='ignore', invalid='ignore'):
+        f_beta = ((1 + beta**2) * (precision * recall) /
+                  (beta**2 * precision + recall))
+      f_beta[np.isnan(f_beta)] = 0.
+      imax = np.argmax(f_beta)
+      self.threshold_ = - thresholds[imax]
```
bellet: could add the note on the minus as for other scores

metric_learn/base_metric.py
bellet: I guess that the same reasoning could be applied when setting the threshold for other calibration strategies (other than accuracy)?

metric_learn/itml.py
```diff
@@ -148,11 +148,23 @@ class ITML(_BaseITML, _PairsClassifierMixin):
...
       The linear transformation ``L`` deduced from the learned Mahalanobis
       metric (See function `transformer_from_metric`.)
+
+  threshold_ : `float`
+      If the distance metric between two points is lower than this threshold,
+      points will be classified as similar, otherwise they will be
+      classified as dissimilar.
+
+  classes_ : `list`
+      The possible labels of the pairs `LSML` can fit on. `classes_ = [-1, 1]`,
```
bellet: ITML not LSML

metric_learn/itml.py
```diff
@@ -148,11 +148,23 @@ class ITML(_BaseITML, _PairsClassifierMixin):
...
+      where -1 means points in a pair are dissimilar (negative label), and 1
+      means they are similar (positive label).
   """
 
   def fit(self, pairs, y, bounds=None):
     """Learn the ITML model.
 
+    The default threshold will be set to the mean of the bounds.
```
bellet: to update, and make sure it says the same for all pairwise metric learners

metric_learn/itml.py
```diff
@@ -176,7 +188,9 @@ def fit(self, pairs, y, bounds=None):
     self : object
         Returns the instance.
     """
-    return self._fit(pairs, y, bounds=bounds)
+    self._fit(pairs, y, bounds=bounds)
+    self.threshold_ = np.mean(self.bounds_)
```
bellet: to remove

test/test_mahalanobis_mixin.py
```diff
@@ -25,7 +26,7 @@ def test_score_pairs_pairwise(estimator, build_dataset):
   X = X[:n_samples]
   model = clone(estimator)
   set_random_state(model)
-  model.fit(input_data, labels)
+  model.fit(*make_args_inc_quadruplets(estimator, input_data, labels))
```
bellet: this quite ugly ;-) if such a function is really handy, maybe try to at least find a shorter name

wdevazelhes:
Thanks for the review @bellet !

>     * It would be nice to have a parameter in `fit` to set the calibration strategy (with accuracy as default)
I agree, I was thinking to go for something like `threshold_params=None`, (None will call accuracy by default) that could be say `{'strategy': 'accuracy'}` or `{'strategy': 'max_tpr', 'threshold':0.1}` for instance, what do you think ?
This way first we don't have a lot of additional arguments in fit, and we don't have to change the code if we change the API of `calibrate_threshold`

metric_learn/base_metric.py
wdevazelhes: Thanks, done

metric_learn/base_metric.py
wdevazelhes: Thanks, done

metric_learn/base_metric.py
wdevazelhes: Agreed, done

metric_learn/base_metric.py
wdevazelhes: Agreed, done

metric_learn/base_metric.py
wdevazelhes: Agreed, what about `min_rate` ?

metric_learn/base_metric.py
wdevazelhes: done

metric_learn/base_metric.py
wdevazelhes: done

metric_learn/base_metric.py
wdevazelhes: done

metric_learn/base_metric.py
wdevazelhes: Agreed, done

metric_learn/base_metric.py
wdevazelhes: done

metric_learn/base_metric.py
wdevazelhes: Agreed, even `minimal value` would be good no ?

metric_learn/base_metric.py
wdevazelhes: Agreed, done

metric_learn/base_metric.py
wdevazelhes: Agreed, that's also what is done in scikit-learn  for `precision_recall_fscore_support`
done

metric_learn/base_metric.py
wdevazelhes: Agreed, done

wdevazelhes:
>I guess that the same reasoning could be applied when setting the threshold for other calibration strategies (other than accuracy)?

You mean the cumsum ? It's already used by `precision_recall_curve`, and `roc_curve`

metric_learn/itml.py
wdevazelhes: Thanks, done

metric_learn/itml.py
wdevazelhes: Do you mean we should use the same default threshold for ITML as for other pairwise metric learners ?  I agree there would be more coherence then
But then maybe we could still allow a method like `set_threshold_bounds` only for ITML that sets the threshold to the mean of the bounds ?

metric_learn/itml.py
wdevazelhes: I guess if the above is yes you mean remove this and replace it by `calibrate_threshold` like in all pairwise metric learners ?

doc/weakly_supervised.rst
```diff
@@ -148,8 +148,44 @@ tuples you're working with (pairs, triplets...). See the docstring of the
...
+samples, we need to set a distance threshold, so that points closer (in the
+learned space) than this threshold are predicted as similar, and points further
+away are predicted as dissimilar. Several methods are possible for this
+thresholding.
+
+- **default**: Unless explicitely stated in the `fit` method documentation
+  of the estimator, the threshold is set with the method
+  `set_default_threshold` on the trainset.
+
+- **manual**: calling `set_threshold`, the user can
```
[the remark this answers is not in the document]
wdevazelhes: Agreed, done

doc/weakly_supervised.rst
```diff
@@ -148,8 +148,44 @@ tuples you're working with (pairs, triplets...). See the docstring of the
...
+  manually set the threshold to a particular value.
+
+- **calibrated**: calling `calibrate_threshold`, the user can
```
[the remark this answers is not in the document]
wdevazelhes: done

doc/weakly_supervised.rst
```diff
@@ -148,8 +148,44 @@ tuples you're working with (pairs, triplets...). See the docstring of the
...
+  calibrate the threshold to achieve a particular score on a validation set,
+  the score being among the classical scores for classification (accuracy, f1
+  score...).
+
+
+See also: `sklearn.calibration`.
```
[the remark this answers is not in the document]
wdevazelhes: I agree

metric_learn/base_metric.py
```diff
@@ -295,7 +295,21 @@ def get_mahalanobis_matrix(self):
...
+  """
+  Attributes
+  ----------
+  threshold_ : `float`
+      If the distance metric between two points is lower than this threshold,
+      points will be classified as similar, otherwise they will be
+      classified as dissimilar.
+
+  classes_ : `list`
+      The possible labels of the pairs `MMC` can fit on. `classes_ = [-1, 1]`,
```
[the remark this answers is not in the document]
wdevazelhes: Oups, thanks, done

wdevazelhes:
> Do we need the classes_ attribute in this PR? Otherwise it could be saved for #173

No indeed, I'll put it in the PR fixing #173

metric_learn/base_metric.py
wdevazelhes: Agreed, done

metric_learn/base_metric.py
wdevazelhes: Thanks, done

metric_learn/base_metric.py
wdevazelhes: Thanks, yes I got a bit confused on this one :p

test/test_mahalanobis_mixin.py
wdevazelhes: I think such a function is needed because we don't have y for quadruplets learners
There might be a way to change this like making our `build_dataset` function return some "dataset" object that would be `(a, b)` (like `(X, y), (pairs, y_pairs)`...) for most estimators, but `(quadruplets,)` for quadruplets. Then we could pass it like this: `fit(*dataset)`. But then it would be problematic say if we want to split the dataset: `train_test_split(dataset)` would return either 4 or 2 objects, and it'll become hard to separate them... So I think we should deal with forming the appropriate object just before using them with `fit`, to have more freedom to control the objects before...
I'm not 100% sure about this, but I think it's OK to have this function for now, and anyway it shouldn't be too difficult to change later if we find a better solution

And I agree, the name is really bad :p I'll change it, what about `remove_y_quadruplets` ?

metric_learn/base_metric.py
wdevazelhes: I agree, done

metric_learn/base_metric.py
wdevazelhes: Yes, a point is added in two different scikit-learn codes, but it's both times harmless

metric_learn/base_metric.py
wdevazelhes: Agreed, done

metric_learn/base_metric.py
wdevazelhes: Agreed, done

metric_learn/base_metric.py
wdevazelhes: Yes, I would use this strategy for all strategies

I was about to implement it but then I wondered: what would we do in case the best threshold is at an extreme point (ex: we want to take all points)? Would we say that the optimal threshold is the score of this point (i.e. the worst score) ? Or the score of this point - 1 ? I know this case would rarely happen, but we should need to decide what to do... I think we could go for just the score of the point, it seems less arbitrary...

Note: scikit-learn can use the +1 strategy see https://github.com/scikit-learn/scikit-learn/blob/62d205980446a1abc1065f4332fd74eee57fcf73/sklearn/metrics/ranking.py#L642 (but here it's more justified, it's because we want to have a case where we reject all points. Since the equality is tested with the threshold when classifying, it's normal to have something higher than all scores)
By the way I need to deal with this case: I've added it to the TODO

wdevazelhes:
@bellet Thanks for your review, I addressed all your comments


Before merging, there just remains the following:
- Before implementing the mean of the smallest score to accept and the higher score to reject we need to decide what to do in the case where we want to accept all points (see comment https://github.com/metric-learn/metric-learn/pull/168#discussion_r267818701)
- We need to choose what to do for ITML (threshold with the mean of the bounds or default calibrate threshold)
- We need to add for accuracy (and others) the extreme point case (see the TODO)

bellet:
> Thanks for the review @bellet !
> 
> > ```
> > * It would be nice to have a parameter in `fit` to set the calibration strategy (with accuracy as default)
> > ```
> 
> I agree, I was thinking to go for something like `threshold_params=None`, (None will call accuracy by default) that could be say `{'strategy': 'accuracy'}` or `{'strategy': 'max_tpr', 'threshold':0.1}` for instance, what do you think ?
> This way first we don't have a lot of additional arguments in fit, and we don't have to change the code if we change the API of `calibrate_threshold`

Yes this is a good idea. I would maybe name this parameter `calibration_params`

metric_learn/base_metric.py
bellet: when true negative rate is specified it corresponds the max value desired
so maybe simply `rate`?

metric_learn/itml.py
```diff
@@ -170,13 +178,21 @@ def fit(self, pairs, y, bounds=None):
...
+        given, `calibrate_threshold` will use the default parameters.
 
     Returns
     -------
     self : object
         Returns the instance.
     """
-    return self._fit(pairs, y, bounds=bounds)
+    self._fit(pairs, y, bounds=bounds)
+    self.calibrate_threshold(pairs, y, **(threshold_params if
```
bellet: For all algorithms we should test that `threshold_params` has allowed values at the beginning of `fit` otherwise we will fit for nothing (which may take some time) if bad params have been given...

metric_learn/base_metric.py
bellet: Or `desired_rate`

metric_learn/base_metric.py
bellet: I don't think so, as discussed in other comment?

metric_learn/base_metric.py
```diff
@@ -369,6 +381,162 @@ def score(self, pairs, y):
...
+      y_ordered = y_valid[scores_sorted_idces]
+      # finds the threshold that maximizes the accuracy:
+      cum_tp = stable_cumsum(y_ordered == 1)  # cumulative number of true
+      # positives
+      cum_tn_inverted = stable_cumsum(y_ordered[::-1] == -1)
+      cum_tn = np.concatenate([[0], cum_tn_inverted[:-1]])[::-1]
+      cum_accuracy = (cum_tp + cum_tn) / n_samples
+      imax = np.argmax(cum_accuracy)
+      # note: we want a positive threshold (distance), so we take - threshold
+      self.threshold_ = - scores_sorted[imax]
```
bellet: shouldn't the threshold be positive?

test/test_mahalanobis_mixin.py
bellet: Yes, renaming is gine

bellet:
> > I guess that the same reasoning could be applied when setting the threshold for other calibration strategies (other than accuracy)?
> 
> You mean the cumsum ? It's already used by `precision_recall_curve`, and `roc_curve`

You can ignore this, you replied in the proper thread, not sure why it appeared out of context

bellet:
Sorry for the messy review (several tabs opened).

About your questions:
- Let's do the same for ITML as for all other algorithms for consistency
- For the extreme point question I do not think it matters much. You may just use +1 / -1 to give a safety margin. This is not an interesting case anyway!

metric_learn/base_metric.py
bellet: nevermind, got confused again haha (this is why the comment is useful)

bellet: maybe the comment can be a bit more explicit: "we are working with negative distances but we want the threshold to be with respect to the actual distances so we take minus sign" or something like that

metric_learn/base_metric.py
wdevazelhes: Why is that? 

When `max_tnr` is specified, we want the max `tnr` for a minimum `tpr` no ?
(and if `max_tpr` is specified we want the max `tpr` for a minimum `tnr`)

Because I think that's what we do currently:
- In the doc we have: 
```
      'max_tpr'
          Selects a decision threshold that yields the highest true positive
          rate with true negative rate at least equal to the value of the
          parameter `min_rate`.
      'max_tnr'
          Selects a decision threshold that yields the highest true negative
          rate with true positive rate at least equal to the value of the
          parameter `min_rate`.
```

- And in the code we have: 
```python
    if strategy == 'max_tpr':
      indices = np.where(1 - fpr >= min_rate)[0]
      max_tpr_index = np.argmax(tpr[indices])
      # note: we want a positive threshold (distance), so we take - threshold
      self.threshold_ = - thresholds[indices[max_tpr_index]]

    if strategy == 'max_tnr':
      indices = np.where(tpr >= min_rate)[0]
      max_tnr_index = np.argmax(1 - fpr[indices])
      # note: we want a positive threshold (distance), so we take - threshold
      self.threshold_ = - thresholds[indices[max_tnr_index]]
```

bellet: Yes, nevermind! I misread. I think it is somewhat more common to fix the error rate (false negative or false positive) while maximizing the accuracy (true positive or true negative). But it is equivalent to what you are doing, so let's keep it this way

metric_learn/base_metric.py
```diff
@@ -490,20 +491,39 @@ def calibrate_threshold(self, pairs_valid, y_valid, strategy='accuracy',
...
+      scores_sorted = np.concatenate([[scores_sorted[0] + 1], scores_sorted])
+
       # finds the threshold that maximizes the accuracy:
       cum_tp = stable_cumsum(y_ordered == 1)  # cumulative number of true
       # positives
+      # we need to add the point where all samples are rejected:
+      cum_tp = np.concatenate([[0.], cum_tp])
       cum_tn_inverted = stable_cumsum(y_ordered[::-1] == -1)
-      cum_tn = np.concatenate([[0], cum_tn_inverted[:-1]])[::-1]
+      cum_tn = np.concatenate([[0.], cum_tn_inverted])[::-1]
```
wdevazelhes: we need to include the last `cum_tn_inverted` now, since it's the one where all samples will be rejected

metric_learn/base_metric.py
```diff
@@ -490,20 +491,39 @@ def calibrate_threshold(self, pairs_valid, y_valid, strategy='accuracy',
...
+        self.threshold_ = - np.mean(scores_sorted[imax: imax + 2])
+      # note: if the best is to reject all points it's already one of the
+      # thresholds (scores_sorted[0] + 1)
       return self
 
     if strategy == 'f_beta':
-      precision, recall, thresholds = precision_recall_curve(
+      fps, tps, thresholds = _binary_clf_curve(
           y_valid, self.decision_function(pairs_valid), pos_label=1)
+
```
wdevazelhes: We need to take part of the code (but not all) from `precision_recall_curve` here because there was a section of the code in `precision_recall_curve` that deleted points when the maximum recall is attained. This was OK when we set the threshold to the the last value we accept, but it's not OK with the new computation of the threshold because we need all the points to put the threshold in the middle of two points. (Otherwise the algo can think some point is the last point of the list and so be set to this value - 1, whereas it should be the mean between this value and the next one)
I can detail more this if needed

metric_learn/base_metric.py
```diff
@@ -516,26 +536,45 @@ def calibrate_threshold(self, pairs_valid, y_valid, strategy='accuracy',
...
+        self.threshold_ = - np.mean(thresholds[imax: imax + 2])
+      # Note: we don't need to deal with rejecting all points (i.e. threshold =
+      # max_scores + 1), since this can never happen to be optimal
+      # (see a more detailed discussion in test_calibrate_threshold_extreme)
       return self
 
     fpr, tpr, thresholds = roc_curve(y_valid,
                                      self.decision_function(pairs_valid),
-                                     pos_label=1)
+                                     pos_label=1, drop_intermediate=False)
```
wdevazelhes: We need `drop_intermediate=False` here for a similar problem than above: it worked fine when we were setting the threshold to an exact value of a point, but if we remove some points the mean between one point and the next can be wrong (because the "next" point is not the actual "next" point: the actual "next" point can be removed)

wdevazelhes:
I just added the new computation of the threshold as we said, so I'll just refactor it a bit (right now it works, but it's a bit messy), and then I should document how we set the threshold with the new computation, then I should add the check before fitting for the `calibration_params`, address the few remaining comments and we should be good to go

bellet:
If it remains quite messy, we have to ask ourselves whether this mean strategy is worth the extra complexity in the code

wdevazelhes:
> If it remains quite messy, we have to ask ourselves whether this mean strategy is worth the extra complexity in the code

I agree, also, I think some of the complexity comes from the fact that we use scikit-learn functions and deal with the special cases where they remove points in curves etc... Maybe re-implementing the code would be simpler, more robust to sklearn's changes, have less comments to explain how we deal with special cases, probably some parts could also be mutualized like the cumsums so it wouldn't be a code as big as just recoding each part... I'll also think about that to see what it would look like this way

bellet:
If we call sklearn routines in a straightforward way (avoiding edges cases), I think there is no reason to expect changes. So this is an option (i.e., revert back to picking a simple returned threshold instead of averaging). Re-implementing everything on ROC/PR curves seems like an overkill?

wdevazelhes:
Yes I agree, I'll put the recent changes in a branch of my fork in case we need these changes one day, but I agree for now we can stick to the first version

wdevazelhes:
I'll just need to still cherry pick in the changes the way we can put a threshold to the worst score -1, it's the only solution if we want to reject everything (which is a case that is now also tested)

metric_learn/base_metric.py
wdevazelhes: agreed, done

wdevazelhes:
I came back to the previous implementation for the threshold, put a calibration params validation before fit, and addressed all the comments. Just a last questions before final review and merge: I noted in the todo that we could raise a warning when calibrating the threshold on the trainset at fit time. Is it really necessary in the end ? We already add a little comment about that in the User Guide. Isn't it enough ?

bellet:
LGTM. Warning is not necessary (and would be quite annoying).
Maybe @perimosocordiae or @terrytangyuan want to have a final look at this, otherwise we can merge

Issue #153, opened 2019-01-10
Put the metric_plotting example into a doc in sphinx-gallery


wdevazelhes:
This example would easily be converted in a sphinx-gallery ready example. It would be cool because it would appear in the docs.

bellet:
Not sure why this wasn't closed automatically when merging #180

Issue #171, opened 2019-02-15
Remove random seed in fit


wdevazelhes:
For supervised metric learners, we can give a random seed in `fit`, that will be used for sampling random pairs/quadruplets from the X and y. I think this seed is not data-dependent, so it should be put in the init. Doing that would also make it easier to have deterministic tests (we would just set the random state of the estimator once and for all)

PR #180, opened 2019-03-08
[MRG] Export notebook to gallery

### LINKED ISSUE #141: Update metric_plotting.ipynb with the new API
PR #139 introduce some API changes that break the example jupyter notebook. We should update it to work, using the new API

wdevazelhes:
Fixes #141 #153

Hi, I've just converted @bhargavvader 's notebook from #27 into a sphinx-gallery file (with this snippet: https://gist.github.com/chsasank/7218ca16f8d022e02a9c0deb94a310fe). This way, it will appear nicely in the documentation, and can also allow us to check if every algorithms work fine. 
There are a few things to change to make the PR mergeable (to compile the doc, you need sphinx-gallery): 
- [x] As dicussed with @bellet, the iris dataset is maybe not the most expressive dataset for metric learning, we might want to find a dataset where classes are even more mixed and where metric learning gives a very advantageous separation
- [x] Some parts seem to be broken (see the logo "broken"), I need to see why
- [x] On my computer, the plan of the notebook appears in the left toolbar, we might not want that (we might want to see only two tabs (because we have two examples in metric-learn/examples) on the left sidebar and not tens of tabs)
- [x] Some examples seem not to work super well in terms of separation, I need to see why


bellet:
I assume that in this PR you will also be fixing #141

bhargavvader:
Glad to see this going into documentation soon 💃

bellet:
And this fixes #153

wdevazelhes:
> I assume that in this PR you will also be fixing #141



> And this fixes #153

Yes indeed, thanks, I forgot to add it in the PR description 
Done now

wdevazelhes:
Regarding the dataset, should I go for the faces dataset for instance ? (Probably the supervised version since we use `_Supervised` algorithms here)

wdevazelhes:
I just ran another `make html` (after merging with master, and cleaning my folder), and now there's no "broken" message anymore (however no image is printed, only the sphinx gallery default logo, but anyway I don't know what we would want to appear here ? A concatenation of all images printed ?)

bellet:
> Regarding the dataset, should I go for the faces dataset for instance ? (Probably the supervised version since we use `_Supervised` algorithms here)

Yes. Digits could also be an option. Note that you will have to use dimensionality reduction (say t-SNE) to visualize things in 2D.

Otherwise we could work with a 2D or 3D dataset but there is not so many nice things to show in such cases

wdevazelhes:
> Yes. Digits could also be an option. Note that you will have to use dimensionality reduction (say t-SNE) to visualize things in 2D.
> 
> Otherwise we could work with a 2D or 3D dataset but there is not so many nice things to show in such cases

Allright, I tried to launch the example with digits, but there is a conditionning problem with SDML, so I think we'll need to wait for #162 to be merged to finish this one

wdevazelhes:
Here is a first result of plotting with the faces dataset (with the digits dataset points are already well separated out with tsne on the raw dataset): 

tsne from the raw dataset: 
image

tsne after lmnn:
image

tsne after sdml:
image

tsne after lsml:
image

tsne after nca:
image

tsne after lfda:
image

RCA failed (I still need to understand why)

The improvement is not very clear, except maybe for lmnn, but for lfda for instance it's worse... Though it could work with appropriate tuning. 
I took faces because in the PR for NCA it's one dataset where we had a significant improvement (see this figure: https://user-images.githubusercontent.com/31916524/41354636-c759ddce-6f1f-11e8-8ac9-9a9fd36b8af2.png), but I'll try with balance, it was also good and it's easy to fetch now with scikit-learn with the openml fetcher.

wdevazelhes:
I tried with the balance dataset and reducing the dimension gives a bit weird results (like I think balance kind of defines a grid of regularly spaced points, which seems a bit like an artificial dataset) (but anyway running the default version of the algos didn't work so much either)
I also tried with the isolet dataset and it didn't work that much 
So in the end I think the best for now is the faces dataset. 
Maybe I can re-run the faces dataset, but this time grid-searching for the best parameters
I'll also try to find another dataset hoping it would work better

bellet:
An option would be to add some noise dimensions (this is a bit artificial but will definitely work)

Otherwise you have to look for harder datasets - if they t-SNE visualization is already very good on the original representation, it is hard to obtain clear visual improvements

wdevazelhes:
That's right, it works really well indeed with noise: 

Here are the images with 5 columns of gaussian noise N(0, 10) added to iris (the legend with sepal etc is wrong, but the image is good, it's the results after applying t-sne): 
tsne on original space: 
image

tsne after LMNN: 
image


tsne after ITML:
image

tsne after SDML: 
image

tsne after LSML:
image



tsne after NCA: 
image

tsne after LFDA: 
image

tsne after RCA:
image

So maybe we can use this with a quick intro paragraph saying "let's say we have a noisy dataset: let's take iris and add some columns of noise etc.." ?

Otherwise do you know any dataset with a lot of noise already inside it ?

Also, LSLM and SDML work not so good, so I guess as soon as we settle on the dataset I'll try to fine tune them and if it still doesn't work maybe we can remove them ?

wdevazelhes:
In the end I used iris with 5 columns of noise added, in the last commit is a version that I think is mergeable for the example, let me know what you think
Here is the html folder that was generated: 

[html.zip](https://github.com/metric-learn/metric-learn/files/3063854/html.zip)

wdevazelhes:
Note that there is still `LSML` that does not work really well, but it's still better than the initial example, so I think it's OK ?

perimosocordiae:
I haven't read all the text yet, but I agree this is a good approach and a reasonable example to use.

wdevazelhes:
I just pushed an example that uses `make_classification` which I guess looks pretty good now,
I just needed to add a scaling to the noise components of the dataset by hand (to mix more the initial points while still making them easy to fit for metric learning), because it's not an argument of `make_classification` I think, and I also tried to add a lot more of noise components instead but it does not seem as efficient (I need to add more than hundreds of them to see the effect so I'm not sure it's a good idea given that we have 100 samples (I could add more samples but then the example takes more time to run))

@bellet I also wrote something about the fact that some algorithms don't try to cluster all points from the same class in one same cluster, while others implicitly do that, which can be seen quite clearly from these examples (since every class has a distribution with 2 clusters: `n_clusters_per_class=2`), as we discussed. I just wrote a paragraph because I think we see the effect quite clearly on the examples, so maybe there's no need to put the two figures of applying the algorithms to one dataset with 1 cluster per class, and a dataset with several clusters per class as mentioned? 
Or maybe there is need because it's good to have multi-cluster cases where algorithms that enforce mono-clustering fail clearly ? But then I don't know exactly how to tweak the dataset to do that (since now these algorithms mostly manage to group the data in mono-clusters), maybe with less dimensions, to give less degrees of freedom to the algorithms ? 

If I understood correctly, the algorithm that don't cluster similar points in a unique cluster are `LMNN`, `NCA`, and `LFDA`, is that right ?

wdevazelhes:
If these changes are fine I think we are good to merge

bellet:
I just had a quick look and it looks good! I have a few nitpick suggestions on the wording and presentation, will write a review asap

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# 
+# We will be using the IRIS data-set to illustrate the plotting. You can
+# read more about the IRIS data-set here:
+# `link <https://en.wikipedia.org/wiki/Iris_flower_data_set>`__.
+
+X, Y = make_classification(n_classes=3, n_clusters_per_class=2,
+                           n_informative=3, class_sep=4., n_features=5,
+                           n_redundant=0, shuffle=True)
+
```
bellet: maybe add `n_samples=100` explicitly

and add explanation about what the dataset looks like:
- 100 points in 3 classes with 2 clusters per class
- 5 features, among which 3 are informative (correlated with the class labels) and two are random noise with large magnitude

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+##########################################################################
+# Let's make the noise of the non-informative features higher, and shuffle
+# the dataset:
+
+X[:, 3:] *= 20
+X, Y = shuffle(X, Y)
```
bellet: useless as you have used `shuffle=True`

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+
+X, Y = make_classification(n_classes=3, n_clusters_per_class=2,
+                           n_informative=3, class_sep=4., n_features=5,
+                           n_redundant=0, shuffle=True)
+
+##########################################################################
+# Let's make the noise of the non-informative features higher, and shuffle
+# the dataset:
+
+X[:, 3:] *= 20
```
bellet: use `scale=[1, 1, 20, 20, 20]` in `make_classification` instead

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+X, Y = shuffle(X, Y)
+
+
+###########################################################################
+# Note that the dimensionality of the data is 5, so to plot the
+# transformed data in 2D, we will use the t-sne algorithm. (See
+# `sklearn.manifold.TSNE`).
+
+def plot(X, Y):
```
bellet: rename to `plot_tsne`?

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+    plt.yticks(())
+
+    plt.show()
+
+###################################
+# Let's now plot the dataset as is.
+
+plot(X, Y)
+
+
```
bellet: add comment on the visualization explaining that classes are mixed up due to the noisy features

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+# Why is Metric Learning useful? We can, with prior knowledge of which
+# points are supposed to be closer, figure out a better way to understand
+# distances between points. Especially in higher dimensions when Euclidean
+# distances are a poor way to measure distance, this becomes very useful.
+# 
+# Basically, we learn this distance:
+# :math:`D(x,y)=\sqrt{(x-y)\,M^{-1}(x-y)}`. And we learn this distance by
+# learning a Matrix :math:`M`, based on certain constraints.
+# 
+# Some good reading material for the same can be found
```
bellet: also point to the section "What is Metric Learning"? of the doc

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+# all data instances in the training set are surrounded by at least k
+# instances that share the same class label. If this is achieved, the
+# leave-one-out error (a special case of cross validation) is minimized.
+# You'll notice that the points from the same labels are closer together,
+# but they are not necessary in a same cluster. This is particular to LMNN
+# and we'll see that some other algorithms implicitly enforce points from
+# the same class to cluster together.
+# 
+# You can find the paper
+# `here <http://jmlr.csail.mit.edu/papers/volume10/weinberger09a/weinberger09a.pdf>`__.
```
bellet: instead of paper, it would be better to have a link to the documentation and docstring pages

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+# Implements an efficient sparse metric learning algorithm in high
+# dimensional space via an :math:`l_1`-penalised log-determinant
+# regularization. Compare to the most existing distance metric learning
+# algorithms, the algorithm exploits the sparsity nature underlying the
+# intrinsic high dimensional feature space.
+# 
+# Link to paper here:
+# `SDML <http://lms.comp.nus.edu.sg/sites/default/files/publication-attachments/icml09-guojun.pdf>`__.
+# 
+# One feature which we'd like to show off here is the use of random seeds.
```
bellet: not sure this makes much sense here? perhaps remove this paragraph? or keep it but without mentioning constraints specifically (just randomness in the algorithm)

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+# transform our input space
+X_lmnn = lmnn.transform(X)
+
+
+######################################################################
+# So what have we learned? The matrix :math:`M` we talked about before.
+# Let's see what it looks like.
+# 
+
+lmnn.metric()
```
bellet: does not print in the doc. should use `print`? but if it is not possible to easily interpret what we see, we can get rid of it

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+# LSML is a simple, yet effective, algorithm that learns a Mahalanobis
+# metric from a given set of relative comparisons. This is done by
+# formulating and minimizing a convex loss function that corresponds to
+# the sum of squared hinge loss of violated constraints.
+# 
+# Link to paper:
+# `LSML <http://web.cs.ucla.edu/~weiwang/paper/ICDM12.pdf>`__
+# 
+
+lsml = metric_learn.LSML_Supervised(num_constraints=200, tol=0.0001, max_iter=10000, verbose=True)
```
bellet: remove verbose

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
+"""
+Algorithms walkthrough
+~~~~~~~~~~~~~~~~~~~~~~
+
+This is a small walkthrough which illustrates all the Metric Learning
```
bellet: not all actually

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+plot(X_itml, Y)
+
+
+######################################################################
+# Sparse Determinant Metric Learning
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# 
+# Implements an efficient sparse metric learning algorithm in high
+# dimensional space via an :math:`l_1`-penalised log-determinant
+# regularization. Compare to the most existing distance metric learning
```
bellet: compared

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+X_lsml = lsml.fit_transform(X, Y)
+
+plot(X_lsml, Y)
+
+
+######################################################################
+# Neighborhood Components Analysis
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# 
+# NCA is an extrememly popular metric-learning algorithm, and one of the
```
bellet: extremely
maybe remove last part of the sentence

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+# 
+
+rca = metric_learn.RCA_Supervised(num_chunks=30, chunk_size=2)
+X_rca = rca.fit_transform(X, Y)
+
+plot(X_rca, Y)
+
+
+######################################################################
+# Manual Constraints
```
bellet: We do not say explicitly what is meant by "constraint" before so I think this part is quite confusing

Also I think we want to insist on the fact that many metric learning algorithms only need the weak supervision given by constraints, not labels. In many applications, this is easier to obtain.

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+# constraints! A bit different from our old result but not too different.
+# We can also notice that it might be better to rely on the randomised
+# algorithms under the hood to make our constraints if we are not very
+# sure how we want our transformed space to be.
+# 
+# RCA and SDML also have their own specific ways of taking in inputs -
+# it's worth one's while to poke around in the constraints.py file to see
+# how exactly this is going on.
+# 
+# This brings us to the end of this tutorial! Have fun Metric Learning :)
```
bellet: maybe also add that metric-learn is compatible with sklearn and so we can easily do model selection, cross validation, scoring etc and refer to the doc for more details.

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
+"""
+Algorithms walkthrough
+~~~~~~~~~~~~~~~~~~~~~~
+
+This is a small walkthrough which illustrates all the Metric Learning
+algorithms implemented in metric-learn, and also does a quick
+visualisation which can help understand which algorithm might be best
+suited for you.
+
+Of course, depending on the data set and the constraints your results
```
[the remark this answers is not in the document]
wdevazelhes: Agreed, done

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
+"""
+Algorithms walkthrough
+~~~~~~~~~~~~~~~~~~~~~~
+
+This is a small walkthrough which illustrates all the Metric Learning
+algorithms implemented in metric-learn, and also does a quick
```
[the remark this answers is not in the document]
wdevazelhes: Agreed, done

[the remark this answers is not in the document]
wdevazelhes: Agreed, done

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+suited for you.
+
+Of course, depending on the data set and the constraints your results
+will look very different; you can just follow this and change your data
+and constraints accordingly.
+
+"""
+
+# License: BSD 3 clause
+# Authors: Bhargav Srinivasa Desikan <bhargavvader@gmail.com>
```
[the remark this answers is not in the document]
wdevazelhes: Agreed, done

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,410 @@
...
+# visualisation imports
+import matplotlib.pyplot as plt
+np.random.seed(42)
+
+
+######################################################################
+# Loading our data-set and setting up plotting
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# 
+# We will be using the IRIS data-set to illustrate the plotting. You can
```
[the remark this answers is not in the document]
wdevazelhes: Thanks, I forgot to update this indeed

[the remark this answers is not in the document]
wdevazelhes: done

examples/plot_metric_examples.py
wdevazelhes: I agree, done

examples/plot_metric_examples.py
wdevazelhes: That's right (actually I wanted to use shuffle=False so that then I rescale the features (because if shuffle=True it shuffles the features too), but with your next comment there is no need to do that in fact)

examples/plot_metric_examples.py
wdevazelhes: Thanks, I didn't see that, it's very convenient !

examples/plot_metric_examples.py
wdevazelhes: Agreed, done

examples/plot_metric_examples.py
wdevazelhes: I agree, I'll add an explanation, let me know what you think

examples/plot_metric_examples.py
wdevazelhes: I agree, done

examples/plot_metric_examples.py
wdevazelhes: Agreed, done

examples/plot_metric_examples.py
wdevazelhes: I agree, the simpler is to remove it I guess, since we'll also change this and put the random seed in init (see #171)

examples/plot_metric_examples.py
wdevazelhes: Agreed, done

examples/plot_metric_examples.py
wdevazelhes: Indeed, done

examples/plot_metric_examples.py
wdevazelhes: Thanks, done for extremely 
For the last part, is it because it can make people that don't know the algorithm think that if it's one of the first few it might be not cutting edge/outdated ? 
Done

bellet: just that it is maybe not such a relevant thing to mention here. you could also argue this for LMNN, and others like MMC

examples/plot_metric_examples.py
wdevazelhes: I agree, done

examples/plot_metric_examples.py
wdevazelhes: I agree, I'll reformulate this paragraph, let me know what you think

examples/plot_metric_examples.py
wdevazelhes: That's right, they are also one of the first ones indeed

examples/plot_metric_examples.py
wdevazelhes: Agreed, I will add MMC, but I don't know for MLKR, I think I'll just talk about it when describing NCA ? Since the cost function is very similar except that MLKR uses a soft-nearest neighbors for regression as I understood

examples/plot_metric_examples.py
wdevazelhes: I think we can also delete the `num_constraints=200` in all `*_Supervised` algorithms what do you think ?

bellet: i guess so, unless it increases the computation time too much

examples/plot_metric_examples.py
bellet: true, MLKR is for regression. it is a good idea to mention it there as proposed
maybe change "all the metric learning algorithms" to "most metric learning algorithms"

wdevazelhes: In the end, I added most metric learning algorithms because we don't talk about `Covariance`, (we could maybe ?) And I added `MLKR` with `make_regression` task because I found the results were pretty cool too! Even if it breaks a bit the outline because it's just between supervised algorithms and the "constraints" section... Tell me what you think ?

wdevazelhes:
@bellet Thanks for your review, I addressed your comments

examples/plot_metric_examples.py
wdevazelhes: That's right, I tried and it tooked almost the same time, so I removed them

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+plot_tsne(X, Y)
+
+#########################################################################
+# We can see that the classes appear mixed up: this is because t-sne
+# is based on preserving the original neighborhood of points in the embedding
+# space, but this original neighborhood is based on the euclidean
+# distance in the input space, in which the contribution of the noisy
+# features is high. So even if points from the same class are close to
+# each other in some subspace of the input space, this is not the case in the
+# total input space.
```
bellet: maybe "this is not the case when considering all dimensions of the input space"

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# :math:`D(x,y)=\sqrt{(x-y)\,M^{-1}(x-y)}`. And we learn this distance by
+# learning a Matrix :math:`M`, based on certain constraints.
+#
+# For more information, check the :ref:`intro_metric_learning` section
+# from the documentation. Some good reading material can also be found
+# `here <https://arxiv.org/pdf/1306.6709.pdf>`__. It serves as a
+# good literature review of Metric Learning.
+#
+# We will briefly explain the metric-learning algorithms implemented by
+# metric-learn, before providing some examples for it's usage, and also
```
bellet: its

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# ^^^^^^^^^^^^^^^
+#
+# Why is Metric Learning useful? We can, with prior knowledge of which
+# points are supposed to be closer, figure out a better way to understand
+# distances between points. Especially in higher dimensions when Euclidean
+# distances are a poor way to measure distance, this becomes very useful.
+# 
+# Basically, we learn this distance:
+# :math:`D(x,y)=\sqrt{(x-y)\,M^{-1}(x-y)}`. And we learn this distance by
+# learning a Matrix :math:`M`, based on certain constraints.
```
bellet: - there is repetition here "we learn this distance"
- add very quick explanation of what is meant by "constraint" (can be inspired from What is metric learning? page). For instance something like, "we learn the parameters :math:`M` of this distance to satisfy certain constraints on the distance between points, for example requiring that points of the same class are close together and points of different class are far away."

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+#
+# For more information, check the :ref:`intro_metric_learning` section
+# from the documentation. Some good reading material can also be found
+# `here <https://arxiv.org/pdf/1306.6709.pdf>`__. It serves as a
+# good literature review of Metric Learning.
+#
+# We will briefly explain the metric-learning algorithms implemented by
+# metric-learn, before providing some examples for it's usage, and also
+# discuss how to go about doing manual constraints.
```
bellet: remove manual constraint and replace by something like "discuss how to perform metric learning with weaker supervision than class labels"

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# - See more in the documentation of the class :py:class:`RCA
+#   <metric_learn.rca.RCA>`
+
+rca = metric_learn.RCA_Supervised(num_chunks=30, chunk_size=2)
+X_rca = rca.fit_transform(X, Y)
+
+plot_tsne(X_rca, Y)
+
+######################################################################
+# Metric Learning for Kernel Regression
```
bellet: Regression example: Metric Learning for Kernel Regression

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# An algorithm very similar to NCA but for regression is Metric
```
bellet: Before this sentence, need to make clear that we are switching to regression setting:
"The previous algorithms took as input a dataset with class labels. Metric learning can also be useful for regression, when the labels are real numbers."

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+#
+# - See more in the :ref:`User Guide <mlkr>`
+# - See more in the documentation of the class :py:class:`MLKR
+#   <metric_learn.mlkr.MLKR>`
+#
+# To illustrate MLKR, let's use the dataset
+# `sklearn.datasets.make_regression` the same way as we did with the
+# classification  before. The dataset will contain: 100 points of 5 features
+# each, among which 3 are random noise but informative (used to generate the
+# regression target from a linear model, and two are random noise with the
```
bellet: missing closing parenthesis

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+plot_tsne(X_mlkr, Y_reg, plt.cm.Oranges)
+
+######################################################################
+# Points that have the same value to regress are now closer to each
+# other ! This could improve the performance of
+# `sklearn.neighbors.KNeighborsRegressor` for instance.
+
+
+######################################################################
+# Constraints
```
bellet: Rename as "Metric learning from weaker supervision"

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# ^^^^^^^^^^^
+#
+# To learn the right metric, so far we have always given the labels of the
```
bellet: "right" maybe not useful?

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# data to supervise the algorithms. However, in many applications,
+# it is easier to obtain information about whether two samples are
+# similar or dissimilar. For instance, when annotating a dataset of face
+# images, it is easier for an annotator to tell if two faces belong to the same
+# person or not, rather than finding the ID of the face among a huge database
+# of every person's faces.
```
bellet: Add: "Note that for some problems (e.g., in information retrieval where the goal is to rank documents by similarity to a query document), there is no notion of individual label but one can gather information on which pairs of points are similar or dissimilar."

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+#
+# Fortunately, one of the strength of metric learning is the ability to
+# deal with such data. Indeed, some of the algorithms we've mentioned have
```
bellet: - to learn from such weaker supervision.
- some of the algorithms we've used above

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# alternate ways to pass some supervision about the metric we want to learn.
+# The way to go is to pass a 3D array `pairs` of pairs, as well as an array
+# of labels `y` such that for each `i` between `0` and `n_pairs`
+# we want `X[i, 0, :]` and `X[i, 1, :]` to be similar if `y[i] == 1`, and we
+# want them to be dissimilar if `y[i] == -1`. In other words, we want to
+# enforce a metric that projects similar points closer together and
+# dissimilar points further away from each other.
+# (See also the section: :ref:`weakly_supervised_section`.)
+#
+# This kind of input is possible for ITML, SDML, and MMC.
```
bellet: put this sentence before the pointer to the doc

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# Fortunately, one of the strength of metric learning is the ability to
+# deal with such data. Indeed, some of the algorithms we've mentioned have
+# alternate ways to pass some supervision about the metric we want to learn.
+# The way to go is to pass a 3D array `pairs` of pairs, as well as an array
+# of labels `y` such that for each `i` between `0` and `n_pairs`
+# we want `X[i, 0, :]` and `X[i, 1, :]` to be similar if `y[i] == 1`, and we
+# want them to be dissimilar if `y[i] == -1`. In other words, we want to
+# enforce a metric that projects similar points closer together and
+# dissimilar points further away from each other.
+# (See also the section: :ref:`weakly_supervised_section`.)
```
bellet: see [ref] for details on other kinds of weak supervision that some algorithms can work with

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+#
+# This kind of input is possible for ITML, SDML, and MMC.
+#
+# We're going to create these constraints through the labels we have, i.e
```
bellet: For the purpose of this example, we're going to explicitly create these pairwise constraints

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# :math:`Y`.
+#
+# This is done internally through metric learn anyway (do check out the
```
bellet: Note that this is what metric-learn did under the hood in the previous examples

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# `constraints` module!) - but we'll try our own version of this. I'm
```
bellet: I --> We

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# going to go ahead and assume that two points labelled the same will be
+# closer than two points in different labels.
+#
+# Do keep in mind that we are doing this method because we know the labels
```
bellet: move this above, right after the sentence saying that we are going to create constraints from the labels

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+######################################################################
+# Now that we've created our constraints, let's see what it looks like!
+# 
+
+print(pairs)
+print(pairs_labels)
+
+
+######################################################################
+# Using our constraints, let's now train ITML again. We should keep in
```
bellet: not sure the last sentence is needed (already said that before)

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+
+X_itml = itml.transform(X)
+
+plot_tsne(X_itml, Y)
+
+
+######################################################################
+# And that's the result of ITML after being trained on our manual
+# constraints! A bit different from our old result but not too different.
+# We can also notice that it might be better to rely on the randomised
```
bellet: this last sentence is not very clear. i would remove it

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+# Finally, one of the main advantages of metric-learn is its out-of-the box
+# compatibility with scikit-learn, for doing `model selection
+# <https://scikit-learn.org/stable/model_selection.html>`__,
+# cross-validation, and scoring for instance. Indeed, supervised algorithms are
+# regular `sklearn.base.TransformerMixin` that can be plugged into any
+# pipeline or cross-validation procedure. And weakly-supervised estimators are
+# also compatible with scikit-learn, since their input dataset format described
+# above allows to be sliced along the first dimension when doing
+# cross-validations (see also this :ref:`section <sklearn_compat_ws>`). See
+# also some :ref:`use cases <use_cases>` where you could use scikit-learn
```
bellet: where you could combine metric learning with scikit-learn estimators

examples/plot_metric_examples.py
```diff
@@ -0,0 +1,474 @@
...
+#
+# Finally, one of the main advantages of metric-learn is its out-of-the box
+# compatibility with scikit-learn, for doing `model selection
+# <https://scikit-learn.org/stable/model_selection.html>`__,
+# cross-validation, and scoring for instance. Indeed, supervised algorithms are
+# regular `sklearn.base.TransformerMixin` that can be plugged into any
+# pipeline or cross-validation procedure. And weakly-supervised estimators are
+# also compatible with scikit-learn, since their input dataset format described
+# above allows to be sliced along the first dimension when doing
+# cross-validations (see also this :ref:`section <sklearn_compat_ws>`). See
```
bellet: avoid repetition of "see"

examples/plot_metric_examples.py
wdevazelhes: Agreed, done

examples/plot_metric_examples.py
wdevazelhes: Thanks, done

examples/plot_metric_examples.py
wdevazelhes: Agreed, done

examples/plot_metric_examples.py
wdevazelhes: Agreed, done

examples/plot_metric_examples.py
wdevazelhes: done

examples/plot_metric_examples.py
wdevazelhes: That's right, done

examples/plot_metric_examples.py
wdevazelhes: thanks, done

examples/plot_metric_examples.py
wdevazelhes: That's right, done

examples/plot_metric_examples.py
wdevazelhes: that's right, done

examples/plot_metric_examples.py
wdevazelhes: That's an important point indeed, done

examples/plot_metric_examples.py
wdevazelhes: done

examples/plot_metric_examples.py
wdevazelhes: done

examples/plot_metric_examples.py
wdevazelhes: done

examples/plot_metric_examples.py
wdevazelhes: done

examples/plot_metric_examples.py
wdevazelhes: done

examples/plot_metric_examples.py
wdevazelhes: done

examples/plot_metric_examples.py
wdevazelhes: Agreed, done

wdevazelhes:
Thanks for the review @bellet, I addressed all your comments

bellet:
Thanks! I have pushed a few small updates myself to avoid another reviewing cycle. I think we can merge once CI passes

PR #186, opened 2019-03-20
[MRG] Update pytest

wdevazelhes:
This makes us update to the lastest pytest

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -160,7 +160,6 @@ def test_check_input_invalid_type_of_inputs(type_of_inputs):
 #  ---------------- test check_input with 'tuples' type_of_input' ------------
 
 
-@pytest.fixture
 def tuples_prep():
   """Basic array for testing when using a preprocessor"""
   tuples = np.array([[1, 2],
@@ -168,7 +167,6 @@ def tuples_prep():
   return tuples
 
 
-@pytest.fixture
 def tuples_no_prep():
   """Basic array for testing when using no preprocessor"""
   tuples = np.array([[[1., 2.3], [2.3, 5.3]],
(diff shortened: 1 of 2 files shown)


wdevazelhes:
Apparently I used fixtures in a bad way sometimes, which was still supported in old pytest versions but not anymore so I deleted fixtures and replaced them by regular functions when needed

test/test_utils.py
```diff
@@ -410,15 +404,15 @@ def test_check_tuples_behaviour_auto_dtype(tuples_no_prep):
...
   with pytest.warns(None) as record:
-      check_input(tuples_no_prep, type_of_inputs='tuples')  # numeric type
+      check_input(tuples_no_prep(), type_of_inputs='tuples')  # numeric type
   assert len(record) == 0
 
   # not numeric type
-  tuples_no_prep = np.array([[['img1.png'], ['img2.png']],
-                             [['img3.png'], ['img5.png']]])
-  tuples_no_prep = tuples_no_prep.astype(object)
+  tuples_no_prep_bis = np.array([[['img1.png'], ['img2.png']],
```
wdevazelhes: If I let `tuples_no_prep` here, I have an error saying `tuples_no_prep` is referenced before assignment... I didn't understand why ? Since `tuples_no_prep` is a function defined above in the module ?  But changing into `tuples_no_prep_bis` it works
(I tried with a simple example and I still don't understand: for instance this script returns `UnboundLocalError: local variable 'a' referenced before assignment
` when executed: 
```python
def a():
    return 5

def test():
    c = a()
    a = 3

if __name__ == '__main__':
    test()
```
Does python automatically detects that the variable "a"  is assigned in "test" and that it must be a variable local to "test" ?)

perimosocordiae: Yeah, Python treats local variables specially at the bytecode level. Here's the result of `dis.dis(test)` using your example code:

```
  5           0 LOAD_FAST                0 (a)
              3 CALL_FUNCTION            0
              6 STORE_FAST               1 (c)

  6           9 LOAD_CONST               1 (3)
             12 STORE_FAST               0 (a)
             15 LOAD_CONST               0 (None)
             18 RETURN_VALUE
```

The `LOAD_FAST` opcode is specifically looking for a local variable named `a`, which is why you get the UnboundLocalError.

Changing the assignment of `a` to a different name results in a `LOAD_GLOBAL` opcode, which works as expected:

```
In [5]: def test2():
   ...:     c = a()
   ...:     b = 3
   ...:

In [6]: test2()

In [7]: dis.dis(test2)
  2           0 LOAD_GLOBAL              0 (a)
              3 CALL_FUNCTION            0
              6 STORE_FAST               0 (c)

  3           9 LOAD_CONST               1 (3)
             12 STORE_FAST               1 (b)
             15 LOAD_CONST               0 (None)
             18 RETURN_VALUE
```

You could also mark `a` global in the function, but this will result in overwriting the module-level function:

```
In [8]: def test3():
   ...:     global a
   ...:     c = a()
   ...:     a = 3
   ...:

In [9]: test3()

In [10]: test3()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-10-5dea5f7ba8cb> in <module>()
----> 1 test3()

<ipython-input-8-32a55428a886> in test3()
      1 def test3():
      2     global a
----> 3     c = a()
      4     a = 3
      5

TypeError: 'int' object is not callable

In [11]: dis.dis(test3)
  3           0 LOAD_GLOBAL              0 (a)
              3 CALL_FUNCTION            0
              6 STORE_FAST               0 (c)

  4           9 LOAD_CONST               1 (3)
             12 STORE_GLOBAL             0 (a)
             15 LOAD_CONST               0 (None)
             18 RETURN_VALUE
```

perimosocordiae:
Merged, thanks!

test/test_utils.py
wdevazelhes: Very interesting, thanks ! :)

PR #188, opened 2019-03-26
[MRG] Fix: SDML on sandwich example

### LINKED ISSUE #138: SDML does not work in sandwich example

bellet:
Fixes #138 

Thanks to #162 we can now make SDML work on the sandwich example by reducing `balance_param` (see figure)

sandwich


wdevazelhes:
Great !

PR #191, opened 2019-04-11
[MRG] make transformer_from_metric more robust

### LINKED ISSUE #175: `transformer_from_metric` can return wrong results or nan
#### Description
Our strategy to compute `transformer_from_metric`can lead to wrong results: for instance, there is a bug in the following example:
-  the transformation matrix will have nans in it because some elements of the Mahalanobis matrix are negative.  In this case we should maybe put to zero some elements that are close to zero but a bit negative
- Also, in the particular case of the example, all the coeffs of the Mahalanobis matrix are at the same order of magnitude (1e-9, see below), so in fact its not really a diagonal matrix. It will not raise a nan or bug but still will lead to a wrong result because the matrix will be considered as a diagonal matrix. We should not just test the diagonal in absolute but we should probably test the diagonal *relatively* to the non-diagonal coefficients

#### Steps/Code to Reproduce

```python
from metric_learn import ITML
from sklearn.datasets import load_iris
from sklearn.utils import shuffle
from metric_learn import Constraints
from sklearn.utils import check_random_state
import numpy as np

SEED = 42

input_data, labels = load_iris(return_X_y=True)
X, y = shuffle(input_data, labels, random_state=SEED)
num_constraints = 50
constraints = Constraints(y)
pairs = (constraints
      .positive_negative_pairs(num_constraints, same_length=True,
                               random_state=check_random_state(SEED)))
c = np.vstack([np.column_stack(pairs[:2]), np.column_stack(pairs[2:])])
target = np.concatenate([np.ones(pairs[0].shape[0]),
                           - np.ones(pairs[0].shape[0])])
c, target = shuffle(c, target, random_state=SEED)

itml = ITML()
itml.fit(X[c], target)
print(itml.get_mahalanobis_matrix())
print(itml.predict(X[c]))
```

#### Expected Results
No error is thrown, the transformation matrix contains real values, and the result is not all -1

#### Actual Results
```ptb
/home/will/anaconda3/envs/py27sklearnmin/bin/python /home/will/.PyCharmCE2018.3/config/scratches/scratch_50.py
[[ nan  nan  nan  nan]
/home/will/Code/metric-learn/metric_learn/_util.py:346: RuntimeWarning: invalid value encountered in sqrt
 [ nan  nan  nan  nan]
  return np.sqrt(metric)
 [ nan  nan  nan  nan]
 [ nan  nan  nan  nan]]
/home/will/Code/metric-learn/metric_learn/base_metric.py:335: RuntimeWarning: invalid value encountered in greater_equal
  return 2 * (self.decision_function(pairs) >= - self.threshold_) - 1
[-1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1
 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1
 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1
 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1 -1]
```

#### Versions
```
Linux-4.4.0-142-generic-x86_64-with-debian-stretch-sid
('Python', '2.7.15 |Anaconda, Inc.| (default, May  1 2018, 23:32:55) \n[GCC 7.2.0]')
('NumPy', '1.8.2')
('SciPy', '0.13.3')
('Scikit-Learn', '0.20.3')
('Metric-Learn', '0.4.0')
```

## Note: here is the Mahalanobis matrix that I got in debug mode: 

```
[[  1.79429787e-09   1.06777934e-09  -3.56650702e-09  -1.25100147e-09]
 [  1.06777933e-09   2.07953604e-09  -3.35935951e-09  -2.34694895e-09]
 [ -3.56650696e-09  -3.35935864e-09   8.69685077e-09   3.69865990e-09]
 [ -1.25100146e-09  -2.34694906e-09   3.69866011e-09   4.22204202e-09]]
```

## Note: here is our current implem of `transformer_from_metric`:
https://github.com/metric-learn/metric-learn/blob/bf5c7224cc7ad4c025e15b247a80e076b7f75062/metric_learn/_util.py#L345-L351

wdevazelhes:
Fixes #175 

This PR fixes the problem we had with `transformer_from_metric`, by checking if the matrix is diagonal depending on the relative gap between the maximum absolute value of the outer diagonal coeffs, and the minimum absolute value of the diagonal coefficients, rather than what was done before.

It is also more robust for detecting whether to use Cholesky or the eigendecomposition: instead of computing the determinant to check if the matrix is not definite, it tries to do Cholesky and if an error is returned it does the eigendecomposition. Indeed, maybe the determinant can be not null if there is a very small eigenvalue (`v`=1e-5) (that should be considered as null), but a lot of big eigenvalues (see one example in the test)

It also raises an error message when the input is not symmetric and a warning when switching from the Cholesky decomposition to the eigendecompostion if the Cholesky decomposition is not doable.

TODO:
- [x] to make a good test, check the value of `v` for which cholesky will fail, and from it build the example where the determinant is not "np.close" to zero

### DIFF
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -335,10 +335,10 @@ def _check_sdp_from_eigen(w, tol=None):
   w : array-like, shape=(n_eigenvalues,)
     Eigenvalues to check for non semidefinite positiveness.
 
-  tol : float, optional
+  tol : positive `float`, optional
     Negative eigenvalues above - tol are considered zero. If
-    tol is None, and w are `metric`'s eigenvalues, and eps is the
-    epsilon value for datatype of w, then tol is set to w.max() * len(w) * eps.
+    tol is None, and eps is the epsilon value for datatype of w, then tol
+    is set to w.max() * len(w) * eps.
 
   See Also
   --------
@@ -347,33 +347,26 @@ def _check_sdp_from_eigen(w, tol=None):
   """
   if tol is None:
     tol = w.max() * len(w) * np.finfo(w.dtype).eps
-  if any(w[w < 0] < - tol):
+  assert tol >= 0, ValueError("tol should be positive.")
+  if any(w < - tol):
       raise ValueError("Matrix is not positive semidefinite (PSD).")
 
 
 def transformer_from_metric(metric, tol=None):
-  """Computes the transformation matrix from the Mahalanobis matrix.
+  """Returns the transformation matrix from the Mahalanobis matrix.
 
-  Since by definition the metric `M` is positive semi-definite (PSD), it
-  admits a Cholesky decomposition: L = cholesky(M).T. However, currently the
-  computation of the Cholesky decomposition used does not support
-  non-definite matrices. If the metric is not definite, this method will
-  return L = V.T w^( -1/2), with M = V*w*V.T being the eigenvector
-  decomposition of M with the eigenvalues in the diagonal matrix w and the
-  columns of V being the eigenvectors. If M is diagonal, this method will
-  just return its elementwise square root (since the diagonalization of
-  the matrix is itself).
+  Returns the transformation matrix from the Mahalanobis matrix, i.e. the
+  matrix L such that metric=L.T.dot(L).
 
   Parameters
   ----------
   metric : symmetric `np.ndarray`, shape=(d x d)
     The input metric, from which we want to extract a transformation matrix.
 
-  tol : positive float, optional
+  tol : positive `float`, optional
     Eigenvalues of `metric` between 0 and - tol are considered zero. If tol is
-    None, and w are `metric`'s eigenvalues, and eps is the epsilon value for
-    datatype of w, then tol is set to w.max() * len(w) * eps.
-
+    None, and w_max is `metric`'s largest eigenvalue, and eps is the epsilon
+    value for datatype of w, then tol is set to w_max * metric.shape[0] * eps.
 
   Returns
   -------
@@ -382,13 +375,21 @@ def transformer_from_metric(metric, tol=None):
   """
   if not np.allclose(metric, metric.T):
     raise ValueError("The input metric should be symmetric.")
+  # If M is diagonal, we will just return the elementwise square root:
   if np.array_equal(metric, np.diag(np.diag(metric))):
     _check_sdp_from_eigen(np.diag(metric), tol)
     return np.diag(np.sqrt(np.maximum(0, np.diag(metric))))
   else:
     try:
+      # if `M` is positive semi-definite, it will admit a Cholesky
+      # decomposition: L = cholesky(M).T
       return np.linalg.cholesky(metric).T
     except LinAlgError:
+      # However, currently np.linalg.cholesky does not support indefinite
+      # matrices. So if the latter does not work we will return L = V.T w^(
+      # -1/2), with M = V*w*V.T being the eigenvector decomposition of M with
+      # the eigenvalues in the diagonal matrix w and the columns of V being the
+      # eigenvectors.
       w, V = np.linalg.eigh(metric)
       _check_sdp_from_eigen(w, tol)
       return V.T * np.sqrt(np.maximum(0, w[:, None]))
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -1,5 +1,7 @@
+import warnings
 import numpy as np
 import six
+from numpy.linalg import LinAlgError
 from sklearn.utils import check_array
 from sklearn.utils.validation import check_X_y
 from metric_learn.exceptions import PreprocessorError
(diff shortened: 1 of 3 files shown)


wdevazelhes:
Now that I think about it, maybe the warning is not a good thing to throw ? Maybe it would be better a verbose flag to decide whether to print the warning (which would not be anymore a warning but a regular print) ? What do you think ?

wdevazelhes:
I agree, I just pushed a more recent version that takes into account these remarks

metric_learn/_util.py
```diff
@@ -324,7 +326,32 @@ def check_collapsed_pairs(pairs):
...
+  level, with a default value of the tolerance depending on the eigenvalues.
+
+  Parameters
+  ----------
+  w : array-like, shape=(n_eigenvalues,)
+    Eigenvalues to check for non semidefinite positiveness.
+
+  tol : float, optional
+    Negative eigenvalues above - tol are considered zero. If
+    tol is None, and w are `metric`'s eigenvalues, and eps is the
```
bellet: "and w are `metric`'s eigenvalues" is not needed

metric_learn/_util.py
```diff
@@ -324,7 +326,32 @@ def check_collapsed_pairs(pairs):
...
+    epsilon value for datatype of w, then tol is set to w.max() * len(w) * eps.
+
+  See Also
+  --------
+  np.linalg.matrix_rank for more details on the choice of tolerance (the same
+    strategy is applied here)
+  """
+  if tol is None:
+    tol = w.max() * len(w) * np.finfo(w.dtype).eps
+  if any(w[w < 0] < - tol):
```
bellet: `w<0` not needed

metric_learn/_util.py
```diff
@@ -337,18 +364,34 @@ def transformer_from_metric(metric):
...
   the matrix is itself).
 
+  Parameters
+  ----------
+  metric : symmetric `np.ndarray`, shape=(d x d)
+    The input metric, from which we want to extract a transformation matrix.
+
+  tol : positive float, optional
+    Eigenvalues of `metric` between 0 and - tol are considered zero. If tol is
+    None, and w are `metric`'s eigenvalues, and eps is the epsilon value for
```
bellet: same as above

test/test_transformer_metric_conversion.py
```diff
@@ -76,6 +81,102 @@ def test_mlkr(self):
...
+
+  def test_non_psd_raises(self):
+    """Checks that a non PSD matrix (i.e. with negative eigenvalues) will
+    raise an error when passed to transformer_from_metric"""
+    rng = np.random.RandomState(42)
+    D = np.diag([1, 5, 3, 4.2, -4, -2, 1])
+    P = ortho_group.rvs(7, random_state=rng)
+    M = P.dot(D).dot(P.T)
+    with pytest.raises(ValueError) as raised_error:
+      transformer_from_metric(M)
```
bellet: here you could also easily test that the error is raised in the diagonal case, for instance when calling `transformer_from_metric(D)`

metric_learn/_util.py
bellet: although it actually prevents that anything is done in case the provided `tol` is negative (which is not supposed to be the case according to the docstring but is not checked)

metric_learn/_util.py
wdevazelhes: That's right, done

metric_learn/_util.py
wdevazelhes: That's right, I'll let it be w < - tol and add a quick test to check that tol is positive

metric_learn/_util.py
wdevazelhes: How to explain how the tolerance is set here then without talking about the eigenvalues ? Though maybe an explanation talking about the highest eigenvalue could be better ? Like:
```
    Eigenvalues of `metric` between 0 and - tol are considered zero. If tol is
    None, and w_max is `metric`'s largest eigenvalue, and eps is the epsilon
    value for datatype of w, then tol is set to w_max * metric.shape[0] * eps.
```

test/test_transformer_metric_conversion.py
wdevazelhes: I agree, will do

wdevazelhes:
Thanks for the review @bellet, I addressed all the comments, there's just this one that needs to be resolved

metric_learn/_util.py
bellet: nevermind, indeed here we need to mention eigenvalues
the suggested reformulation looks good

bellet:
LGTM!

Issue #125, opened 2018-09-13
Do we want to preprocess data within our algorithms?


bellet:
Some algorithms include some data preprocessing: RCA and LMNN have a parameter to apply PCA to the data before fitting. I think a better practice (as done in sklearn) is to let the user apply preprocessing steps (such as normalization, dimensionality reduction) separately if needed, as it is more transparent.

If we do want to allow some preprocessing, when we should try to have at least a unified way to do this for all/most algorithms.

Issue #124, opened 2018-09-13
Cannot choose initial metric in the same way for Mahalanobis metric learning algorithms


bellet:
Depending on the Mahalanobis metric learning algorithm, the possible choices for the initial metric are not the same. For instance MMC and ITML accept an initial metric as parameter, NCA has none. The default initial metric also varies from an implementation to the other: for instance MKLR initializes by default with PCA.

There could be some unified interface to set the initial metric for Mahalanobis algorithms

PR #193, opened 2019-04-18
[MRG] Uniformize num_dims and add it for LMNN

### LINKED ISSUE #167: Dimensionality reduction for all algorithms
For all Mahalanobis metric learners, we should be able to reduce the dimension. For those that optimize the transformation matrix `L`, this can be done explicitely by setting the matrix at init to have shape (num_dims, n_features). For the others (that optimize the metric `M`), we could provide the user with `num_dims` which could be set to:
- a number k > n_features: in this case we would do the eigendecomposition of M and would only keep the k components with highest eigenvalues
- similar to scikit-learn's PCA, it could also be a value between 0 and 1 for say some threshold on the eigenvalues, or even a string, for some custom strategy (for instance the elbow rule)

This is the current state in the package: 
- All metric learners that use `transformer_from_metric` (Covariance, LSML, MMC, and SDML) do not have a `num_dims` argument
- All others optimize explicitely `L`, and have a `num_dims` argument (LFDA, MLKR, NCA, RCA) except LMNN, that could have one

Also, should we replace `num_dims` by `n_components`, like this is the case in scikit learn linear transformers ? This is also what we did for this PR on NCA in scikit-learn https://github.com/scikit-learn/scikit-learn/pull/10058

This is also related to #124, since we should check that in the case of a custom matrix for initializing the explicit transformer it is consistent with the desired dimension

wdevazelhes:
This solves #167 partially (only gives a num_dims argument to LMNN and ensures that the check is done uniformly between the algorithms)

### DIFF
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -404,3 +404,15 @@ def validate_vector(u, dtype=None):
   if u.ndim > 1:
     raise ValueError("Input vector should be 1-D.")
   return u
+
+
+def _check_num_dims(n_features, num_dims):
+  """Checks that num_dims is less that n_features and deal with the None
+  case"""
+  if num_dims is None:
+    dim = n_features
+  else:
+    if not 0 < num_dims <= n_features:
+      raise ValueError('Invalid num_dims, must be in [1, %d]' % n_features)
+    dim = num_dims
+  return dim
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -413,5 +413,5 @@ def _check_n_components(n_features, n_components):
   if n_components is None:
     return n_features
   if 0 < n_components <= n_features:
-    return n_features
+    return n_components
   raise ValueError('Invalid n_components, must be in [1, %d]' % n_features)
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -407,11 +407,11 @@ def validate_vector(u, dtype=None):
   return u
 
 
-def _check_num_dims(n_features, num_dims):
-  """Checks that num_dims is less than n_features and deal with the None
+def _check_n_components(n_features, n_components):
+  """Checks that n_components is less than n_features and deal with the None
   case"""
-  if num_dims is None:
+  if n_components is None:
     return n_features
-  if 0 < num_dims <= n_features:
+  if 0 < n_components <= n_features:
     return n_features
-  raise ValueError('Invalid num_dims, must be in [1, %d]' % n_features)
+  raise ValueError('Invalid n_components, must be in [1, %d]' % n_features)
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -408,12 +408,10 @@ def validate_vector(u, dtype=None):
 
 
 def _check_num_dims(n_features, num_dims):
-  """Checks that num_dims is less that n_features and deal with the None
+  """Checks that num_dims is less than n_features and deal with the None
   case"""
   if num_dims is None:
-    dim = n_features
-  else:
-    if not 0 < num_dims <= n_features:
-      raise ValueError('Invalid num_dims, must be in [1, %d]' % n_features)
-    dim = num_dims
-  return dim
+    return n_features
+  if 0 < num_dims <= n_features:
+    return n_features
+  raise ValueError('Invalid num_dims, must be in [1, %d]' % n_features)
--- a/test/test_fit_transform.py
+++ b/test/test_fit_transform.py
@@ -96,9 +96,7 @@ class TestFitTransform(unittest.TestCase):
     res_2 = lfda.fit_transform(self.X, self.y)
 
     # signs may be flipped, that's okay
-    if np.sign(res_1[0,0]) != np.sign(res_2[0,0]):
-        res_2 *= -1
-    assert_array_almost_equal(res_1, res_2)
+    assert_array_almost_equal(abs(res_1), abs(res_2))
 
   def test_rca_supervised(self):
     seed = np.random.RandomState(1234)
--- a/test/test_fit_transform.py
+++ b/test/test_fit_transform.py
@@ -88,11 +88,11 @@ class TestFitTransform(unittest.TestCase):
     assert_array_almost_equal(res_1, res_2)
 
   def test_lfda(self):
-    lfda = LFDA(k=2, num_dims=2)
+    lfda = LFDA(k=2, n_components=2)
     lfda.fit(self.X, self.y)
     res_1 = lfda.transform(self.X)
 
-    lfda = LFDA(k=2, num_dims=2)
+    lfda = LFDA(k=2, n_components=2)
     res_2 = lfda.fit_transform(self.X, self.y)
 
     # signs may be flipped, that's okay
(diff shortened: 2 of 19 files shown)


wdevazelhes:
Note: For LMNN, if we have shogun, it will do a pca if asked so, but I couldn't find how the dimension of pca is chosen... Also, maybe we'll not have to bother as soon as we have solved #124, because maybe we will chose that the 'pca' option triggers a scikit-learn PCA in both shogun/not shogun cases...

bellet:
Have you checked what is the pca option of Shogun for? It could be simply a preprocessing applied to the dataset, not an initialization. In which case we will set this option to false when solving #125

wdevazelhes:
> Have you checked what is the pca option of Shogun for? It could be simply a preprocessing applied to the dataset, not an initialization of the algorithm. In which case we will set this option to false when solving #125

Yes, according to this page: http://shogun.ml/notebook/latest/LMNN.html, (in cell 5), they say PCA is used to find the initial matrix

bellet:
> > Have you checked what is the pca option of Shogun for? It could be simply a preprocessing applied to the dataset, not an initialization of the algorithm. In which case we will set this option to false when solving #125
> 
> Yes, according to this page: http://shogun.ml/notebook/latest/LMNN.html, (in cell 5), they say PCA is used to find the initial matrix

Indeed. Looking at this, I do not think they reduce the dimension anyway:
https://shogun-list.shogun-toolbox.narkive.com/n6B3o0Mf/shogun-lmnn-pca-error
But I agree it will be safer to perform the PCA initialization ourselves and pass it to Shogun's train's method

metric_learn/_util.py
```diff
@@ -405,3 +405,15 @@ def validate_vector(u, dtype=None):
   if u.ndim > 1:
     raise ValueError("Input vector should be 1-D.")
   return u
+
+
+def _check_num_dims(n_features, num_dims):
+  """Checks that num_dims is less that n_features and deal with the None
+  case"""
+  if num_dims is None:
+    dim = n_features
```
perimosocordiae: As a style nitpick, I prefer early-return for cases like this:

```python
if num_dims is None:
  return n_features
if 0 < num_dims <= n_features:
  return num_dims
raise ValueError(...)
```

metric_learn/_util.py
```diff
@@ -405,3 +405,15 @@ def validate_vector(u, dtype=None):
   if u.ndim > 1:
     raise ValueError("Input vector should be 1-D.")
   return u
+
+
+def _check_num_dims(n_features, num_dims):
+  """Checks that num_dims is less that n_features and deal with the None
```
perimosocordiae: typo: "less than"

metric_learn/_util.py
```diff
@@ -405,3 +405,15 @@ def validate_vector(u, dtype=None):
...
+
+
+def _check_num_dims(n_features, num_dims):
+  """Checks that num_dims is less that n_features and deal with the None
+  case"""
+  if num_dims is None:
+    dim = n_features
+  else:
+    if not 0 < num_dims <= n_features:
+      raise ValueError('Invalid num_dims, must be in [1, %d]' % n_features)
```
perimosocordiae: Some existing code would only warn if `num_dims > n_features`. I think it's probably better to error out here, but we should keep in mind that this is technically a back-compat break.

metric_learn/_util.py
wdevazelhes: That's seems better I agree, done

metric_learn/_util.py

metric_learn/_util.py
wdevazelhes: That's true, maybe adding it in the changelog is enough ? Something like "for all the algorithms that have a parameter `num_dims`, it will now be checked to be between 1 and `n_features`, with `n_features` the number of dimensions of the input space" ?

wdevazelhes: I just added it to the release draft

wdevazelhes:
Also, I kept the name `num_dims` for now, but in scikit-learn's `NCA` we went for `n_components`, which respects more scikit-learn's conventions. It would be good to change the name wouldn't it ? If so I'll add deprecation warning for it

bellet:
I agree that `n_components` would be more intuitive, +1 for changing this before the 0.5.0 release with deprecation warning

wdevazelhes:
I investigated the travis fail and there seems to be an issue with some shapes mismatch for LMNN when doing dimensionality reduction. There seems to be a problem with the gradient: Looking into this line the gradient seems to be of the shape of an outer product of Xs (when it should be of the shape of L):

https://github.com/metric-learn/metric-learn/blob/d4badc8adc0a8ce6689dd9a95e89181efaf5ee24/metric_learn/lmnn.py#L196

Looking at the PR on LMNN in scikit-learn, there might be a dot product with L missing:
https://github.com/scikit-learn/scikit-learn/pull/8602/files#diff2914dd42441ed8b594e69b2fe04d23f5R762

Doing this dot product solves the problem
 
I'll look into it

wdevazelhes:
I just pushed a commit that warns the user with a ChangedBehaviorWarning when the init is not set, in the cases when the default init is not the same as it was before, let me know what you think, this happens in the following cases:
- lsml: it used the inverse covariance by default (instead of identity now)
- mmc: it used identity/10 (instead of just identity)
- mlkr: it used pca (instead of 'auto')
- nca: it used np.eye(X.shape[1])/(np.maximum(X.max(axis=0)-X.min(axis=0), EPS) (instead of 'auto')
- sdml: it used the inverse covariance by default (since use_cov was True by default), instead of identity

test/test_fit_transform.py
```diff
@@ -96,9 +96,7 @@ def test_lfda(self):
     res_2 = lfda.fit_transform(self.X, self.y)
 
     # signs may be flipped, that's okay
-    if np.sign(res_1[0,0]) != np.sign(res_2[0,0]):
-        res_2 *= -1
-    assert_array_almost_equal(res_1, res_2)
+    assert_array_almost_equal(abs(res_1), abs(res_2))
```
wdevazelhes: The tests were failing because of a sign error, I think it's because of the eigenvalue decomposition in LFDA, that can lead to different signs in the output. I did this modif because I thought that any sign switching of any column still keeps the result valid am I right (any symmetry wrt some of the axis are still valid right?) (Not just a global sign switching as before)? But maybe I should rather try to change the code to make LFDA deterministic ? What do you think @perimosocordiae @bellet ?

wdevazelhes: I just raised an issue for that (#211),  I guess maybe let's do like that for now and later fix this non deterministic problem ?

wdevazelhes:
If you agree with my last modifs (see comment https://github.com/metric-learn/metric-learn/pull/193#pullrequestreview-245994079), we should be ready to merge

test/test_fit_transform.py
perimosocordiae: The sign switch is expected, as each eigenvector may be negated independently of the others. The result is valid in the sense that the underlying optimization problem is still solved, and distances in the transformed space are equivalent. Individual points in the transformed space won't always be the same, though.

I don't think it's worth trying to enforce a deterministic result, but I could probably be persuaded otherwise.

bellet: Let's discuss it in the issue you opened

PR #194, opened 2019-04-23
[MRG] Remove preprocessing the data for RCA

### LINKED ISSUE #125: Do we want to preprocess data within our algorithms?

wdevazelhes:
Fixes #125 

This PR removes the preprocessing of the data in the algorithms. 

TODO: 

- [x] when the uniformized init is merged #195, ensure that the ChangedBehaviorMessage explains that in some cases 'pca' init is not used as before, and in order to use it one should set the 'init' parameter --> done in #195
- [x] add a small test to test the newly changed warning for RCA when the inner covariance matrix is singular


metric_learn/rca.py
```diff
@@ -98,26 +96,24 @@ def fit(self, X, chunks):
...
     """
+    if self.pca_comps != 'deprecated':
+      warnings.warn('"pca_comps" parameter is not used.'
+                    ' It has been deprecated in version 0.5.0 and will be'
+                    'removed in 0.6.0', DeprecationWarning)
+
     X = self._prepare_inputs(X, ensure_min_samples=2)
 
     # PCA projection to remove noise and redundant information.
-    if self.pca_comps is not None:
```
wdevazelhes: It appears that the data preprocessing (either pca or centering), was done only at fit time, so then the fit and transform were not coherent. But now it will be. 
I'll add a `ChangedBehaviourWarning` though to ensure users know what happens

metric_learn/rca.py
```diff
@@ -98,26 +96,24 @@ def fit(self, X, chunks):
...
-      pca = decomposition.PCA(n_components=self.pca_comps)
-      X_t = pca.fit_transform(X)
-      M_pca = pca.components_
```
wdevazelhes: Note that this code was giving a PCA initialization at the same time, so for now we'll remove it, but I think I'll do the PR about initialization before merging this PR into master, and then we can merge it into this PR to keep the same possibility of initialization with PCA

bellet:
There is still a warning that mentions pca_comps to update. if the matrix is low-rank (due to high dimension compared to the number of points) , we can suggest the user to use a dimensionality reduction technique such as PCA as preprocessing. Unless we have a better idea

wdevazelhes:
I think this PR is ready to merge as soon as the initialization PR is merged (#195)

wdevazelhes:
@bellet @perimosocordiae I guess this PR is ready to merge now

metric_learn/rca.py
```diff
@@ -105,25 +105,30 @@ def fit(self, X, chunks):
...
+
     X, chunks = self._prepare_inputs(X, chunks, ensure_min_samples=2)
 
-    # PCA projection to remove noise and redundant information.
-    if self.pca_comps is not None:
-      pca = decomposition.PCA(n_components=self.pca_comps)
-      X_t = pca.fit_transform(X)
-      M_pca = pca.components_
-    else:
-      X_t = X - X.mean(axis=0)
```
bellet: why is this centering step gone?

metric_learn/rca.py
```diff
@@ -105,25 +105,30 @@ def fit(self, X, chunks):
...
-      M_pca = None
+    warnings.warn("RCA will no longer be trained on a preprocessed version "
```
bellet: Do we need a ChangedBehaviorWarning? The default behavior was `pca_comps=None`
Maybe the information in this warning should rather go to the deprecation warning?

metric_learn/rca.py
wdevazelhes: I guess because we should remove any pre-processing step, but I agree I didn't talk about it at all, maybe we should keep the `ChangedBehaviorWarning` message below, but rather replace "no longer trained on a preprocessed version" by "no longer trained on centered data by default", and encourage to use a "StandardScaler" if needed ?

metric_learn/rca.py
wdevazelhes: see above

metric_learn/rca.py
bellet: Fair enough (I double-checked and this centering is not part of standard RCA)
Maybe keep ChangedBehaviorWarning but change to "no longer center the data before training RCA" (no need to mention scaler I think)
And in the deprecation warning, add the fact that PCA preprocessing should now be done by the user

Finally, have you checked the influence of removing the centering step on the examples?

bellet:
I think addressing https://github.com/metric-learn/metric-learn/pull/194#discussion_r292387277 and then we can merge

wdevazelhes:
> I think addressing [#194 (comment)](https://github.com/metric-learn/metric-learn/pull/194#discussion_r292387277) and then we can merge

Done, I'll just check quickly what it changes on iris for instance, as you recommended for check-proof

wdevazelhes:
So I guess we're OK to merge now

PR #195, opened 2019-04-23
[MRG] Uniformize initialization for all algorithms

### LINKED ISSUE #124: Cannot choose initial metric in the same way for Mahalanobis metric learning algorithms

### LINKED ISSUE #202: LSML fails on some classical inputs
while finalizing #180, I encountered problems with `LSML`: examples that shouldn't be problems fail: here are a snippets that failed

```python
import numpy as np
from metric_learn import LSML_Supervised

X, Y = make_classification(n_classes=3, n_clusters_per_class=2,
                           n_informative=3, random_state=42, class_sep=4.,
                           n_features=5)

lsml = LSML_Supervised(num_constraints=200, tol=0.0001,
                       max_iter=10000, verbose=True)
X_lsml = lsml.fit_transform(X, Y)
```
The error returned is: `ValueError: The input metric should be symmetric.`

wdevazelhes:
Fixes #124 
Fixes #202 

TODO: 
- [x] Uniformize all algorithms that learn transformer except MLKR
- [x] Try to refactor MLKR to benefit as much as possible from the already made changed
- [x] Warning: do the tests for all algos that learn a transformer, not only for the supervised ones (currently not the case) (currently it does not change anything but it could, and semantically that's how the test should be)
- [x] Uniformize all the algorithms that learn a mahalanobis matrix
- <strike>[ ] Remove the remaining num_dims if there are, after merging with #167 </strike> we should first merge this PR
- [x] Check whether some algos can have both an init and a prior


wdevazelhes:
I am thinking for algorithms that learn a transformation `L` we could do the same initialization as in scikit-learn's NCA: https://scikit-learn.org/dev/modules/generated/sklearn.neighbors.NeighborhoodComponentsAnalysis.html)
But the problem is that for algorithms that learn the metric matrix, they don't have a `n_components` arguments, I was thinking we could do an initialization where we would take L=(lda) with n_components=None (by the way PCA in this case would'nt make sense because it would just rotate the data isn't it ? (if we assume it's already centered)), and have A_0 = L**T.dot(L), what do you think ? But then what to do for choosing what to do for 'auto' ? Maybe lda then ?

wdevazelhes:
By the way I was thinking it could be cool to have an option to give any estimator as init, as long as it has an `components_` attribute after fitting, so that it would allow to use `TruncatedSVD` for instance, or any other custom estimator. But this is out of the scope of this PR I guess

bellet:
For algorithms learning the transformation, it seems a good idea to re-use as much of the code you wrote for NCA. These algorithms are typically nonconvex so a good initialization can make a large difference. I remember that some benchmarks you had done for NCA seemed to show that LDA init was good, so let's go for that.

For algorithms learning the metric matrix: these are typically convex so the init is not as important (it mostly affects convergence speed). We also have to be careful with initialization that would lead to singular matrices (such as LDA) which could potentially make some solvers fail. My 2 cents: I would focus on 'identity', 'covariance' (corresponding to the *inverse* covariance), 'random' and numpy array, with the default to 'identity' unless some algorithms have a more natural init (e.g., SDML could be initialized to 'covariance'). I would not introduce LDA without a proper benchmark showing some concrete benefits on convergence speed and ensuring that it does not cause any issue.

bellet:
> By the way I was thinking it could be cool to have an option to give any estimator as init, as long as it has an `components_` attribute after fitting, so that it would allow to use `TruncatedSVD` for instance, or any other custom estimator. But this is out of the scope of this PR I guess

I guess you mean that this estimator would be fitted within our own fit method? Could be convenient indeed (and PCA/LDA init could be obtained that way) but I think not a high priority and there may be some implementation subtleties. Better keep this for later

test/test_mahalanobis_mixin.py
```diff
@@ -138,9 +139,6 @@ def test_embed_dim(estimator, build_dataset):
   assert str(raised_error.value) == err_msg
   # we test that the shape is also OK when doing dimensionality reduction
   if type(model).__name__ in {'LFDA', 'MLKR', 'NCA', 'RCA'}:
-    # TODO:
```
wdevazelhes: This will be changed (for good) in #193 (see https://github.com/metric-learn/metric-learn/pull/193/files#diff-2d2fa8dfd3b4bf0d149166d5e3efa526)

test/metric_learn_test.py
```diff
@@ -475,9 +475,7 @@ def test_singleton_class(self):
 
       EPS = np.finfo(float).eps
       A = np.zeros((X.shape[1], X.shape[1]))
-      np.fill_diagonal(A,
```
wdevazelhes: I have to double check here I don't think that's what I should have done

test/metric_learn_test.py
```diff
@@ -475,9 +475,7 @@ def test_singleton_class(self):
...
-                       1. / (np.maximum(X.max(axis=0) - X.min(axis=0), EPS)))
-      nca = NCA(max_iter=30, num_dims=X.shape[1])
+      nca = NCA(init=A, max_iter=30, num_dims=X.shape[1])
```
wdevazelhes: Same here

test/metric_learn_test.py
```diff
@@ -488,9 +486,7 @@ def test_one_class(self):
       y = self.iris_labels[self.iris_labels == 0]
       EPS = np.finfo(float).eps
       A = np.zeros((X.shape[1], X.shape[1]))
-      np.fill_diagonal(A,
-                       1. / (np.maximum(X.max(axis=0) - X.min(axis=0), EPS)))
-      nca = NCA(max_iter=30, num_dims=X.shape[1])
```
wdevazelhes: Same here

test/metric_learn_test.py
```diff
@@ -646,8 +646,9 @@ def test_iris(self):
     c, d = np.nonzero(np.triu(~mask, k=1))
 
     # Full metric
-    mmc = MMC(convergence_threshold=0.01)
-    mmc.fit(*wrap_pairs(self.iris_points, [a,b,c,d]))
+    n_features = self.iris_points.shape[1]
+    mmc = MMC(convergence_threshold=0.01, init=np.eye(n_features) / 10)
```
wdevazelhes: The previous default init was identity divided by 10

test/metric_learn_test.py
```diff
@@ -366,7 +366,7 @@ def test_sdml_works_on_non_spd_pb_with_skggm(self):
     X, y = load_iris(return_X_y=True)
     sdml = SDML_Supervised(balance_param=0.5, sparsity_param=0.01,
                            init='covariance')
-    sdml.fit(X, y)
+    sdml.fit(X, y, random_state=np.random.RandomState(42))
```
wdevazelhes: Fixing this random state makes the test pass (otherwise there is the error `RuntimeError: There was a problem in SDML when using skggm's graphical lasso solver.`). I tried to understand why but I couldn't even reproduce the "no-bug" from this travis build for instance (https://travis-ci.org/metric-learn/metric-learn/jobs/527783562), locally, so I must be doing something wrong (I created a virtualenv with the same versions of python and packages but it still fails so I don't know what could go wrong)

test/test_mahalanobis_mixin.py
```diff
@@ -194,7 +194,7 @@ def test_get_metric_equivalent_to_explicit_mahalanobis(estimator,
   a, b = (rng.randn(n_features), rng.randn(n_features))
   expected_dist = mahalanobis(a[None], b[None],
                               VI=model.get_mahalanobis_matrix())
-  assert_allclose(metric(a, b), expected_dist, rtol=1e-15)
+  assert_allclose(metric(a, b), expected_dist, rtol=1e-13)
```
wdevazelhes: This is to fix the test that didn't pass in this build: https://travis-ci.org/metric-learn/metric-learn/builds/531766687?utm_source=github_status&utm_medium=notification 
I didn't manage to reproduce the bug locally either (it passes for me, with the same versions of packages than travis, so I don't know what may be going wrong...)
But I guess 1e-13 is still a reasonable tolerance

metric_learn/lsml.py
```diff
@@ -23,36 +23,70 @@
...
+    init : string or numpy array, optional (default='identity')
+         Initialization of the linear transformation. Possible options are
+         'identity', 'covariance', 'random', and a numpy array of shape
+         (n_features, n_features).
+
+         'identity'
+            An identity matrix of shape (n_features, n_features).
+
+         'covariance'
+            The inverse covariance matrix.
```
wdevazelhes: TODO: replace by pseudo-inverse

metric_learn/sdml.py
```diff
@@ -112,7 +112,11 @@ def _fit(self, pairs, y):
                                                   return_inverse=True)
     diff = pairs[:, 0] - pairs[:, 1]
     loss_matrix = (diff.T * y).dot(diff)
-    emp_cov = prior_inv + self.balance_param * loss_matrix
+    emp_cov = (prior_inv + self.balance_param * loss_matrix +
+               # We add a small value on the diagonal in case the
+               # emp_cov matrix is singular (see
+               # #https://github.com/metric-learn/metric-learn/issues/202)
+               np.eye(diff.shape[1]) * 1e-10)
```
wdevazelhes: This value makes the test pass, I think it's a value small enough to not interfere with the other coeffs, what do you think ?

wdevazelhes: I'll revert it given the strategy we'll adopt (see https://github.com/metric-learn/metric-learn/pull/203#pullrequestreview-237083152)

wdevazelhes:
I'll revert the previous commits and adopt the strategy discussed with @bellet in https://github.com/metric-learn/metric-learn/pull/203#pullrequestreview-237083152

bellet:
I think we should definitely check that the init is PSD (i.e., defines a valid Mahalanobis distance) for all algorithms, and return an error otherwise. For the covariance initialization, I do not see any problem with using the pseudo-inverse instead of the inverse as this will lead to a valid PSD matrix as long as the covariance matrix is PSD.

For specific algorithms which are supposed to enforce strict PD (like the LogDet-based ones such as LSML and ITML), we need to check that the init is PD and return an error otherwise.

Note that I think we should make a distinction between 'prior' and 'init': 'prior' corresponds to a regularization (there is a term in the objective which tries to keep the solution close to the prior and thus affects the solution), while 'init" is simply an initialization to iterative solvers (for instance, it does not change the solution when the optimization problem is convex). So I would suggest to keep the two different names, but allow them to be set in the same way (identity, inverse cov, etc).

wdevazelhes:
Thanks for the clarifications @bellet 
I just pushed something that I was doing yesterday, but I am currently updating it accordingly

metric_learn/itml.py
```diff
@@ -32,7 +32,8 @@ class _BaseITML(MahalanobisMixin):
```
wdevazelhes: test

metric_learn/itml.py
```diff
@@ -23,7 +23,7 @@
 from sklearn.base import TransformerMixin
 from .base_metric import _PairsClassifierMixin, MahalanobisMixin
 from .constraints import Constraints, wrap_pairs
-from ._util import vector_norm, transformer_from_metric
+from ._util import transformer_from_metric, _initialize_metric_mahalanobis
 
 
 class _BaseITML(MahalanobisMixin):
```
wdevazelhes: test

metric_learn/itml.py
```diff
@@ -87,10 +87,8 @@ def __init__(self, gamma=1., max_iter=1000, convergence_threshold=1e-3,
     self.gamma = gamma
     self.max_iter = max_iter
     self.convergence_threshold = convergence_threshold
-    self.init = init  # explain that it is good to keep the scale with
-    # the bounds
-    # TODO: see for other inits how it behave wrt the bound
-    self.A0 = A0  # TODO: deprecate
+    self.init = init
```
wdevazelhes: @bellet if I remember correctly we were saying that since ITML computes bounds, it is needed that the init is coherent with these bounds, and with the default identity it should be OK, but otherwise it would not, so here are some of the things that could be done (I don't remember if we said other things):
- we could compute the bounds after the init, but maybe if the init is very bad, the bounds would be bad too ? 
- we could rescale the init so that we have the maximum of constraints that satisfy the bounds (this avoids to wait for gradient iterations to converge to this)
What do you think ?

test/metric_learn_test.py
wdevazelhes: done, I updated with a random matrix

test/metric_learn_test.py
wdevazelhes: done

wdevazelhes: done

test/metric_learn_test.py
wdevazelhes: done

metric_learn/lsml.py
wdevazelhes: Done: No, because it will be the inverse (return an error if not definite)

wdevazelhes: no, it's an inverse, because if cov is not definite we return an error

wdevazelhes:
I guess the PR is now ready for review, and there are the following comments that need to be addressed: 
- What to do for the scaling of the init in ITML ? (see https://github.com/metric-learn/metric-learn/pull/195#discussion_r285181411)
- Do we want to allow both prior and init at the same time for now,  for LSML for instance ? The init would be the init of the gradient ? Or having the default init is fine ?  If we want that, maybe in another PR ? 
- I modified NCA's inits with respect to what it was before, should I add a ChangedBehaviourWarning for that ? Here is the diff:

```diff
-  A = np.zeros((num_dims, d))
- np.fill_diagonal(A, 1./(np.maximum(X.max(axis=0)-X.min(axis=0), EPS)))

+ A = _initialize_transformer(num_dims, X, labels, self.init, self.verbose)
```

bellet:
>     * What to do for the scaling of the init in ITML ? (see [#195 (comment)](https://github.com/metric-learn/metric-learn/pull/195#discussion_r285181411))

Identity by default is fine. But you should switch to `prior` instead of `init` for ITML

>     * Do we want to allow both prior and init at the same time for now,  for LSML for instance ? The init would be the init of the gradient ? Or having the default init is fine ?  If we want that, maybe in another PR ?

Having both, with `init` set to `prior` by default would be nice

>     * I modified NCA's inits with respect to what it was before, should I add a ChangedBehaviourWarning for that ? Here is the diff:
> 
> 
> ```diff
> -  A = np.zeros((num_dims, d))
> - np.fill_diagonal(A, 1./(np.maximum(X.max(axis=0)-X.min(axis=0), EPS)))
> 
> + A = _initialize_transformer(num_dims, X, labels, self.init, self.verbose)
> ```

Generally I think it is a good idea to add such warning everywhere the default behavior (i.e., type of init) changed

metric_learn/_util.py
```diff
@@ -411,3 +426,298 @@ def validate_vector(u, dtype=None):
...
+  random_state = check_random_state(random_state)
+  transformation = init
+  if isinstance(init, np.ndarray):
+    pass
+  else:
+    n_samples = input.shape[0]
+    if init == 'auto':
+      if has_classes:
+        n_classes = len(np.unique(y))
+      if (has_classes and num_dims <= min(n_features, n_classes - 1)):
```
nvauquie: You can remove these parentheses

metric_learn/_util.py
```diff
@@ -411,3 +426,298 @@ def validate_vector(u, dtype=None):
...
+    The initial matrix to use M, and its inverse if `return_inverse=True`.
+  """
+  n_features = input.shape[-1]
+  if isinstance(init, np.ndarray):
+    # we copy the array, so that if we update the metric, we don't want to
+    # update the init
+    init = check_array(init, copy=True)
+
+    # Assert that init.shape[1] = n_features
+    if (init.shape) != (n_features,) * 2:
```
nvauquie: You can remove the parentheses around `init.shape`

wdevazelhes:
I just pushed a commit that warns the user with a `ChangedBehaviorWarning` when the init is not set, in the cases when the default init is not the same as it was before, let me know what you think, this happens in the following cases:
- lsml: it used the inverse covariance by default (instead of identity now)
- mmc: it used identity/10 (instead of just identity)
- mlkr: it used pca (instead of 'auto')
- nca: it used `np.eye(X.shape[1])/(np.maximum(X.max(axis=0)-X.min(axis=0), EPS)` (instead of 'auto')
- sdml: it used the inverse covariance by default (since use_cov was True by default), instead of identity

metric_learn/_util.py
```diff
@@ -411,3 +426,298 @@ def validate_vector(u, dtype=None):
...
+    The input labels (or not if there are no labels).
+
+  init : string or numpy array, optional (default='auto')
+      Initialization of the linear transformation. Possible options are
+      'auto', 'pca', 'lda', 'identity', 'random', and a numpy array of shape
+      (n_features_a, n_features_b).
+
+      'auto'
+          Depending on ``num_dims``, the most reasonable initialization will
+          be chosen. If ``num_dims <= n_classes`` we use 'lda' (if possible,
```
bellet: remove "if possible, "?

metric_learn/_util.py
```diff
@@ -411,3 +426,298 @@ def validate_vector(u, dtype=None):
...
+          see the description of 'lda' init), as it uses labels information.
+          If not, but ``num_dims < min(n_features, n_samples)``, we use
+          'pca', as it projects data in meaningful directions (those of
```
bellet: in --> onto

metric_learn/_util.py
```diff
@@ -411,3 +426,298 @@ def validate_vector(u, dtype=None):
...
+      'lda'
+          ``min(num_dims, n_classes)`` most discriminative
+          components of the inputs passed to :meth:`fit` will be used to
+          initialize the transformation. (If ``num_dims > n_classes``,
+          the rest of the components will be zero.) (See
+          `sklearn.discriminant_analysis.LinearDiscriminantAnalysis`).
+          This initialization is possible only if `has_classes == True`.
+
+      'identity'
+          If ``num_dims`` is strictly smaller than the
```
bellet: although it should be obvious, maybe start by clearly saying that that this uses identity matrix

metric_learn/_util.py
```diff
@@ -411,3 +426,298 @@ def validate_vector(u, dtype=None):
...
+        transformation = lda.scalings_.T[:num_dims]
+      if verbose:
+        print('done in {:5.2f}s'.format(time.time() - init_time))
+  return transformation
+
+
+def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
+                                   return_inverse=False, strict_pd=False,
+                                   matrix_name='matrix'):
+  """Returns a standard mahalanobis matrix that can be used as a prior or an
```
bellet: a standard mahalanobis --> a PSD
and add "as a prior or an initialization for the Mahalanobis distance"

metric_learn/_util.py
```diff
@@ -411,3 +426,298 @@ def validate_vector(u, dtype=None):
...
+  initialization
+
+  Parameters
+  ----------
+  input : array-like
+    The input samples (can be tuples or regular samples).
+
+  init : string or numpy array, optional (default='identity')
+         Initialization of the linear transformation. Possible options are
```
bellet: not linear transformation

metric_learn/_util.py
```diff
@@ -411,3 +426,298 @@ def validate_vector(u, dtype=None):
...
+
+         'identity'
+            An identity matrix of shape (n_features, n_features).
+
+         'covariance'
+            The (pseudo-)inverse covariance matrix (raises an error if the
+            covariance matrix is not definite and `strict_pd == True`)
+
+         'random'
+             The initial transformation will be a random PD matrix of shape
```
bellet: not transformation

metric_learn/base_metric.py
```diff
@@ -288,7 +288,7 @@ def get_mahalanobis_matrix(self):
 
     Returns
     -------
-    M : `numpy.ndarray`, shape=(n_components, n_features)
+    M : `numpy.ndarray`, shape=(num_dims, n_features)
```
bellet: shouldn't this always be `(n_features, n_features)`?

metric_learn/itml.py
```diff
@@ -44,24 +45,59 @@ def __init__(self, gamma=1., max_iter=1000, convergence_threshold=1e-3,
 
     convergence_threshold : float, optional
 
-    A0 : (d x d) matrix, optional
-        initial regularization matrix, defaults to identity
+    prior : string or numpy array, optional (default='identity')
+         Initialization of the linear transformation. Possible options are
```
bellet: not initialization, and not linear transformation

metric_learn/itml.py
```diff
@@ -249,17 +286,46 @@ def __init__(self, gamma=1., max_iter=1000, convergence_threshold=1e-3,
           `bounds` was deprecated in version 0.5.0 and will
           be removed in 0.6.0. Set `bounds` at fit time instead :
           `itml_supervised.fit(X, y, bounds=...)`
-    A0 : (d x d) matrix, optional
-        initial regularization matrix, defaults to identity
+
+    prior : string or numpy array, optional (default='identity')
+         Initialization of the linear transformation. Possible options are
```
bellet: same here

metric_learn/lsml.py
```diff
@@ -20,36 +20,62 @@
...
+         positive definite (PD).
+
+         'identity'
+            An identity matrix of shape (n_features, n_features).
+
+         'covariance'
+            The inverse covariance matrix.
+
+         'random'
+            The initial transformation will be a random PD matrix of shape
```
bellet: again

metric_learn/lsml.py
```diff
@@ -202,8 +231,27 @@ def __init__(self, tol=1e-3, max_iter=1000, prior=None,
...
```
bellet: again

metric_learn/lsml.py
```diff
@@ -217,9 +265,14 @@ def __init__(self, tol=1e-3, max_iter=1000, prior=None,
     preprocessor : array-like, shape=(n_samples, n_features) or callable
         The preprocessor to call to get tuples from indices. If array-like,
         tuples will be formed like this: X[indices].
+    random_state : int or numpy.RandomState or None, optional (default=None)
+        A pseudo random number generator object or a seed for it if int. If
+        ``init='random'``, ``random_state`` is used to initialize the random
+        transformation.
```
bellet: again

metric_learn/lsml.py
```diff
@@ -20,36 +20,62 @@
...
-        guess at a metric [default: inv(covariance(X))]
     verbose : bool, optional
         if True, prints information while learning
     preprocessor : array-like, shape=(n_samples, n_features) or callable
         The preprocessor to call to get tuples from indices. If array-like,
         tuples will be formed like this: X[indices].
+    random_state : int or numpy.RandomState or None, optional (default=None)
+        A pseudo random number generator object or a seed for it if int. If
+        ``init='random'``, ``random_state`` is used to initialize the random
+        transformation.
```
bellet: again

metric_learn/mmc.py
```diff
@@ -34,20 +35,49 @@ class _BaseMMC(MahalanobisMixin):
...
     Parameters
     ----------
     max_iter : int, optional
     max_proj : int, optional
     convergence_threshold : float, optional
-    A0 : (d x d) matrix, optional
-        initial metric, defaults to identity
-        only the main diagonal is taken if `diagonal == True`
+    init : string or numpy array, optional (default='identity')
+         Initialization of the linear transformation. Possible options are
```
bellet: again

metric_learn/mmc.py
```diff
@@ -34,20 +35,49 @@ class _BaseMMC(MahalanobisMixin):
...
+         (n_features, n_features).
+
+         'identity'
+            An identity matrix of shape (n_features, n_features).
+
+        'covariance'
+            The (pseudo-)inverse of the covariance matrix.
+
+         'random'
+            The initial transformation will be a random SPD matrix of shape
```
bellet: again

metric_learn/mmc.py
```diff
@@ -56,29 +86,42 @@ def __init__(self, max_iter=100, max_proj=10000, convergence_threshold=1e-3,
     preprocessor : array-like, shape=(n_samples, n_features) or callable
         The preprocessor to call to get tuples from indices. If array-like,
         tuples will be gotten like this: X[indices].
+    random_state : int or numpy.RandomState or None, optional (default=None)
+        A pseudo random number generator object or a seed for it if int. If
+        ``init='random'``, ``random_state`` is used to initialize the random
+        transformation.
```
bellet: again

metric_learn/mmc.py
```diff
@@ -432,9 +475,36 @@ def __init__(self, max_iter=100, max_proj=10000, convergence_threshold=1e-6,
          be removed in 0.6.0.
     num_constraints: int, optional
         number of constraints to generate
-    A0 : (d x d) matrix, optional
-        initial metric, defaults to identity
-        only the main diagonal is taken if `diagonal == True`
+    init : string or numpy array, optional (default='identity')
+         Initialization of the linear transformation. Possible options are
```
bellet: again

metric_learn/mmc.py
```diff
@@ -432,9 +475,36 @@ def __init__(self, max_iter=100, max_proj=10000, convergence_threshold=1e-6,
...
+         (n_features, n_features).
+
+         'identity'
+             An identity matrix of shape (n_features, n_features).
+
+         'covariance'
+             The (pseudo-)inverse of the covariance matrix.
+
+         'random'
+             The initial transformation will be a random SPD matrix of shape
```
bellet: again

metric_learn/mmc.py
```diff
@@ -446,11 +516,16 @@ def __init__(self, max_iter=100, max_proj=10000, convergence_threshold=1e-6,
     preprocessor : array-like, shape=(n_samples, n_features) or callable
         The preprocessor to call to get tuples from indices. If array-like,
         tuples will be formed like this: X[indices].
+    random_state : int or numpy.RandomState or None, optional (default=None)
+        A pseudo random number generator object or a seed for it if int. If
+        ``init='random'``, ``random_state`` is used to initialize the random
+        transformation.
```
bellet: again

metric_learn/sdml.py
```diff
@@ -46,23 +47,58 @@ def __init__(self, balance_param=0.5, sparsity_param=0.01, use_cov=True,
...
+         positive definite (PD).
+
+         'identity'
+            An identity matrix of shape (n_features, n_features).
+
+         'covariance'
+            The inverse covariance matrix.
+
+         'random'
+            The initial transformation will be a random PD matrix of shape
```
bellet: again

metric_learn/sdml.py
```diff
@@ -73,11 +109,19 @@ def _fit(self, pairs, y):
...
-      X = np.vstack({tuple(row) for row in pairs.reshape(-1, pairs.shape[2])})
-      prior_inv = np.atleast_2d(np.cov(X, rowvar=False))
-    else:
-      prior_inv = np.identity(pairs.shape[2])
+    # if the prior is the default (identity), we raise a warning just in case
+    if self.prior == 'identity':
+      msg = ("Warning, as of version 0.5.0, the default prior is now "
+             "'identity', instead of 'covariance'. If you still want to use "
+             "the inverse of the covariance matrix as a prior, "
+             "set 'prior'=='covariance' (it was the default in previous "
```
bellet: the content of the parenthesis "it was the default..." is redundant, you can remove it

metric_learn/mmc.py
```diff
@@ -56,29 +86,42 @@ def __init__(self, max_iter=100, max_proj=10000, convergence_threshold=1e-3,
...
-    # init metric
-    if self.A0 is None:
-      self.A_ = np.identity(pairs.shape[2])
-      if not self.diagonal:
-        # Don't know why division by 10... it's in the original code
-        # and seems to affect the overall scale of the learned metric.
-        self.A_ /= 10.0
-    else:
-      self.A_ = check_array(self.A0)
+    msg = ("Warning, as of version 0.5.0, the default prior is now "
```
bellet: init instead of prior

metric_learn/_util.py

metric_learn/_util.py

metric_learn/_util.py
```diff
@@ -335,28 +341,37 @@ def check_collapsed_pairs(pairs):
 def _check_sdp_from_eigen(w, tol=None):
   """Checks if some of the eigenvalues given are negative, up to a tolerance
   level, with a default value of the tolerance depending on the eigenvalues.
+  It also returns whether the matrix is definite.
```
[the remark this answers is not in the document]
wdevazelhes: That's right, done, yes I used the same tolerance as for checking if the eigenvalues are negative (the tolerance is a sort of threshold to detect zeros eigenvalues that can be used in any side of 0, + or -). Do you think I should allow to set a different threshold for returning if the matrix is PSD ?

metric_learn/_util.py
wdevazelhes: Yes, done

metric_learn/_util.py
wdevazelhes: thanks, done

metric_learn/_util.py
wdevazelhes: I agree, done

metric_learn/_util.py
wdevazelhes: thanks, done

metric_learn/_util.py
wdevazelhes: thanks, done, I've replaced it by "Specification for the matrix to initialize"

metric_learn/_util.py
wdevazelhes: thanks, done

metric_learn/base_metric.py
wdevazelhes: that's right, I didn't pay attention, thanks

metric_learn/itml.py
wdevazelhes: thanks, done, I've replaced it by "Initialization of the Mahalanobis matrix"

metric_learn/itml.py
wdevazelhes: done

metric_learn/lsml.py
wdevazelhes: done

metric_learn/lsml.py
wdevazelhes: done

metric_learn/lsml.py
wdevazelhes: done

metric_learn/lsml.py
wdevazelhes: done

metric_learn/mmc.py
wdevazelhes: done

metric_learn/mmc.py
wdevazelhes: done

metric_learn/mmc.py
wdevazelhes: done

metric_learn/mmc.py
wdevazelhes: doçne

metric_learn/mmc.py
wdevazelhes: done

metric_learn/sdml.py
wdevazelhes: done

metric_learn/sdml.py
wdevazelhes: that's right, done

metric_learn/mmc.py
wdevazelhes: thanks, done

wdevazelhes:
@bellet Regarding the init that default as prior if possible, I think I have to parse a bit the algorithms to understand which part in the computations is the actual prior and which part is the initial matrix for the iterative algorithms: I'd be more comfortable in making a guess at it in a new PR so that it's easier to comment and talk about it, without polluting this PR, what do you think we merge this PR first and do this in a next PR ? Also the PR has become quite big now so I guess it would improve the readability of the next changes

wdevazelhes:
If you agree, I addressed all your comments, so I guess we can merge this PR

metric_learn/_util.py
```diff
@@ -335,28 +341,38 @@ def check_collapsed_pairs(pairs):
...
     strategy is applied here)
   """
   if tol is None:
-    tol = w.max() * len(w) * np.finfo(w.dtype).eps
+    tol = np.abs(w).max() * len(w) * np.finfo(w.dtype).eps
   if tol < 0:
     raise ValueError("tol should be positive.")
   if any(w < - tol):
-      raise ValueError("Matrix is not positive semidefinite (PSD).")
+    raise NonPSDError
```
perimosocordiae: You're missing parens here: `raise NonPSDError()`

metric_learn/_util.py
```diff
@@ -411,3 +427,301 @@ def validate_vector(u, dtype=None):
...
+  elif init not in authorized_inits:
+    raise ValueError(
+        "`init` must be '{}' "
+        "or a numpy array of shape (num_dims, n_features)."
+        .format("', '".join(authorized_inits)))
+
+  random_state = check_random_state(random_state)
+  if isinstance(init, np.ndarray):
+    return init
+  else:
```
perimosocordiae: Especially for long functions with lots of nesting like this one, I prefer the "no else" style:

```python
if condition:
  return foo
# everything else at original indent
```

metric_learn/_util.py
```diff
@@ -411,3 +427,301 @@ def validate_vector(u, dtype=None):
...
+    n_samples = input.shape[0]
+    if init == 'auto':
```
perimosocordiae: This might be simpler to test if we broke out pieces into standalone functions. For example, the "auto-select" logic could be it's own function.

metric_learn/lmnn.py
```diff
@@ -19,18 +19,61 @@
 from six.moves import xrange
 from sklearn.metrics import euclidean_distances
 from sklearn.base import TransformerMixin
+
+from metric_learn._util import _initialize_transformer
```
perimosocordiae: `from ._util import ...`

metric_learn/lmnn.py
```diff
@@ -102,6 +156,8 @@ def fit(self, X, y):
         self._loss_grad(X, L, dfG, impostors, 1, k, reg, target_neighbors, df,
                         a1, a2))
 
+    it = 1  # we already made one iteration
```
perimosocordiae: This seems like a no-op line. Maybe just update the "main loop" comment?

metric_learn/lsml.py
```diff
@@ -66,14 +92,18 @@ def _fit(self, quadruplets, weights=None):
...
-      M = self.prior
-      prior_inv = np.linalg.inv(self.prior)
+    # if the prior is the default (identity), we raise a warning just in case
+    if self.prior == 'identity':
+      msg = ("Warning, as of version 0.5.0, the default prior is now "
+             "'identity', instead of 'covariance'. If you still want to use "
+             "the inverse of the covariance matrix as a prior, "
+             "set 'prior'=='covariance'. This warning will disappear in "
+             "v0.6.0.")
+      warnings.warn(msg, ChangedBehaviorWarning)
```
perimosocordiae: This warning will be annoying for users that want to use identity initialization intentionally. We could instead keep using `None` as the default and only warn when the user hasn't set the prior explicitly.

metric_learn/mlkr.py
```diff
@@ -82,24 +122,33 @@ def fit(self, X, y):
...
-      elif A.shape != (m, d):
-          raise ValueError('A0 needs shape (%d,%d) but got %s' % (
-              m, d, A.shape))
+      # if the init is the default (identity), we raise a warning just in case
+      if self.init == 'auto':
+        msg = ("Warning, as of version 0.5.0, the default init is now "
+               "'auto', instead of 'pca'. If you still want to use "
+               "PCA as an init, set 'init'=='pca'. This warning will "
+               "disappear in v0.6.0.")
+        warnings.warn(msg, ChangedBehaviorWarning)
```
perimosocordiae: Same issue here with intentional / explicit parameter vs default value.

metric_learn/_util.py
wdevazelhes: I tried with a command line python 2 and 3 and it worked without the parenthesis, is it a best practice to put the parenthesis ?

metric_learn/_util.py
wdevazelhes: That's right, it's better, done

metric_learn/_util.py
wdevazelhes: I agree, done, and tested the function

metric_learn/lmnn.py
wdevazelhes: Thanks, done

metric_learn/lmnn.py
wdevazelhes: I am not sure I understand, I think the `it=1` still useful for coherence, if one puts `max_iter=1`, it would break otherwise (variable not defined)

metric_learn/lsml.py
wdevazelhes: I agree, let me know what you think of the new error messages (says that init was not set because is None, and that in the next version it will be set to 'identity' for instance, and that the warning will disappear)

metric_learn/mlkr.py
wdevazelhes: Agreed, done

wdevazelhes:
@perimosocordiae thanks for your review ! I addressed all your comments (except https://github.com/metric-learn/metric-learn/pull/195#discussion_r291175951)
If you all agree, I think it's good to merge

bellet:
> @bellet Regarding the init that default as prior if possible, I think I have to parse a bit the algorithms to understand which part in the computations is the actual prior and which part is the initial matrix for the iterative algorithms: I'd be more comfortable in making a guess at it in a new PR so that it's easier to comment and talk about it, without polluting this PR, what do you think we merge this PR first and do this in a next PR ? Also the PR has become quite big now so I guess it would improve the readability of the next changes

Sounds good. This is not a major feature anyway (in most cases it is natural to use the prior as the init)

metric_learn/_util.py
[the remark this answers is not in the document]
bellet: No I think it is fine like this

metric_learn/_util.py
perimosocordiae: Without parens you're `raise`-ing the class itself, rather than an instance of the class.

metric_learn/lmnn.py
perimosocordiae: Ah, I didn't see that we're referencing `it` after the loop.

metric_learn/lsml.py
perimosocordiae: Yes, this is good. One small nit: instead of "set 'prior'=='covariance'.", we should do "set `prior='covariance'`."

metric_learn/mlkr.py
perimosocordiae: set `init='pca'`

wdevazelhes: Agreed, will do

wdevazelhes: Agreed, will do

metric_learn/lsml.py
wdevazelhes: Agreed, will do

metric_learn/_util.py
wdevazelhes: That's right, it makes sense

wdevazelhes:
@bellet @perimosocordiae, I addressed all your comments and everything is green, I think it's ready to merge :)

metric_learn/lsml.py
wdevazelhes: done

metric_learn/mlkr.py
wdevazelhes: done

wdevazelhes:
@bellet @perimosocordiae I addressed all your comments so I guess the PR is ready to merge now :)

perimosocordiae:
Merged!

PR #197, opened 2019-04-29
[MRG] Add checks for labels when having pairs

### LINKED ISSUE #164: Ensure labels for positive/negative pairs are as expected

wdevazelhes:
Fixes #164


bellet:
LGTM

wdevazelhes:
I think we are good to merge

PR #198, opened 2019-05-02
[MRG] Fix RCA_Supervised sklearn compat test

wdevazelhes:
I noticed that we didn't test `RCA_Supervised` scikit-learn compatibility

 A first problem in scikit-learn's `check_estimator` that prevented the test to pass was that `RCA_Supervised` couldn't form the default `n_chunks` number of chunks for one of the toy examples that scikit-learn had run. Fixing `num_chunks=2` ensures a normal use of the algorithm is tested (if we had put 1 chunk then it would have become a corner case)and is robuts when we cannot do a lot of chunks of data.

A second problem was just a bug when the input was int, because we couldn't substract inplace the mean to `chunk_data` (which was an array of ints, and numpy returns an error in this case).  Casting to float this object fixes it

### DIFF
--- a/metric_learn/rca.py
+++ b/metric_learn/rca.py
@@ -26,7 +26,11 @@ from .constraints import Constraints
 def _chunk_mean_centering(data, chunks):
   num_chunks = chunks.max() + 1
   chunk_mask = chunks != -1
-  chunk_data = data[chunk_mask]
+  # Warning: we need to ensure we don't overwrite the data
+  # through slices hence we do a copy. We will also need to
+  # ensure the data is float so that we can substract the
+  # mean on it
+  chunk_data = data[chunk_mask].astype(float, copy=True)
   chunk_labels = chunks[chunk_mask]
   for c in xrange(num_chunks):
     mask = chunk_labels == c
@@ -98,7 +102,7 @@ class RCA(MahalanobisMixin, TransformerMixin):
         When ``chunks[i] == -1``, point i doesn't belong to any chunklet.
         When ``chunks[i] == j``, point i belongs to chunklet j.
     """
-    X = self._prepare_inputs(X, ensure_min_samples=2)
+    X, chunks = self._prepare_inputs(X, chunks, ensure_min_samples=2)
 
     # PCA projection to remove noise and redundant information.
     if self.pca_comps is not None:
@@ -109,7 +113,6 @@ class RCA(MahalanobisMixin, TransformerMixin):
       X_t = X - X.mean(axis=0)
       M_pca = None
 
-    chunks = np.asanyarray(chunks, dtype=int)
     chunk_mask, chunked_data = _chunk_mean_centering(X_t, chunks)
 
     inner_cov = np.atleast_2d(np.cov(chunked_data, rowvar=0, bias=1))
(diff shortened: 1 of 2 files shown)


metric_learn/rca.py
```diff
@@ -26,7 +26,11 @@
 def _chunk_mean_centering(data, chunks):
   num_chunks = chunks.max() + 1
   chunk_mask = chunks != -1
-  chunk_data = data[chunk_mask]
+  # Warning: we need to ensure we don't overwrite the data
+  # through slices hence we do a copy. We will also need to
+  # ensure the data is float so that we can substract the
+  # mean on it
+  chunk_data = data[chunk_mask].astype(float, copy=True)
```
perimosocordiae: Logical indexing always creates a copy, so the `copy=True` isn't necessary here. In fact, `astype()` defaults to returning a new copy each time, so what we really want here is `copy=False` to avoid unnecessary copies.

test/test_sklearn_compat.py
```diff
@@ -89,9 +89,13 @@ def stable_init(self, sparsity_param=0.01, num_labeled='deprecated',
     dSDML.__init__ = stable_init
     check_estimator(dSDML)
 
-  # This fails because the default num_chunks isn't data-dependent.
-  # def test_rca(self):
-  #   check_estimator(RCA_Supervised)
+  def test_rca(self):
+    def stable_init(self, num_dims=None, pca_comps=None,
+                    chunk_size=2, preprocessor=None):
```
perimosocordiae: I'd use `**kwargs` here

metric_learn/rca.py
wdevazelhes: That's right, thanks

test/test_sklearn_compat.py
wdevazelhes: I agree that would be better in general, but for here I think it might be useful to simulate dRCA to have the same arguments names as a real RCA, so that if scikit-learn have checks that depend on the arguments of RCA, they will be taken into account, what do you think ? 
Though it's not the case here so maybe yes we can put **kwargs for simplicity

bellet:
Can you briefly explain the problem and its solution in the PR description?

wdevazelhes:
> Can you briefly explain the problem and its solution in the PR description?

Yes sorry
done

bellet:
Thanks. Why is this `stable_init` thing needed?

wdevazelhes:
> Thanks. Why is this `stable_init` thing needed?

In fact there was also another bug because the default `n_chunks` was too big for some toy problem in scikit-learn `check_estimator`, so the `stable_init` is a way to fix it (I updated the PR description, I had forgotten there was this bug too)

bellet:
OK. I guess this could be fixed in a more robust way by making sure that `n_chunks*chunk_size` is not larger than the number of points, or having some kind of elegant break of the chunk generation procedure when no new chunk can be created. But let's keep this for another PR

test/test_sklearn_compat.py
perimosocordiae: I'm not sure how scikit-learn's testing works, but that seems plausible. This is fine as-is.

wdevazelhes:
Merging, since travis test are green, and 2 approvals

Issue #166, opened 2019-02-01
Move from python3.4 to newer (and remove support for python 2.7 ?)


wdevazelhes:
As of march 2019, python 3.4 will cease to be maintained. I think we should update the version required (and the one travis checks). 

Since v 0.21, scikit-learn requires python>=3.5. Btw they also don't support python 2.7 anymore. 

To use skggm we should use at least python 3.6.x (see comment https://github.com/metric-learn/metric-learn/pull/162#issuecomment-459733439)

So after all, maybe we could only support python >=3.6.x ? What do you think ?

bellet:
@perimosocordiae and @terrytangyuan probably have a much better informed opinion than myself about this. But indeed, if sklearn stops supporting python 2.7, we probably should as well, and move to 3.6. Since we are preparing a big release the timing is probably right?

perimosocordiae:
I believe numpy and scipy just released their last version supporting python 2, so it's definitely time to start thinking about going 3-only.

How about making v0.5 our last python 2 release? This way any legacy users can still use the new API design.

perimosocordiae:
For the minimum Python 3 version, I think bumping to 3.5 is certainly fine. I'm a little hesitant about going to 3.6, but I haven't seen a recent breakdown of 3.x user frequencies recently so I may be out of date.

Regarding skggm, we can get around version mismatch issues by making it an optional dependency and falling back to sklearn's implementation otherwise.

terrytangyuan:
+1 to drop Python 2 support. It would be better to support 3.5 though if we can (try some workarounds like what @perimosocordiae suggested). If we cannot get around with skggm issue, then 3.6 sounds fine.

bellet:
Agree that making 0.5 the last Python 2 compatible release is good
For Python 3: we could move to 3.4 or 3.5 and raise an error if skgmm is not installed as mentioned by @wdevazelhes ?

sklearn's implementation of graph lasso is definitely not working for us. We will be attending the sklearn sprint later this month though so we can try to work on that then and see if we can get it fixed

perimosocordiae:
I just realized that we haven't dropped python2 support yet. Version 0.5 was almost a year ago now, so I think it's about time we did it. :-)

PR #199, opened 2019-05-10
[MRG] Be compatible with newer scikit-learn

wdevazelhes:
Scikit-learn just released their last version 0.21, which does not support python<3.5, so I think we can update the travis script to download the 0.20.3 for python3.4 and python2.7, but still use the newest release for python3.6 (so that we test the behaviour on the new scikit-learn), what do you think ? 

Also, there were a few changes that need to be made to work with scikit-learn's new version, which this PR addresses

### DIFF
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -22,8 +22,7 @@ def check_input(input_data, y=None, preprocessor=None,
                 dtype='numeric', order=None,
                 copy=False, force_all_finite=True,
                 multi_output=False, ensure_min_samples=1,
-                ensure_min_features=1, y_numeric=False,
-                warn_on_dtype=False, estimator=None):
+                ensure_min_features=1, y_numeric=False, estimator=None):
   """Checks that the input format is valid, and converts it if specified
   (this is the equivalent of scikit-learn's `check_array` or `check_X_y`).
   All arguments following tuple_size are scikit-learn's `check_X_y`
@@ -88,10 +87,6 @@ def check_input(input_data, y=None, preprocessor=None,
     is originally 1D and ``ensure_2d`` is True. Setting to 0 disables
     this check.
 
-  warn_on_dtype : boolean (default=False)
-    Raise DataConversionWarning if the dtype of the input data structure
-    does not match the requested dtype, causing a memory copy.
-
   estimator : str or estimator instance (default=`None`)
     If passed, include the name of the estimator in warning messages.
 
@@ -111,7 +106,7 @@ def check_input(input_data, y=None, preprocessor=None,
                             copy=copy, force_all_finite=force_all_finite,
                             ensure_min_samples=ensure_min_samples,
                             ensure_min_features=ensure_min_features,
-                            warn_on_dtype=warn_on_dtype, estimator=estimator)
+                            estimator=estimator)
 
   # We need to convert input_data into a numpy.ndarray if possible, before
   # any further checks or conversions, and deal with y if needed. Therefore
@@ -309,9 +304,8 @@ class ArrayIndexer:
                     accept_sparse=True, dtype=None,
                     force_all_finite=False,
                     ensure_2d=False, allow_nd=True,
-                    ensure_min_samples=0,
-                    ensure_min_features=0,
-                    warn_on_dtype=False, estimator=None)
+                    ensure_min_samples=0, ensure_min_features=0,
+                    estimator=None)
     self.X = X
 
   def __call__(self, indices):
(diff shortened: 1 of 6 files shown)


metric_learn/_util.py
```diff
@@ -88,10 +87,6 @@ def check_input(input_data, y=None, preprocessor=None,
     is originally 1D and ``ensure_2d`` is True. Setting to 0 disables
     this check.
 
-  warn_on_dtype : boolean (default=False)
```
wdevazelhes: scikit-learn 0.21 deprecates warn_on_dtype (see: https://github.com/scikit-learn/scikit-learn/issues/13324). Since we never use it (except to test it works), I suggest we get rid of it

test/test_base_metric.py
```diff
@@ -5,86 +5,108 @@
 from sklearn import clone
 from sklearn.utils.testing import set_random_state
 from test.test_utils import ids_metric_learners, metric_learners
+import string
```
perimosocordiae: Is this import used?

test/test_base_metric.py
```diff
@@ -5,86 +5,108 @@
...
+
+
+def remove_spaces(s):
+  s = s.replace(' ', '')
+  s = s.replace('\n', '')
+  return s
```
perimosocordiae: I'd rather use a regex here:

`return re.sub('\s+', ' ', s)`

test/test_base_metric.py
```diff
@@ -5,86 +5,108 @@
...
+    self.assertEqual(remove_spaces(str(metric_learn.Covariance())),
+                     remove_spaces("Covariance(preprocessor=None)"))
 
   def test_lmnn(self):
     self.assertRegexpMatches(
         str(metric_learn.LMNN()),
         r"(python_)?LMNN\(convergence_tol=0.001, k=3, learn_rate=1e-07, "
-        r"max_iter=1000,\n      min_iter=50, preprocessor=None, "
-        r"regularization=0.5, use_pca=True,\n      verbose=False\)")
+        r"max_iter=1000,\n(\ +)min_iter=50, preprocessor=None, "
```
perimosocordiae: `\s+` is a bit more readable, IMO

test/test_base_metric.py
```diff
@@ -5,86 +5,108 @@
...
-""".strip('\n'))
-    self.assertEqual(str(metric_learn.ITML_Supervised()), """
+""".strip('\n')))
+    self.assertEqual(remove_spaces(str(metric_learn.ITML_Supervised())),
+                     remove_spaces("""
 ITML_Supervised(A0=None, bounds='deprecated', convergence_threshold=0.001,
         gamma=1.0, max_iter=1000, num_constraints=None,
         num_labeled='deprecated', preprocessor=None, verbose=False)
-""".strip('\n'))
+""".strip('\n')))
```
perimosocordiae: No need to strip newlines if you're also calling `remove_spaces()`, right?

terrytangyuan:
Yes, @perimosocordiae here's the issue: https://github.com/metric-learn/metric-learn/issues/166

test/test_base_metric.py
wdevazelhes: No, indeed

test/test_base_metric.py
wdevazelhes: That's right, it's better
You meant `return re.sub('\s+', '', s)` to remove the spaces right ?

test/test_base_metric.py
wdevazelhes: I agree, it's better, I didn't know `\s`

test/test_base_metric.py
wdevazelhes: That's right indeed

wdevazelhes:
Thanks for the review @perimosocordiae, I just addressed your comments

test/test_base_metric.py
perimosocordiae: Yeah, should be an empty replacement string.

test/test_base_metric.py
```diff
@@ -7,84 +8,103 @@
...
   def test_lmnn(self):
     self.assertRegexpMatches(
-        str(metric_learn.LMNN()),
-        r"(python_)?LMNN\(convergence_tol=0.001, k=3, learn_rate=1e-07, "
-        r"max_iter=1000,\n      min_iter=50, preprocessor=None, "
-        r"regularization=0.5, use_pca=True,\n      verbose=False\)")
+      str(metric_learn.LMNN()),
+      r"(python_)?LMNN\(convergence_tol=0.001, k=3, learn_rate=1e-07, "
+      r"max_iter=1000,(\s+)min_iter=50, preprocessor=None, "
+      r"regularization=0.5, use_pca=True,(\s+)verbose=False\)")
```
perimosocordiae: Don't need to wrap the whitespace matchers in parens.

wdevazelhes: That's right, done

perimosocordiae:
+1 to merge when the travis results come back.

wdevazelhes:
The results are green merging

PR #201, opened 2019-05-13
[MRG] FIX LMNN gradient and cost function

### LINKED ISSUE #46: LMNN - objective calculate?
The current implementation of [LMNN ](https://github.com/all-umass/metric-learn/blob/master/metric_learn/lmnn.py)python version.
I read the paper of LMNN of NIPS05 version, jmlr08 version, I find that the objective of the optimization process in LMNN consisting of the hinge loss of impostors, namely
`$$[1+||L(x_i - x_j)||^2-||L(x_i - x_l)||^2]_+$$`

But the current implementation is
```python
objective = total_active * (1-reg)
objective += G.flatten().dot(L.T.dot(L).flatten())
```
I guess the 2-nd row is calculating the trace of  M * C, but I cannot figure where is the hinge loss?

wdevazelhes:
Fixes #46 
There seems to be problems in LMNN, that we might want to fix for the next release: 

(Note: to find out what to do we can also compare to https://github.com/scikit-learn/scikit-learn/pull/8602 )

-  Looking into this line the gradient seems to be of the shape of an outer product of Xs (when it should be of the shape of L) (see https://github.com/metric-learn/metric-learn/pull/193#issuecomment-489228612). (There might be a dot product with L missing:
(cf. https://github.com/scikit-learn/scikit-learn/pull/8602/files#diff2914dd42441ed8b594e69b2fe04d23f5R762)) This blocks PR #193:

https://github.com/metric-learn/metric-learn/blob/d4badc8adc0a8ce6689dd9a95e89181efaf5ee24/metric_learn/lmnn.py#L196

- Some part of the loss function seems to be an integer count, when it should be the hinge loss ? as suggested by #46: 

https://github.com/metric-learn/metric-learn/blob/d4badc8adc0a8ce6689dd9a95e89181efaf5ee24/metric_learn/lmnn.py#L198

I created a test that should pass when this is fixed (like the one in scikit-learn's PR)

I'll look into it, but I'm not yet that familiar with LMNN so you might know how to fix it

### DIFF
--- a/metric_learn/lmnn.py
+++ b/metric_learn/lmnn.py
@@ -105,7 +105,7 @@ class python_LMNN(_base_LMNN):
       # objective than the previous L, following the gradient:
       while True:
         # the next point next_L to try out is found by a gradient step
-        L_next = L - 2 * learn_rate * G
+        L_next = L - learn_rate * G
         # we compute the objective at next point
         # we copy variables that can be modified by _loss_grad, because if we
         # retry we don t want to modify them several times
@@ -191,10 +191,12 @@ class python_LMNN(_base_LMNN):
     # do the gradient update
     assert not np.isnan(df).any()
     G = dfG * reg + df * (1 - reg)
+
+    grad = 2 * L.dot(G)
     # compute the objective function
     objective = total_active * (1 - reg)
     objective += G.flatten().dot(L.T.dot(L).flatten())
-    return G, objective, total_active, df, a1, a2
+    return grad, objective, total_active, df, a1, a2
 
   def _select_targets(self, X, label_inds):
     target_neighbors = np.empty((X.shape[0], self.k), dtype=int)
--- a/metric_learn/lmnn.py
+++ b/metric_learn/lmnn.py
@@ -191,12 +191,11 @@ class python_LMNN(_base_LMNN):
     # do the gradient update
     assert not np.isnan(df).any()
     G = dfG * reg + df * (1 - reg)
-
-    grad = 2 * L.dot(G)
+    G = L.dot(G)
     # compute the objective function
     objective = total_active * (1 - reg)
-    objective += G.flatten().dot(L.T.dot(L).flatten())
-    return grad, objective, total_active, df, a1, a2
+    objective += G.flatten().dot(L.flatten())
+    return 2 * G, objective, total_active, df, a1, a2
 
   def _select_targets(self, X, label_inds):
     target_neighbors = np.empty((X.shape[0], self.k), dtype=int)
--- a/test/metric_learn_test.py
+++ b/test/metric_learn_test.py
@@ -2,7 +2,7 @@ import unittest
 import re
 import pytest
 import numpy as np
-from scipy.optimize import check_grad
+from scipy.optimize import check_grad, approx_fprime
 from six.moves import xrange
 from sklearn.metrics import pairwise_distances
 from sklearn.datasets import load_iris, make_classification, make_regression
@@ -21,7 +21,7 @@ from metric_learn import (LMNN, NCA, LFDA, Covariance, MLKR, MMC,
                           RCA_Supervised, MMC_Supervised, SDML)
 # Import this specially for testing.
 from metric_learn.constraints import wrap_pairs
-from metric_learn.lmnn import python_LMNN
+from metric_learn.lmnn import python_LMNN, _sum_outer_products
 
 
 def class_separation(X, labels):
@@ -119,6 +119,61 @@ class TestLMNN(MetricTestCase):
       csep = class_separation(lmnn.transform(self.iris_points),
                               self.iris_labels)
       self.assertLess(csep, 0.25)
+
+  def test_loss_grad_lbfgs(self):
+    """Test gradient of loss function
+    Assert that the gradient is almost equal to its finite differences
+    approximation.
+    """
+    rng = np.random.RandomState(42)
+    X, y = make_classification(random_state=rng)
+    L = rng.randn(rng.randint(1, X.shape[1] + 1), X.shape[1])
+    lmnn = LMNN()
+
+    k = lmnn.k
+    reg = lmnn.regularization
+
+    X, y = lmnn._prepare_inputs(X, y, dtype=float,
+                                ensure_min_samples=2)
+    num_pts, num_dims = X.shape
+    unique_labels, label_inds = np.unique(y, return_inverse=True)
+    lmnn.labels_ = np.arange(len(unique_labels))
+    lmnn.transformer_ = np.eye(num_dims)
+
+    target_neighbors = lmnn._select_targets(X, label_inds)
+    impostors = lmnn._find_impostors(target_neighbors[:, -1], X, label_inds)
+
+    # sum outer products
+    dfG = _sum_outer_products(X, target_neighbors.flatten(),
+                              np.repeat(np.arange(X.shape[0]), k))
+    df = np.zeros_like(dfG)
+
+    # storage
+    a1 = [None]*k
+    a2 = [None]*k
+    for nn_idx in xrange(k):
+      a1[nn_idx] = np.array([])
+      a2[nn_idx] = np.array([])
+
+    # initialize L
+
+    def fun(L):
+        return lmnn._loss_grad(X, L.reshape(-1, X.shape[1]), dfG, impostors, 1,
+                               k, reg,
+                               target_neighbors, df, a1, a2)[1]
+
+    def grad(L):
+        return lmnn._loss_grad(X, L.reshape(-1, X.shape[1]), dfG, impostors,
+                               1, k, reg,
+          target_neighbors, df, a1, a2)[0].ravel()
+
+    # compute relative error
+    epsilon = np.sqrt(np.finfo(float).eps)
+    rel_diff = (check_grad(fun, grad, L.ravel()) /
+                           np.linalg.norm(approx_fprime(L.ravel(), fun,
+                                                       epsilon)))
+               # np.linalg.norm(grad(L))
+    np.testing.assert_almost_equal(rel_diff, 0., decimal=5)
 
 
 def test_convergence_simple_example(capsys):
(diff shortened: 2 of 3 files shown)


bellet:
You must be careful when comparing with https://github.com/scikit-learn/scikit-learn/pull/8602 because this version uses a *squared* hinge loss, which obviously changes the cost function and its gradient

wdevazelhes:
I realized in fact there was just one matrix product with L that was missing, I added it and added a toy example too for the loss function

metric_learn/lmnn.py
```diff
@@ -191,10 +191,12 @@ def _loss_grad(self, X, L, dfG, impostors, it, k, reg, target_neighbors, df,
     # do the gradient update
     assert not np.isnan(df).any()
     G = dfG * reg + df * (1 - reg)
+
+    grad = 2 * L.dot(G)
```
wdevazelhes: I put the multiplication by 2 here so that we can test the real gradient using `check_grad`

metric_learn/lmnn.py
```diff
@@ -191,10 +191,12 @@ def _loss_grad(self, X, L, dfG, impostors, it, k, reg, target_neighbors, df,
...
     # compute the objective function
     objective = total_active * (1 - reg)
```
wdevazelhes: In fact this was right, it's the term coming from the +1 that is in the hinge loss (cf. question from #46)

wdevazelhes:
I guess it is ready to merge now

test/metric_learn_test.py
```diff
@@ -120,6 +120,112 @@ def test_iris(self):
...
+        return lmnn._loss_grad(X, L.reshape(-1, X.shape[1]), dfG, impostors,
+                               1, k, reg, target_neighbors, df.copy(),
+                               list(a1), list(a2))[1]
+
+    def grad(L):
+        # we copy variables that can be modified by _loss_grad, because we
+        # want to have the same result when applying the function twice
+        return lmnn._loss_grad(X, L.reshape(-1, X.shape[1]), dfG, impostors,
+                               1, k, reg, target_neighbors, df.copy(),
+                               list(a1), list(a2))[0].ravel()
```
perimosocordiae: This test seems particularly brittle, as it replicates so much LMNN functionality. Ideally we could structure the LMNN code to make this simpler, but at the least we should avoid copy-paste here:

```python
def loss_grad(flat_L):
  return lmnn._loss_grad(X, flat_L.reshape(...), ...)

fun = lambda x: loss_grad(x)[1]
grad = lambda x: loss_grad(x)[0].ravel()
```

test/metric_learn_test.py
```diff
@@ -120,6 +120,112 @@ def test_iris(self):
...
+                np.linalg.norm(approx_fprime(L.ravel(), fun, epsilon)))
+    np.testing.assert_almost_equal(rel_diff, 0., decimal=5)
+
+
+@pytest.mark.parametrize('X, y, loss', [(np.array([[0], [1], [2], [3]]),
+                                         [1, 1, 0, 0], 3.0),
+                                        (np.array([[0], [1], [2], [3]]),
+                                         [1, 0, 0, 1], 26.)])
+def test_toy_ex_lmnn(X, y, loss):
+  """Test that the loss give the right result on a toy example"""
```
perimosocordiae: Where did you get these values from?

wdevazelhes: I computed them by hand, according to the paper (note that it's only the loss function and that I set L to identity so it's easy since the distances are easy to get from this 1D example).
Maybe I can add a small detail of the computation as a comment ?

wdevazelhes:
> Empirically, how does this PR affect the behavior of LMNN on iris / make_blobs / sandwich / etc ?

This is the difference between master and this PR for LMNN: 

master: 
image

this PR: 
image

So I guess a significant improvement, I'll check for the other datasets

wdevazelhes:
But I agree I haven't been through every computation of the algorithm, I just looked more or less if the loss is correct at initialization (it seems to be on the toy examples), but I didn't check the part on recomputing which constraints are active after each iteration etc
But if we assume the loss is correct, it is, from the code:
`loss = constant + tr(A L.T L)`, with `A = dfG * reg + df * (1 - reg) = constant`, 
so the gradient is necessarily `grad = L (A + A.T) = 2 * L A` ( deriving from equation 3 page of this course: http://cs229.stanford.edu/notes/cs229-notes1.pdf)
So I guess now there is just need to check whether the matrix A is the right one, (i.e. it contains the right outer product differences) ?

wdevazelhes:
For iris, we don't see a big difference:

```python
from sklearn.datasets import load_iris
from metric_learn import LMNN
from sklearn.manifold import TSNE
import matplotlib.pyplot as plt
from sklearn.pipeline import make_pipeline

X, y = load_iris(return_X_y=True)

red_dim = make_pipeline(LMNN(), TSNE())
X_e = red_dim.fit_transform(X, y)

plt.scatter(X_e[:, 0], X_e[:, 1], c=y)
plt.show()
```

With master: 
image

With this PR: 
image

wdevazelhes:
From this example (inspired from #180, i.e. a dataset with two informative features and a big noise for the other features), the difference is pretty clear: 
```python
from sklearn.utils import shuffle
from metric_learn import LMNN
from sklearn.manifold import TSNE
import matplotlib.pyplot as plt
from sklearn.pipeline import make_pipeline

X, y = make_classification(random_state=42, n_redundant=0, shuffle=False)
X[:, 2:] *= 10
X, y = shuffle(X, y)

red_dim = make_pipeline(LMNN(), TSNE())
X_e = red_dim.fit_transform(X, y)

plt.scatter(X_e[:, 0], X_e[:, 1], c=y)
plt.show()
```
With this PR: 
image

With master: 
image

test/metric_learn_test.py
wdevazelhes: Yes you're right, let's go for this clearer expression
done
(note: I didn't use the lambda expression since I was returned a flake8 error, but maybe it's something we want to ignore and still go for the lambda ?)

wdevazelhes:
Sorry my last commit was for the wrong branch, I'll revert it

bellet:
> > Empirically, how does this PR affect the behavior of LMNN on iris / make_blobs / sandwich / etc ?
> 
> This is the difference between master and this PR for LMNN:
> 
> master:
> image
> 
> this PR:
> image
> 
> So I guess a significant improvement, I'll check for the other datasets

This makes me think that we should fix a random seed in the sandwich example so that it is stable when run several times ?

bellet:
What about the problem in the objective function?

metric_learn/lmnn.py
```diff
@@ -191,10 +191,11 @@ def _loss_grad(self, X, L, dfG, impostors, it, k, reg, target_neighbors, df,
     # do the gradient update
     assert not np.isnan(df).any()
     G = dfG * reg + df * (1 - reg)
+    G = L.dot(G)
     # compute the objective function
     objective = total_active * (1 - reg)
-    objective += G.flatten().dot(L.T.dot(L).flatten())
-    return G, objective, total_active, df, a1, a2
+    objective += G.flatten().dot(L.flatten())
+    return 2 * G, objective, total_active, df, a1, a2
```
bellet: I guess the factor 2 comes from differentiating `L.T.dot(L)`. It would be clearer to put it directly in `G` then

wdevazelhes:
> What about the problem in the objective function?

You mean the one with the count being added ? I think it's maybe normal in the end (see comment https://github.com/metric-learn/metric-learn/pull/201#discussion_r286015109). In fact this count should be the sum of all the "+ 1" (the margin) in the hinge losses that are not 0 in the total sum. (But I didn't check yet whether `total_active` is indeed this number of active hinge losses)

bellet:
> > What about the problem in the objective function?
> 
> You mean the one with the count being added ? I think it's maybe normal in the end (see comment [#201 (comment)](https://github.com/metric-learn/metric-learn/pull/201#discussion_r286015109)). In fact this count should be the sum of all the "+ 1" (the margin) in the hinge losses that are not 0 in the total sum. (But I didn't check yet whether `total_active` is indeed this number of active hinge losses)

The hinge loss is equal to max(0, 1 - a). (in the case of LMNN, a is the distance with the impostor minus the distance with the target neighbor)
It is equal to 0 when the margin is satisfied (a >= 1). if not it is equal to the margin violation (1 - a), not "+1"...

wdevazelhes:
> > 
> 
> The hinge loss is equal to max(0, 1 - a). (in the case of LMNN, a is the distance with the impostor minus the distance with the target neighbor)
> It is equal to 0 when the margin is satisfied (a >= 1). if not it is equal to the margin violation (1 - a), not "+1"...

I was not thinking that "+1'" only was the value of the hinge loss, but I was thinking it was the "+1" part of the hinge loss (sum(1-a) = sum(1) - sum(a) = **total_active** - sum(a)). I had the impression that the "df" part later (some outer products), with operations with L, accounted for the sum of "a"s (distance with the impostor minus the distance with the target neighbor) so in total there was something like: `push_loss = total_active - sum_of_the_differences_btw_distance_with_the_impostor_and_distance _with_target_neighbor`
But I definitely need to check if this is indeed the case I didn't deepen the "df" computation part

> You can run some executions showing how the objective function evolves to make sure it goes down at each iteration when small enough step size is used, and compare with the version in master (the number of active constraints should tend to go down but not each time)

Yes good idea I'll do that

wdevazelhes:
Oups, in fact I introduced this bug in PR #101: I forgot to multiply by L when applying the update...

bellet:
Ok fair enough - the objective may be correct but the code is very hard to parse. Let's fix the gradient issue for now and make an issue to maybe refactor the code later to make it more easy to maintain.

wdevazelhes:
So in the end this PR just fixes the bug that I introduced, as well as introduced some tests, so if everybody agrees, I think we are good to merge ? @perimosocordiae ?

perimosocordiae:
Merged!

PR #203, opened 2019-05-13
[MRG] FIX: fix lsml inversion

### LINKED ISSUE #202: LSML fails on some classical inputs

wdevazelhes:
Fixes #202

### DIFF
--- a/metric_learn/lsml.py
+++ b/metric_learn/lsml.py
@@ -39,7 +39,7 @@ class _BaseLSML(MahalanobisMixin):
     tol : float, optional
     max_iter : int, optional
     prior : (d x d) matrix, optional
-        guess at a metric [default: inv(covariance(X))]
+        guess at a metric [default: pinvh(covariance(X))]
     verbose : bool, optional
         if True, prints information while learning
     preprocessor : array-like, shape=(n_samples, n_features) or callable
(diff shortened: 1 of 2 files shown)


wdevazelhes:
> I am not sure that this is a good fix, because logdet divergence is finite only if the matrix is (strictly) positive definite. Hence working the whole time with singular matrices and pseudo-inverses looks like deviating quite a bit from the algorithm.
> 
> I think a better fix may be simply to return an error when the prior is not positive definite. When this happens with the covariance, the identity can be used (this could be done in #180)

That's right, I didn't think of that... 
I noticed that there is the same problem for SDML, and I'm thinking about this case: say the user provides a prior that is not PSD. But the `emp_cov` matrix could be PSD no? In this case should we still forbid a non PSD prior ? I guess so because it does not makes sense given what is a prior ? (it should be a mahalanobis metric)

wdevazelhes:
Since in the end we will use init='identity' for the examples, I guess we don't need to merge this PR quicker, and the work will be done in PR #195 , so I'm closing the PR

bellet:
> I am not sure that this is a good fix, because logdet divergence is finite only if the matrix is (strictly) positive definite. Hence working the whole time with singular matrices and pseudo-inverses looks like deviating quite a bit from the algorithm.
> 
> I think a better fix may be simply to return an error when the prior is not positive definite. When this happens with the covariance, the identity can be used (this could be done in #180)

Let me correct my own claim. While the LogDet divergence is indeed finite only for pairs of PD matrices, in LSML and other algorithms using LogDet (ITML, SDML) having a PSD (not strictly PD) prior is probably fine because the logdet term on the prior can be ignored since it is constant. So this PR could be merged if you use pseudo-inverse only for the prior (but not for the gradient!).

Yet, I think it is probably safer to rule out non-PD priors for these algorithms (see my other comment in #195)

PR #206, opened 2019-05-21
[MRG] Use pseudo-inverse in Covariance

### LINKED ISSUE #204: Update covariance
In the `Covariance` algorithm, we currently do an inverse of the covariance matrix, and the latter could be singular. We need to decide whether to return an error or return the pseudo inverse in this case.

wdevazelhes:
Fixes #204 

This PR uses the pseudo inverse in Covariance, rather than the inverse. See https://github.com/metric-learn/metric-learn/issues/204#issuecomment-492336040

### DIFF
--- a/metric_learn/covariance.py
+++ b/metric_learn/covariance.py
@@ -14,6 +14,7 @@ from sklearn.base import TransformerMixin
 
 from .base_metric import MahalanobisMixin
 from ._util import transformer_from_metric
+import scipy
 
 
 class Covariance(MahalanobisMixin, TransformerMixin):
@@ -35,11 +36,11 @@ class Covariance(MahalanobisMixin, TransformerMixin):
     y : unused
     """
     X = self._prepare_inputs(X, ensure_min_samples=2)
-    M = np.atleast2d(np.cov(X, rowvar = False))
+    M = np.atleast_2d(np.cov(X, rowvar=False))
     if len(M) == 1:
-      M = 1./M
+      M = 1. / M
     else:
-      M = np.linalg.pinvh(M)
+      M = scipy.linalg.pinvh(M)
 
     self.transformer_ = transformer_from_metric(np.atleast_2d(M))
     return self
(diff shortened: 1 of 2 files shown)


metric_learn/covariance.py
```diff
@@ -35,11 +35,11 @@ def fit(self, X, y=None):
     y : unused
     """
     X = self._prepare_inputs(X, ensure_min_samples=2)
-    M = np.cov(X, rowvar = False)
-    if M.ndim == 0:
+    M = np.atleast2d(np.cov(X, rowvar = False))
```
wdevazelhes: I added an atleast2d since we don't want the metric to be scalar

metric_learn/covariance.py
```diff
@@ -35,11 +35,11 @@ def fit(self, X, y=None):
...
+    if len(M) == 1:
```
wdevazelhes: Because of the atleast2d, this needs to change

metric_learn/covariance.py
perimosocordiae: While we're here, remove the spaces in the `rowvar` keyword argument.

metric_learn/covariance.py
perimosocordiae: Can we be more explicit and check `M.size == 1`? Would it be possible for M to be a column vector here?

metric_learn/covariance.py
wdevazelhes: I agree, done

metric_learn/covariance.py
wdevazelhes: Ah yes you're right thanks
I thought that len counted the number of elements but indeed `len(np.array([[1, 2, 3]])) == 1`

wdevazelhes: done

wdevazelhes:
I think it should be good to merge now

metric_learn/covariance.py
```diff
@@ -14,6 +14,7 @@
 
 from .base_metric import MahalanobisMixin
 from ._util import transformer_from_metric
+import scipy
```
perimosocordiae: Style nitpick: move this import up to line 13 (after the numpy import but before the sklearn)

bellet:
Maybe fix the small style issue so we can merge?

metric_learn/covariance.py
wdevazelhes: That's right, done

wdevazelhes:
> Maybe fix the small style issue so we can merge?

Yep, sorry, it's done now I think we can merge

PR #207, opened 2019-05-22
[MRG] Add checks for bounds argument

### LINKED ISSUE #161: improve bounds argument and bounds_ attribute in ITML
See comment: https://github.com/metric-learn/metric-learn/pull/159#discussion_r250234204

We should: 
- Allow tuples, lists, numpy arrays, (any array like, TODO: see if there is other array-like than these 3 types)  as arguments for `bounds` for ITML
- I think a unique type of return for `bounds_` would be better, say a numpy array 
- Do tests to ensure that this works as expected
- Do the right advertising in the dosctring

wdevazelhes:
Fixes #161 

This PR checks that when filling the `bounds` argument in ITML, we use any array-like of two elements
And it adds to the doc that the subsequent `bounds_` attribute will be a `numpy.ndarray`

Question: do we want to ensure that it's actually a 1D array ? For now I only check that is has size 2...

### DIFF
--- a/metric_learn/itml.py
+++ b/metric_learn/itml.py
@@ -69,9 +69,13 @@ class _BaseITML(MahalanobisMixin):
       X = np.vstack({tuple(row) for row in pairs.reshape(-1, pairs.shape[2])})
       self.bounds_ = np.percentile(pairwise_distances(X), (5, 95))
     else:
-      assert len(bounds) == 2
+      bounds = check_array(bounds, allow_nd=False, ensure_min_samples=0,
+                           ensure_2d=False)
+      bounds = bounds.ravel()
+      if bounds.size != 2:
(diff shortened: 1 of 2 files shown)


metric_learn/itml.py
```diff
@@ -192,7 +196,7 @@ def fit(self, pairs, y, bounds=None, calibration_params=None):
     calibration_params = (calibration_params if calibration_params is not
                           None else dict())
     self._validate_calibration_params(**calibration_params)
-    self._fit(pairs, y)
+    self._fit(pairs, y, bounds=bounds)
```
wdevazelhes: This was missing before, but the new tests in this PR made it pop up that it's needed to transmit the bounds argument, otherwise no matter what it will be default (None)

PR #216, opened 2019-06-12
[MRG] Remove shogun dependency

### LINKED ISSUE #132: Remove Shogun dependency
It would be great to improve python_LMNN's performance so as to be able to remove the need for the shogun dependency

### LINKED ISSUE #192: Error with modshogun
I think there is a problem when using shogun: 
First after I installed shogun (the last version, I tried to install with conda forge and also pip), I cannot  `import modshogun` (it throws a package not found error), but instead I can `import shogun`. Maybe I messed up somewhere in the install process or maybe the newest version have a different name or sth..

Even if I change the code with `import shogun` instead of `import modshogun`, there is another error when doing: 

```python
from metric_learn import LMNN
from sklearn.datasets import make_blobs
X, y = make_blobs(n_features=2)
X.shape
lmnn = LMNN()
lmnn.fit(X,y)
```

And it returns the following error:

```pytb
NotImplementedError: Wrong number or type of arguments for overloaded function 'new_MulticlassLabels'.
  Possible C/C++ prototypes are:
    shogun::CMulticlassLabels::CMulticlassLabels()
    shogun::CMulticlassLabels::CMulticlassLabels(int32_t)
    shogun::CMulticlassLabels::CMulticlassLabels(shogun::SGVector< float64_t >)
    shogun::CMulticlassLabels::CMulticlassLabels(shogun::CFile *)
    shogun::CMulticlassLabels::CMulticlassLabels(shogun::CBinaryLabels *)
```

wdevazelhes:
Fixes #132 
Fixes #192 

This removes the shogun dependency

### DIFF
--- a/metric_learn/lmnn.py
+++ b/metric_learn/lmnn.py
@@ -25,7 +25,7 @@ from .base_metric import MahalanobisMixin
 
 
 # commonality between LMNN implementations
-class _base_LMNN(MahalanobisMixin, TransformerMixin):
+class _BaseLMNN(MahalanobisMixin, TransformerMixin):
   def __init__(self, init='auto', k=3, min_iter=50, max_iter=1000,
                learn_rate=1e-7, regularization=0.5, convergence_tol=0.001,
                use_pca=True, verbose=False, preprocessor=None,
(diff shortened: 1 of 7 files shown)


metric_learn/lmnn.py
```diff
@@ -25,7 +25,7 @@
 
 
 # commonality between LMNN implementations
-class _base_LMNN(MahalanobisMixin, TransformerMixin):
+class _BaseLMNN(MahalanobisMixin, TransformerMixin):
```
perimosocordiae: Seems like we can also get rid of this base class, right?

wdevazelhes: Ah yes, that's right! I forgot about that, will do

wdevazelhes: done

wdevazelhes:
Test are green so I guess we can merge

PR #220, opened 2019-06-13
[MRG] Fix quadruplets scoring

wdevazelhes:
There was a problem with scoring in LSML: this PR fixes it (and adds a non-regression test as well as a small test for prediction for pairs learners)

### DIFF
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -641,4 +641,9 @@ class _QuadrupletsClassifierMixin(BaseMetricLearner):
     score : float
       The quadruplets score.
     """
-    return np.mean((self.predict(quadruplets) + 1) / 2)
+    # Since the prediction is a vector of values in {-1, +1}, we need to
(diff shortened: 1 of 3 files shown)


metric_learn/base_metric.py
```diff
@@ -641,4 +641,4 @@ def score(self, quadruplets):
     score : float
       The quadruplets score.
     """
-    return - np.mean(self.predict(quadruplets))
+    return np.mean((self.predict(quadruplets) + 1) / 2)
```
perimosocordiae: This should be equivalent to `pred.mean() / 2  + 0.5`, which is more efficient to compute.

I'd also add a comment explaining that the prediction is a vector of either -1 or 1 values, and that we're rescaling to [0, 1].

wdevazelhes: I agree, done

wdevazelhes:
If everybody agrees, I think we are good to merge

perimosocordiae:
Agreed, merged.

PR #208, opened 2019-05-22
[MRG] Enhance documentation

wdevazelhes:
This PR enhances the documentation by fixing issues about the doc and adding other improvements

Fixes #155 
Fixes #149 
Fixes #150
Fixes #135 

# TODO: 

- [x] Add link to docstring in documentation's titles of algos (fixes #155)
- [x] Fix #149
- [x] Fix #150
- [x] Put the description of algorithms that have a supervised version in the non-supervised version (and not at the top of the page). This allows the user not to scroll up everytime after following a link. We could also make separate pages for algos and their supervised version
- [x] Check that links still work (like MMC puts on the docstring of MMC (had a pb with that, also MahalanobisMixin was not working))
- [x] Check that no `num_dims` remains (they should all be changed into `n_components`, I saw one in MLKR for instance)
- [x] Solve the TODOs inside the .rst files
- [x] Put a list of the methods in the docstring page (automatically), like in scikit-learn. This will make it easier to find methods without having to scroll down every time
- <strike> Ensure that the doctest work </strike> -> postponed, already opened in #156
- [x] some arguments are not documented, ensure that they are all documented
- <strike>Maybe add `metric_learn.constraints.positive_negative_pairs` docstring</strike> -> postponed, discussed in #227
- [x] Address https://github.com/metric-learn/metric-learn/pull/208#pullrequestreview-248843755

Issue #210, opened 2019-05-29
refactor lmnn code


wdevazelhes:
#### Description

LMNN's code could be refactored to be more simple to read (see comment https://github.com/metric-learn/metric-learn/pull/201#issuecomment-496948016)

see also https://github.com/metric-learn/metric-learn/pull/201#discussion_r286032898

zdk123:
Borrowing this issue to request that the LMNN class be refactored to take triplets, as un the SCML code. The current LMNN class could really be named LMNN_Supervised, following the conventions of this package. I could take a stab at this unless there is some reason not to...

PR #223, opened 2019-06-21
[WIP] update impostors

wdevazelhes:
See https://github.com/metric-learn/metric-learn/pull/208#issuecomment-503002436

This fixes the TODO at top of LMNN: it recomputes the impostors at each step

This is a WIP and not tested for now (though it works on a small example but it's not efficient at all, more like a first atempt to do that with very few code change)

Note that in the scikit-learn PR, they do update the impostors https://github.com/scikit-learn/scikit-learn/pull/8602, but they only keep a maximum number of impostors

We also need to decide whether we want impostors updating to be an option or to be always there

Note: for now, I update the impostors, but it's not at all efficient, because the distances comparisons are re-done just after, when computing the active constraints. The advantage of computing impostors is to search for active constraints only amongst them. An idea could be to do the following (I've not thought of it all the way through so it might be wrong):
- for the furthest target neighbors: we compute the impostors (by comparing among all/[a max number of impostors] distances between different labels points and target neighbors)
- amongst those only, we compute the impostors wrt the second furthest neighbor (because if a pair is not an impostor wrt to the furthest target neighbor, it's not either for the second further)
etc 

Note: a quick solution that could be used as a first improvement (and the real improvement would be in V0.6.0) would be to have an option to either use the impostors computed at the beginning, or to use all impostors possible (i.e. all pairs of points of different labels) (which can very easily be implemented in the `find_impostors` function

### DIFF
--- a/metric_learn/lmnn.py
+++ b/metric_learn/lmnn.py
@@ -251,43 +251,44 @@ class LMNN(MahalanobisMixin, TransformerMixin):
                                            target_dist.argsort(axis=1), 1)
     target_dist = np.sort(target_dist, axis=1)
     total_active, push_loss = 0, 0
-    weights = scipy.sparse.csr_matrix((n_samples, n_samples))
+    weights = np.zeros((n_samples, n_samples))
     for c in np.unique(y):  # could maybe avoid this loop and vectorize
-      same_label = y == c  # TODO: I can have this pre-computed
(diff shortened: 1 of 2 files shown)


bellet:
Closing for now as we will go for #228 which is closer to current implementation
We can pick this up later when dealing with #210

PR #224, opened 2019-06-25
[MRG] Remove random_seed in fit and use the one in init

### LINKED ISSUE #171: Remove random seed in fit

wdevazelhes:
Fixes #171 

This PR replaces the `random_seed` in `fit`, by the `random_seed` in `init`.

Note that I also updated the functions `positive_negative_pairs`, `_pairs`, and `adjacency_matrix` to be more like scikit-learn (i.e. have a the `random_state` parameter to be `None` by default), but I don't think it's necessary to put a `ChangedBehaviour` Warning or a `DeprecationWarning`, since those are util functions, that were not previously documented, (and probably less used than `*_Supervised` versions of algorithms), and what is more the previous behaviour will still work (giving `np.random` or any `RandomState`), what do you think ?


test/metric_learn_test.py
```diff
@@ -866,21 +866,6 @@ def test_iris(self):
     csep = class_separation(rca.transform(self.iris_points), self.iris_labels)
     self.assertLess(csep, 0.25)
 
-  def test_feature_null_variance(self):
```
wdevazelhes: while I was at it I removed this test because it tested `pca_comps` which is now deprecated. It failed (when it passed before) probably because of the `random_seed`, which may change because of this PR

test/metric_learn_test.py
```diff
@@ -861,7 +861,8 @@ def test_deprecation_num_dims_lfda(num_dims):
 
 class TestRCA(MetricTestCase):
   def test_iris(self):
-    rca = RCA_Supervised(n_components=2, num_chunks=30, chunk_size=2)
+    rca = RCA_Supervised(n_components=2, num_chunks=30, chunk_size=2,
```
wdevazelhes: I set the random seed here to pass the test. Note that the random seed was not set before, so I checked that the test passed with a random_state 42 before, it did, and now it passes too

wdevazelhes:
I added the `ChangedBehaviorWarning` that we talked about @bellet @nvauquie , if all tests are green, we should be ready to merge

bellet:
LGTM. There is just some conflicts with master to address before we can merge

wdevazelhes:
There was a small pb with the merge but it should be fixed right now, if it's green I think we can merge

metric_learn/constraints.py
```diff
@@ -46,7 +49,8 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
     return a, b, c, d
 
   def _pairs(self, num_constraints, same_label=True, max_iter=10,
-             random_state=np.random):
+             random_state=None):
+    random_state = check_random_state(random_state)
```
perimosocordiae: It's probably safe to remove this check, as all callers of `_pairs` should have done it already.

wdevazelhes: I agree, done

PR #230, opened 2019-07-03
[MRG] Rename transformer_ to components_

### LINKED ISSUE #196: transformer_ or components_ ?
Currently we call the transformer matrix `transformer_` for all algorithms. However in scikit-learn they mostly use `components_` for the (quite) same object. I think we might have discussed this with @bellet and @nvauquie but I'm not sure at all, and don't remember what we decided
If it's better, it would be good to change this before the release to avoid a deprecation cycle

bellet:
Fixes #196 
Also renamed some related methods for consistency


bellet:
I don't think I forgot anything. If CI passes we can merge

PR #231, opened 2019-07-03
[MRG] Deprecate use_pca parameter of LMNN

### LINKED ISSUE #222: Deprecate use_pca for lmnn
There is still a `use_pca` attribute for LMNN, that needs to be deprecated

bellet:
Fixes #222


bellet:
If CI passes, we should be able to merge

PR #238, opened 2019-07-26
[MRG] Address comments for sklearn-contrib integration

wdevazelhes:
Hi, we've made a request for inclusion in  `scikit-learn-contrib`, this PR intends to address the comments of the issue: 
https://github.com/scikit-learn-contrib/scikit-learn-contrib/issues/40 

TODO: 
- [x] Fix flake8 errors (there remains some error due to unused imports in `metric_learn/__init__.py`, but I guess these are needed right ?) And also `inverse_covariance.quic` is imported but unused, but this is normal since it's just to define the variable `HAS_SKGGM`. I don't know if there's another way to bypass this. Here is the log of flake8 after the fixes:

<details>

```
./test/metric_learn_test.py:17:3: F401 'inverse_covariance.quic' imported but unused
./metric_learn/__init__.py:3:1: F401 '.constraints.Constraints' imported but unused
./metric_learn/__init__.py:4:1: F401 '.covariance.Covariance' imported but unused
./metric_learn/__init__.py:5:1: F401 '.itml.ITML' imported but unused
./metric_learn/__init__.py:5:1: F401 '.itml.ITML_Supervised' imported but unused
./metric_learn/__init__.py:6:1: F401 '.lmnn.LMNN' imported but unused
./metric_learn/__init__.py:7:1: F401 '.lsml.LSML' imported but unused
./metric_learn/__init__.py:7:1: F401 '.lsml.LSML_Supervised' imported but unused
./metric_learn/__init__.py:8:1: F401 '.sdml.SDML_Supervised' imported but unused
./metric_learn/__init__.py:8:1: F401 '.sdml.SDML' imported but unused
./metric_learn/__init__.py:9:1: F401 '.nca.NCA' imported but unused
./metric_learn/__init__.py:10:1: F401 '.lfda.LFDA' imported but unused
./metric_learn/__init__.py:11:1: F401 '.rca.RCA' imported but unused
./metric_learn/__init__.py:11:1: F401 '.rca.RCA_Supervised' imported but unused
./metric_learn/__init__.py:12:1: F401 '.mlkr.MLKR' imported but unused
./metric_learn/__init__.py:13:1: F401 '.mmc.MMC_Supervised' imported but unused
./metric_learn/__init__.py:13:1: F401 '.mmc.MMC' imported but unused
./metric_learn/__init__.py:15:1: F401 '._version.__version__' imported but unused
```

</details>

Note that I ignored some errors (E111 (indentation is not a multiple of four),E114 (indentation is not a multiple of four (comment)))

- [x] Put Python 3.7 in the CI tests


test/test_fit_transform.py
```diff
@@ -1,4 +1,3 @@
-import pytest
```
wdevazelhes: I checked and the tests are run even if I remove this line

.travis.yml
```diff
@@ -5,16 +5,13 @@ python:
...
 before_install:
   - sudo apt-get install liblapack-dev
   - pip install --upgrade pip pytest
-  - pip install wheel cython numpy scipy codecov pytest-cov
-  - if $TRAVIS_PYTHON_VERSION == "3.6"; then
-        pip install scikit-learn;
-    else
-        pip install scikit-learn==0.20.3;
-    fi
+  - pip install wheel cython numpy scipy codecov pytest-cov scikit-learn
```
wdevazelhes: I think now for python versions that are not supported by scikit-learn 0.21 and up, pip will now automatically install the V0.20.3. Let's see if it works

wdevazelhes:
I think the coverage failed because sometimes I broke some too long lines, adding lines of code but not increasing coverage. I think the coverage is still high so we should be able to merge

bellet:
About the imports in init, using __all__ as suggested in the thread below should solve the problem
https://stackoverflow.com/questions/31079047/python-pep8-class-in-init-imported-but-not-used

bellet:
For the skggm import, a hacky workaround could be to add an `assert(quic)` ;-)

wdevazelhes:
Thanks @bellet, I just addressed your comments

wdevazelhes:
I also tried to add an automatic flake8 check that will be run on the whole code after each PR (last commit), now that it's all compliant it's possible to do that and not bother about doing it just on the diff
Let's see if it works

bellet:
I think there is a syntax problem with the if statement in bash for string comparison. You are missing the double square brackets [[

wdevazelhes:
> I think there is a syntax problem with the if statement in bash for string comparison. You are missing the double square brackets [[

That's right, it should work now

wdevazelhes:
I just realized I was using an old flake8 so in fact there are some flake8 errors that are not solved, which explains the new fail
I'll fix them too

wdevazelhes:
I just updated the list of errors, removing W504, because it makes impossible situations when activated at the same time as W503 (see this for instance https://gitlab.com/pycqa/flake8/issues/466, also note that scikit-learn has disabled it (see https://github.com/scikit-learn/scikit-learn/blob/4cf4e6e519189c8969f5820ebbca388cf87bf0cb/setup.cfg#L29))

wdevazelhes:
It appears to work now

.travis.yml
```diff
@@ -5,20 +5,21 @@ python:
...
   - if [[ ($TRAVIS_PYTHON_VERSION == "3.6") ||
-          ($TRAVIS_PYTHON_VERSION == "2.7")]]; then
+          ($TRAVIS_PYTHON_VERSION == "3.7") ||
+          ($TRAVIS_PYTHON_VERSION == "2.7") ]]; then
               pip install git+https://github.com/skggm/skggm.git@a0ed406586c4364ea3297a658f415e13b5cbdaf8;
     fi
 script:
+  # we run flake8 to ensure syntax (no need to run it on every version)
+  - if [[ $TRAVIS_PYTHON_VERSION == "3.7" ]];
+      then flake8 --ignore=E111,E114,W504;
```
bellet: I think we should also ignore the other errors that are ignored by default by flake8, just like sklearn does, to avoid unimportant failures which could pop up in subsequent PRs

.travis.yml
```diff
@@ -5,20 +5,21 @@ python:
...
+  - pip install wheel cython numpy scipy codecov pytest-cov scikit-learn flake8
   - if [[ ($TRAVIS_PYTHON_VERSION == "3.6") ||
-          ($TRAVIS_PYTHON_VERSION == "2.7")]]; then
+          ($TRAVIS_PYTHON_VERSION == "3.7") ||
+          ($TRAVIS_PYTHON_VERSION == "2.7") ]]; then
               pip install git+https://github.com/skggm/skggm.git@a0ed406586c4364ea3297a658f415e13b5cbdaf8;
     fi
 script:
+  # we run flake8 to ensure syntax (no need to run it on every version)
+  - if [[ $TRAVIS_PYTHON_VERSION == "3.7" ]];
```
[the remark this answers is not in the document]
wdevazelhes: That's right, thanks
done

.travis.yml
wdevazelhes: I agree, I used the syntax `--extend-ignore` (https://flake8.pycqa.org/en/3.7.8/user/options.html?highlight=extend-ignore#cmdoption-flake8-extend-ignore), so that flake8 will ignore the default errors (including W504), and also E111 and E114

bellet:
> I also tried to add an automatic flake8 check that will be run on the whole code after each PR (last commit), now that it's all compliant it's possible to do that and not bother about doing it just on the diff
> Let's see if it works

About this: I am not sure what is commonly done, but checking only the diff seems to me like a better solution in the long run: if at some point we have a very good reason to merge some PR which is not compliant, we can do so and continue to check compliance of newly added code. This is what is done in sklearn (script is here: https://github.com/scikit-learn/scikit-learn/blob/master/build_tools/circle/flake8_diff.sh).

Not sure how difficult it is to make travis check only the diff but I think it may be preferable?

wdevazelhes:
> About this: I am not sure what is commonly done, but checking only the diff seems to me like a better solution in the long run: if at some point we have a very good reason to merge some PR which is not compliant, we can do so and continue to check compliance of newly added code. This is what is done in sklearn (script is here: https://github.com/scikit-learn/scikit-learn/blob/master/build_tools/circle/flake8_diff.sh).
> 
> Not sure how difficult it is to make travis check only the diff but I think it may be preferable?

Yes I agree it would be a good idea, and in theory it shouldn't be so difficult (locally that's what I do with the command `git diff HEAD..master | flake8 --diff` for instance, but in practice doing it online with travis it seems to be more difficult (given the big file in scikit-learn)... I think it's because on travis, the tests are run when the PR is merged with master already, so one has to go see in the remote what was the last commit before the PR is merged or sth... I'll give it a try I think the most simple is simply to copy paste scikit-learn file (and remove parts which are useless for us (like the "exclude" part), but some parts will be hard to figure out whether they're useless or not for me so we might just keep them ?) what do you think ?

bellet:
The error does not seem to be picked up by travis for some reason

wdevazelhes:
> The error does not seem to be picked up by travis for some reason

Yes, that's because I wasn't actually calling `test_files()`, but calling it seems to create a bug that I don't understand... I'll investigate that

bellet:
Well, it seems to pick up one of the two errors this time?

bellet:
And a last nitpick: it is a little misleading that the check is done within the "Python 3.7" job. If possible it would be more clear to make it a separate travis job?

wdevazelhes:
> Well, it seems to pick up one of the two errors this time?

Yes, it's just that I would have expected it to exit with exit code 1, instead of that it seems to shut down the program. Locally it does the same, (it shuts down my current terminal). But I tried with the scikit-learn's one though (with a fake pep8 error too), and it seems to do the same... Also, it works (terminal is not shut down), when there are no PEP8 errors, so this means it would do what we want (continue to run if there are no pep8 errors)
Maybe that's the expected effect actually, because it's cool it prevents the travis job from continuing to run when there are PEP8 errors

wdevazelhes:
> And a last nitpick: it is a little misleading that the check is done within the "Python 3.7" job. If possible it would be more clear to make it a separate travis job?

Yes I agree, this was a way to have little modifications with what we have currently, but I'll have a look at how scikit-learn does it because they seem to have a more customizable way to deal with travis runs (instead of having just one for each version they actually know how to run different jobs with different parameters indeed)

wdevazelhes:
> Well, it seems to pick up one of the two errors this time?

about the fact that it prints just one error, it could be that it fails and exit at the first of the two errors, but based on what I saw running locally the simple command, the option "--continued=0" in the git diff  should be responsible for that: see below two runs with different options:

```
git diff --unified=0 e7837a1..49235f9 | flake8 --diff --extend-ignore=E111,E114 --show-source
metric_learn/mmc.py:23:18: E221 multiple spaces before operator
    self.max_iter  = max_iter
                 ^
```

```
 git diff e7837a1..49235f9 | flake8 --diff --extend-ignore=E111,E114 --show-source 
metric_learn/mmc.py:23:18: E221 multiple spaces before operator
    self.max_iter  = max_iter
                 ^
metric_learn/mmc.py:35:3: E303 too many blank lines (2)
  def _fit(self, pairs, y):
  ^
```

bellet:
> Maybe that's the expected effect actually, because it's cool it prevents the travis job from continuing to run when there are PEP8 errors

Well what is not that cool is that we only see the first PEP8 error..

bellet:
OK, if you find a way to show all pep8 errors, and it is a different run/job (so that it does not prevent from seeing some other pytest failures down the line), I guess it is fine that it shuts down

bellet:
@perimosocordiae @terrytangyuan you guys may be more knowledgeable than us on these things, feel free to step in :-)

wdevazelhes:
As discussed with @bellet and @nvauquie, I modified `.travis.yml` so as to run different jobs with different configurations, and it seems to work (see the result of https://github.com/metric-learn/metric-learn/pull/238/commits/383a90333e951e33168f9e6925bcdb34871d6d3b). We've decided to go back to running a full PEP8 check on every file, since some errors could be missed by a diff checking (e.g. deleting a line with some function used, but forgetting to delete the import). I left the script `build_toold/travis/flake_diff.sh` if needed in case we want to test on the diff. 
I removed the PEP8 errors introduced for testing, so we should be good to merge

wdevazelhes:
It seems that there is a problem with the code coverage now... I forgot one of the fields `after_success` though, I fixed it, let's see if it works after that

wdevazelhes:
It seems to work now, so I guess we can merge ?

bellet:
Nice!
I think there is still a small issue that the code coverage will run even if the tests fail
To fix this you could probably use stages (the first one for tests and the second one for flake8)? see example in
https://docs.travis-ci.com/user/build-stages
(stage name could be "Syntax checking" and the job name "PEP8" or "flake8")

bellet:
I guess that's fine, there is indeed a single report generated that takes into account each build (https://codecov.io/github/metric-learn/metric-learn/commit/d244bc59ad55ad8257eb82e21b466930c636ebc4), but the results are not shown here
On the other hand the pep8 check passes

Maybe you can quickly try to introduce an error which is not in all builds to see what happens? (easiest way is probably to add an error in one of the skggm tests)

.travis.yml
```diff
@@ -1,27 +1,53 @@
...
-
+language: python
+matrix:
+  include:
+  - name: "Pytest python 2.7 with skggm"
+    python: "2.7"
+    before_install:
+      - sudo apt-get install liblapack-dev
+      - pip install --upgrade pip pytest
+      - pip install wheel cython numpy scipy codecov pytest-cov scikit-learn flake8
```
bellet: can remove flake8 on pytest jobs

.travis.yml
```diff
@@ -1,27 +1,53 @@
...
+    python: "3.7"
+    before_install:
+      - sudo apt-get install liblapack-dev
+      - pip install --upgrade pip pytest
+      - pip install wheel cython numpy scipy codecov pytest-cov scikit-learn flake8
+      - pip install git+https://github.com/skggm/skggm.git@a0ed406586c4364ea3297a658f415e13b5cbdaf8;
+    script:
+      - pytest test --cov;
+      - bash <(curl -s https://codecov.io/bash)
+  - name: "Syntax checking"
```
bellet: maybe name this "Syntax checking with flake8" to give an explicit mention of the tool used to check the syntax (for new contributors)

.travis.yml
wdevazelhes: That's right, thanks

.travis.yml
wdevazelhes: I agree, will do

terrytangyuan:
I’ve never done flake8 check only on the diff before. What’s the motivation for that? If we are worried about having to fix a lot of flake8 errors, I’ve had good experience with automatic code style fixer like [black](https://github.com/psf/black) before.

bellet:
> I’ve never done flake8 check only on the diff before. What’s the motivation for that? If we are worried about having to fix a lot of flake8 errors, I’ve had good experience with automatic code style fixer like [black](https://github.com/psf/black) before.

We considered it in case we ever need to let some flake8 errors through at some point (this is also what sklearn does). But for reasons explained above we agree that, at least for now, flake8 check on the whole project is better

bellet:
Looks like things are working as expected? Let's remove the error and we should be good to merge?

test_flake.py
```diff
@@ -0,0 +1,3 @@
+import nimp
```
bellet: we don't need this file do we?

terrytangyuan:
Using Travis stages is fine too but keep in mind that any subsequent stages won’t start until it’s precious stage passes. This might reduce Travis build time but will increase development time since developers won’t see all the errors besides lint check at once.

bellet:
> Using Travis stages is fine too but keep in mind that any subsequent stages won’t start until it’s previous stage passes. This might reduce Travis build time but will increase development time since developers won’t see all the errors besides lint check at once.

Yep. In the end we are not using stages and running everything in parallel (see last CI results)

test_flake.py
wdevazelhes: That's right, thanks, I forgot I created it in `metric-learn`, and added it by mistake...

wdevazelhes:
After looking again at the build that failed that made me remove the `after_success` flag, I realized there was a timeout error in the python 3.7 build, which is maybe responsible for the fact that the coverage didn't print, so I'll just try to relaunch it, and I guess if it works it's even better/cleaner no ?

wdevazelhes:
Hurray! :D It worked this time, when there was no timeout error, so I guess it's ok to merge ?

bellet:
You can now update the sklearn-contrib integration issue :-)

wdevazelhes:
> You can now update the sklearn-contrib integration issue :-)

Done :-)

PR #241, opened 2019-08-01
Extract skggm commit hash to an environment variable

terrytangyuan:
This makes it easier to upgrade later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: python
 sudo: false
 cache: pip
 language: python
+env:
+  global:
+    - SKGGM_VERSION=a0ed406586c4364ea3297a658f415e13b5cbdaf8
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


bellet:
I was confused by the two travis checks but I just realized it must be because @terrytangyuan  pushed to a new branch on this repo (not his fork), so I guess this is normal
However I am not sure why there is no no test coverage report shown here although a report has been generated, with 2x4 builds (https://codecov.io/github/scikit-learn-contrib/metric-learn/commit/ddca6f3dccbccca902f77da9f00f1154c5b53926)

terrytangyuan:
Good catch. I’ve fixed the issue with two Travis builds in https://github.com/scikit-learn-contrib/metric-learn/pull/242. Not sure about the coverage report though. It looks like a separate issue that needs to be looked into.

bellet:
Thanks merging

PR #242, opened 2019-08-02
Only build on master branch

terrytangyuan:
This will avoid wasting resources in case we are not working on a fork. PR builds will still work.


bellet:
Nice, merged

PR #243, opened 2019-08-02
Add a function that does nothing to test coverage

wdevazelhes:


bellet:
OK. Looks like codecov is clever and show test coverage when there is no actual code modified? So, I guess we can close

PR #248, opened 2019-08-14
[MRG] Update README with paper reference

bellet:


terrytangyuan:
Should we also include this on the website?

bellet:
yes we could. maybe on index page?

terrytangyuan:
Yeah sounds good maybe above the outline.

PR #254, opened 2019-10-29
Break chunks generation in RCA when not enough possible chunks, fixes issue #200

### LINKED ISSUE #200: Break chunks generation in RCA when needed
See comment https://github.com/metric-learn/metric-learn/pull/198#issuecomment-490927459

RobinVogel:
Fixes #200 
See comment https://github.com/scikit-learn-contrib/metric-learn/pull/198#issuecomment-490927459

I count for each set of instances of each label how many chunks it holds, and sum those.
If it is lower than the number of requested chunks, I raise an error.

### DIFF
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -79,10 +79,11 @@ class Constraints(object):
     chunks = -np.ones_like(self.known_label_idx, dtype=int)
     uniq, lookup = np.unique(self.known_labels, return_inverse=True)
     all_inds = [set(np.where(lookup == c)[0]) for c in xrange(len(uniq))]
-    max_chunks = np.sum([len(s) // chunk_size for s in all_inds])
+    max_chunks = int(np.sum([len(s) // chunk_size for s in all_inds]))
     if max_chunks < num_chunks:
-      raise ValueError(('Not enough examples in each class to form '
-                        '%d chunks of %d examples') % (num_chunks, chunk_size))
+      raise ValueError(('Not enough examples in each class to form %d chunks '
+                        'of %d examples - maximum number of chunks is %d'
+                        ) % (num_chunks, chunk_size, max_chunks))
     idx = 0
     while idx < num_chunks and all_inds:
       if len(all_inds) == 1:
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -80,9 +80,9 @@ class Constraints(object):
     uniq, lookup = np.unique(self.known_labels, return_inverse=True)
     all_inds = [set(np.where(lookup == c)[0]) for c in xrange(len(uniq))]
     max_chunks = np.sum([len(s) // chunk_size for s in all_inds])
-    if  max_chunks < num_chunks:
-      raise ValueError('Not enough examples in each class to form '
-                       +'%d chunks of %d examples' % (num_chunks, chunk_size))
+    if max_chunks < num_chunks:
+      raise ValueError(('Not enough examples in each class to form '
+                        '%d chunks of %d examples') % (num_chunks, chunk_size))
     idx = 0
     while idx < num_chunks and all_inds:
       if len(all_inds) == 1:
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -79,6 +79,10 @@ class Constraints(object):
     chunks = -np.ones_like(self.known_label_idx, dtype=int)
     uniq, lookup = np.unique(self.known_labels, return_inverse=True)
     all_inds = [set(np.where(lookup == c)[0]) for c in xrange(len(uniq))]
+    max_chunks = np.sum([len(s) // chunk_size for s in all_inds])
+    if  max_chunks < num_chunks:
+      raise ValueError('Not enough examples in each class to form '
+                       +'%d chunks of %d examples' % (num_chunks, chunk_size))
     idx = 0
     while idx < num_chunks and all_inds:
       if len(all_inds) == 1:
@@ -93,9 +97,6 @@ class Constraints(object):
       inds.difference_update(ii)
       chunks[ii] = idx
       idx += 1
-    if idx < num_chunks:
-      raise ValueError('Unable to make %d chunks of %d examples each' %
-                       (num_chunks, chunk_size))
     return chunks
 
 
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -81,9 +81,10 @@ class Constraints(object):
     all_inds = [set(np.where(lookup == c)[0]) for c in xrange(len(uniq))]
     max_chunks = int(np.sum([len(s) // chunk_size for s in all_inds]))
     if max_chunks < num_chunks:
-      raise ValueError(('Not enough examples in each class to form %d chunks '
-                        'of %d examples - maximum number of chunks is %d'
-                        ) % (num_chunks, chunk_size, max_chunks))
+      raise ValueError(('Not enough possible chunks of %d elements in each'
+                        ' class to form expected %d chunks - maximum number'
+                        ' of chunks is %d'
+                        ) % (chunk_size, num_chunks, max_chunks))
     idx = 0
     while idx < num_chunks and all_inds:
       if len(all_inds) == 1:
--- a/test/test_constraints.py
+++ b/test/test_constraints.py
@@ -25,38 +25,38 @@ def gen_labels_for_chunks(num_chunks, chunk_size,
   return shuffle(labels, random_state=SEED)
 
 
-class TestConstraints(unittest.TestCase):
-  def test_exact_num_points_for_chunks(self):
-    """Checks that the chunk generation works well with just enough points."""
-    for num_chunks, chunk_size in [(5, 10), (10, 50)]:
-      labels = gen_labels_for_chunks(num_chunks, chunk_size)
+@pytest.mark.parametrize("num_chunks, chunk_size", [(5, 10), (10, 50)])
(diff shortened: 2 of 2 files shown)


bellet:
Thanks for getting started on this. The solution looks fine but you need to write some tests to ensure that everything works as intended. You should test that the error is raised when we expect it, but it would also be useful to make sure everything goes through in cases when there is enough point. Generally it is good to test some edge cases (e.g., just enough points, one point missing, etc).

metric_learn/constraints.py
```diff
@@ -79,6 +79,10 @@ def chunks(self, num_chunks=100, chunk_size=2, random_state=None):
     chunks = -np.ones_like(self.known_label_idx, dtype=int)
     uniq, lookup = np.unique(self.known_labels, return_inverse=True)
     all_inds = [set(np.where(lookup == c)[0]) for c in xrange(len(uniq))]
+    max_chunks = np.sum([len(s) // chunk_size for s in all_inds])
+    if max_chunks < num_chunks:
+      raise ValueError(('Not enough examples in each class to form '
+                        '%d chunks of %d examples') % (num_chunks, chunk_size))
```
bellet: maybe the message can me made more informative for the user by specifying the maximal number of chunks that can actually be constructed

RobinVogel:
I took some time to get familiar with tests and wrote one for the two untested cases in the chunk generation that [you discussed](https://github.com/scikit-learn-contrib/metric-learn/pull/254#issuecomment-547533185). I tried to model what I wrote on the existing tests, and wrote something relatively general, at the risk of being verbose. 

I separated the generation of labels from the tests and wrote it myself, since it needs to satisfy some constraints for edge cases.

I hesitated to group the tests in a class, as in `test_sklearn_compat.py`, since `test_constraints.py` is supposed to test the class `Constraints`.

test/test_constraints.py
```diff
@@ -0,0 +1,57 @@
...
+def test_chunk_case_exact_num_points(num_chunks, chunk_size,
+                                     n_classes=10, n_unknown_labels=5):
+  """Checks that the chunk generation works well with just enough points."""
+  labels = gen_labels_for_chunks(num_chunks, chunk_size,
+                                 n_classes=n_classes,
+                                 n_unknown_labels=n_unknown_labels)
+  constraints = Constraints(labels)
+  chunks = constraints.chunks(num_chunks=num_chunks, chunk_size=chunk_size,
+                              random_state=SEED)
+  return chunks
```
bellet: Instead of simply returning `chunks`, it would be nice to test that `chunks` contains `num_chunks` chunks, each composed of `chunk_size` points

metric_learn/constraints.py
```diff
@@ -79,6 +79,11 @@ def chunks(self, num_chunks=100, chunk_size=2, random_state=None):
     chunks = -np.ones_like(self.known_label_idx, dtype=int)
     uniq, lookup = np.unique(self.known_labels, return_inverse=True)
     all_inds = [set(np.where(lookup == c)[0]) for c in xrange(len(uniq))]
+    max_chunks = int(np.sum([len(s) // chunk_size for s in all_inds]))
+    if max_chunks < num_chunks:
+      raise ValueError(('Not enough examples in each class to form %d chunks '
```
bellet: After reading this message again, maybe it would be more clear to remove "in each class"? This may confuse the user into thinking that the number of chunks requested is per class rather than global

test/test_constraints.py
RobinVogel: I will add it.

A remark, I expected chunks to map the *datapoint index* (input order) to the chunk number, but chunks removes -1 instances (unknown labels, according to line 21), which means that one has to consider *datapoint indexes* when removing -1's. But it must be taken into account in the implementation of RCA.

metric_learn/constraints.py
RobinVogel: I tried to make it clear but it is not either per class or global in my opinion. It is  the sum of the number of possible chunks in each class. The criterion is exploited in lines 15 and 16 of `test/test_constraints.py`.

**Example:** With 1000 classes with 2 elements the error will raise when forming chunks of 5 elements, while with 2 classes of 10 elements the error won't raise if 4 chunks of 5 elements are requested. 

**Possible resolutions:**

1. Remove "in each class",
2. Replace by "Not enough possible chunks in each class to form expected %d chunks"
3. Raise *not enough points* when `num_chunks*chunk_size < num_points`, *points need to be more evenly distributed between classes* or *too many classes* in other failure cases.
4. Leave as is.

RobinVogel: > Nice! Indeed you could put the tests into a class, maybe simply `TestChunks` or something

I thought the best way to encapsulate it was to just loop over the parameters I wanted to test in methods of the class (see commit 4f8c247). Trying to fix the data in advances was a bad idea to test all the `Constraint` class.

test/test_constraints.py
```diff
@@ -0,0 +1,63 @@
...
+  unknown_labels = -1 * np.ones(n_unknown_labels, dtype=int)
+
+  labels = np.concatenate([first_labels, remaining_labels, unknown_labels])
+  return shuffle(labels, random_state=SEED)
+
+
+class TestConstraints(unittest.TestCase):
+  def test_exact_num_points_for_chunks(self):
+    """Checks that the chunk generation works well with just enough points."""
+    for num_chunks, chunk_size in [(5, 10), (10, 50)]:
```
wdevazelhes: Instead of having a loop here, you can maybe use `pytest.parametrize` (see https://github.com/scikit-learn-contrib/metric-learn/search?q=parametrize&unscoped_q=parametrize), I find it to make tests debugging easier (it creates sort of "subtests" that can be debugged independently e.g. with IDEs like PyCharm)
(though I'm not sure it will work here since we are in  a unittest class and sometimes they can interfere weirdly with pytest but you can always try if you want)

test/test_constraints.py
bellet: Can you elaborate on this? Not sure I understand what you mean.

Do we test at all `RCA_Supervised` in the presence of unlabeled points? At first sight, it looks like it might fail due to unlabeled points being removed from the chunk array as pointed out by
@RobinVogel. @wdevazelhes, any comment?

test/test_constraints.py
bellet: Indeed `pytest.parametrize` is not compatible with the unittest class:
https://stackoverflow.com/questions/35561843/py-test-parametrize-test-cases-from-classes

I think using `pytest.parametrize` is indeed more elegant and modular, so I would suggest to remove the class and revert to the `pytest.parametrize` formulation you had in a previous commit. Or to use a regular class as in the first response in the link above so that both are compatible

test/test_constraints.py
```diff
@@ -0,0 +1,63 @@
...
+      labels = gen_labels_for_chunks(num_chunks, chunk_size)
+
+      constraints = Constraints(labels)
+      chunks = constraints.chunks(num_chunks=num_chunks, chunk_size=chunk_size,
+                                  random_state=SEED)
+
+      chunk_no, size_each_chunk = np.unique(chunks, return_counts=True)
+
+      self.assertTrue(np.all(size_each_chunk == chunk_size))
```
bellet: Here it is better to use numpy's builtin assert methods which operate directly on arrays. In this case:
https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.testing.assert_array_equal.html

RobinVogel: Done and understood.

test/test_constraints.py
RobinVogel: While using a class `TestConstraints` mimics the `Constraints` class in `constraints.py`, it is not necessary since the file is called `test_constraints.py`. I reverted to the `pytest.parametrize` formulation for brevity.

test/test_constraints.py
RobinVogel: I'll just add an example to make it clearer. Below is a comparison in the presence of unknown labels between the output that I expected from `Constraints.chunks`, i.e. `expected_chunk` and the one that I obtain from `Constraints.chunks`, i.e. `chunks`:
```python
In [1]: from metric_learn.constraints import Constraints                                                                                                                                                        

In [2]: partial_labels = [1, 2, 2, 1, -1, 3, 3]                                                        

In [3]: cons = Constraints(partial_labels)                                                             

In [4]: cons.chunks(num_chunks=3, chunk_size=2)                                                        
Out[4]: array([0, 1, 1, 0, 2, 2])

In [5]: chunks = cons.chunks(num_chunks=3, chunk_size=2)                                               

In [6]: len(chunks), len(partial_labels)                                                               
Out[6]: (6, 7)

In [7]: expected_chunk = [0, 1, 1, 0, -1, 2, 2]
```

bellet: Thanks. And this is indeed a problem for `RCA_Supervised`, see the issue #260 that I just added. I have also added another issue (#261) to make clear in the doc that `-1` is interpreted as unlabeled. You're very welcome to address those @RobinVogel while you have everything well in mind :-)

test/test_constraints.py
wdevazelhes: > I think using `pytest.parametrize` is indeed more elegant and modular, so I would suggest to remove the class and revert to the `pytest.parametrize` formulation you had in a previous commit. Or to use a regular class as in the first response in the link above so that both are compatible

Ah yes that's right sorry @RobinVogel I hadn't seen you already had used `pytest.parametrize` in your previous commits
Yes I guess having tests with explicit names and explicit filenames should be enough for the structure, and pytest's modularity should make it easy to reorganize in the future if needed

wdevazelhes:
This looks great !

PR #263, opened 2019-11-14
Repairs chunk generation for unknown labels, solves issue #260

### LINKED ISSUE #260: [Bug] RCA_Supervised crashes when fit on dataset with unlabeled points
#### Description

While the module `Constraints` provide the ability to have unlabeled points as input (labeled `-1`), the method `chunks` removes unlabeled points in the returned `chunks` array, which thus has different dimension than the dataset `X` passed as input to the `fit` method of `RCA_Supervised`.

I think the most natural and simple solution is to keep the unlabeled points in `chunks` with value `-1`, which is already  interpreted by `RCA` as "not belonging to any chunk".

#### Steps/Code to Reproduce

```python
from metric_learn import RCA_Supervised
import numpy as np

X = np.random.rand(5, 2)
y = [1, 1, -1, 2, 2]

rca = RCA_Supervised(num_chunks=2)
rca.fit(X, y)
```

#### Expected Results
Fit without error

#### Actual Results
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/aurelien/Documents/research/github/metric-learn/metric_learn/rca.py", line 244, in fit
    return RCA.fit(self, X, chunks)
  File "/home/aurelien/Documents/research/github/metric-learn/metric_learn/rca.py", line 132, in fit
    X, chunks = self._prepare_inputs(X, chunks, ensure_min_samples=2)
  File "/home/aurelien/Documents/research/github/metric-learn/metric_learn/base_metric.py", line 101, in _prepare_inputs
    **kwargs)
  File "/home/aurelien/Documents/research/github/metric-learn/metric_learn/_util.py", line 131, in check_input
    y_numeric=y_numeric)
  File "/home/aurelien/anaconda3/lib/python3.7/site-packages/sklearn/utils/validation.py", line 729, in check_X_y
    check_consistent_length(X, y)
  File "/home/aurelien/anaconda3/lib/python3.7/site-packages/sklearn/utils/validation.py", line 205, in check_consistent_length
    " samples: %r" % [int(l) for l in lengths])
ValueError: Found input variables with inconsistent numbers of samples: [5, 4]
```

RobinVogel:
The discussions in pull request #254 were about the generations of chunks in `constraints.py` not providing a map of the indice of the element to the chunk in the presence of unknown labels.
See the example below:

> I'll just add an example to make it clearer. Below is a comparison in the presence of unknown labels between the output that I expected from `Constraints.chunks`, i.e. `expected_chunk` and the one that I obtain from `Constraints.chunks`, i.e. `chunks`:
> 
> ```python
> In [1]: from metric_learn.constraints import Constraints                                                                                                                                                        
> 
> In [2]: partial_labels = [1, 2, 2, 1, -1, 3, 3]                                                        
> 
> In [3]: cons = Constraints(partial_labels)                                                             
> 
> In [4]: cons.chunks(num_chunks=3, chunk_size=2)                                                        
> Out[4]: array([0, 1, 1, 0, 2, 2])
> 
> In [5]: chunks = cons.chunks(num_chunks=3, chunk_size=2)                                               
> 
> In [6]: len(chunks), len(partial_labels)                                                               
> Out[6]: (6, 7)
> 
> In [7]: expected_chunk = [0, 1, 1, 0, -1, 2, 2]
> ```
> 
> The output is not a map such that `expected_chunk[i]` is the chunk of element `i`, which I expected, but might not matter in RCA.

This pull request solves that, which resolves #260 about RCA crashing when used with unknown labels. Since unknown labels seem to be handled in `rca.py` with a variable named `chunk_mask` I expect `rca.py` to work with unknown labels now.

I had to save the partial labels in `constraints.py` because I need to locate the locations of the unknown labels (label < 0) in `partial_labels`. I also corrected the remains of `unittest` in `test_constraints.py`.

### DIFF
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -70,7 +70,7 @@ class Constraints(object):
       warnings.warn("Only generated %d %s constraints (requested %d)" % (
           len(ab), 'positive' if same_label else 'negative', num_constraints))
     ab = np.array(list(ab)[:num_constraints], dtype=int)
-    return self.known_label_idx[ab.T]
+    return known_label_idx[ab.T]
 
   def chunks(self, num_chunks=100, chunk_size=2, random_state=None):
     """
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -79,12 +79,10 @@ class Constraints(object):
     random_state = check_random_state(random_state)
     chunks = -np.ones_like(self.partial_labels, dtype=int)
     uniq, lookup = np.unique(self.partial_labels, return_inverse=True)
-    if self.num_points != len(self.known_labels):
-      unknown_uniq = np.where(uniq < 0)[0]
+    unknown_uniq = np.where(uniq < 0)[0]
     all_inds = [set(np.where(lookup == c)[0]) for c in xrange(len(uniq))
                 if c not in unknown_uniq]
     max_chunks = int(np.sum([len(s) // chunk_size for s in all_inds]))
-    print(max_chunks)
     if max_chunks < num_chunks:
       raise ValueError(('Not enough possible chunks of %d elements in each'
                         ' class to form expected %d chunks - maximum number'
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -23,6 +23,7 @@ class Constraints(object):
     self.num_points, = partial_labels.shape
     self.known_label_idx, = np.where(partial_labels >= 0)
     self.known_labels = partial_labels[self.known_label_idx]
+    self.partial_labels = partial_labels
 
   def adjacency_matrix(self, num_constraints, random_state=None):
     random_state = check_random_state(random_state)
@@ -76,9 +77,11 @@ class Constraints(object):
     the random state object to be passed must be a numpy random seed
     """
     random_state = check_random_state(random_state)
-    chunks = -np.ones_like(self.known_label_idx, dtype=int)
-    uniq, lookup = np.unique(self.known_labels, return_inverse=True)
-    all_inds = [set(np.where(lookup == c)[0]) for c in xrange(len(uniq))]
+    chunks = -np.ones_like(self.partial_labels, dtype=int)
+    uniq, lookup = np.unique(self.partial_labels, return_inverse=True)
+    if self.num_points != len(self.known_labels):
+      uniq = uniq[uniq >= 0]
+    all_inds = [set(np.where(lookup == c)[0]) for c in uniq]
     max_chunks = int(np.sum([len(s) // chunk_size for s in all_inds]))
     if max_chunks < num_chunks:
       raise ValueError(('Not enough possible chunks of %d elements in each'
--- a/metric_learn/constraints.py
+++ b/metric_learn/constraints.py
@@ -20,10 +20,8 @@ class Constraints(object):
   def __init__(self, partial_labels):
     '''partial_labels : int arraylike, -1 indicating unknown label'''
     partial_labels = np.asanyarray(partial_labels, dtype=int)
-    self.num_points, = partial_labels.shape
-    self.known_label_idx, = np.where(partial_labels >= 0)
-    self.known_labels = partial_labels[self.known_label_idx]
     self.partial_labels = partial_labels
+    self.num_points, = partial_labels.shape
 
   def adjacency_matrix(self, num_constraints, random_state=None):
     random_state = check_random_state(random_state)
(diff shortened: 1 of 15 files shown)


metric_learn/constraints.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, partial_labels):
     self.num_points, = partial_labels.shape
     self.known_label_idx, = np.where(partial_labels >= 0)
```
bellet: we are now storing redundant information into the class attributes. I think it would be better to simply save `partial_labels` and reconstruct the other arrays in the methods whenever needed

metric_learn/constraints.py
```diff
@@ -20,10 +20,8 @@ class Constraints(object):
   def __init__(self, partial_labels):
     '''partial_labels : int arraylike, -1 indicating unknown label'''
     partial_labels = np.asanyarray(partial_labels, dtype=int)
-    self.num_points, = partial_labels.shape
-    self.known_label_idx, = np.where(partial_labels >= 0)
-    self.known_labels = partial_labels[self.known_label_idx]
     self.partial_labels = partial_labels
+    self.num_points, = partial_labels.shape
```
bellet: I think we can also remove this, which is used in only one place

RobinVogel:
I will write the two tests.

The tests of code coverage did not pass because of the fact that the `adjacency_matrix(self, num_constraints, random_state=None)` is not tested in `constraints.py`, and I added a line in the function (negative coverage difference). I will look into writing a test for that as well.

test/metric_learn_test.py
```diff
@@ -1136,6 +1136,19 @@ def test_changed_behaviour_warning_random_state(self):
...
+    y2 = np.concatenate((y[:n], np.ones(n)))
+
+    rca = RCA_Supervised(num_chunks=2)
+    rca.fit(X[:n], y[:n])
+
+    rca2 = RCA_Supervised(num_chunks=2)
+    rca2.fit(X, y2)
+
+    np.testing.assert_array_equal(rca.components_, rca2.components_)
+
```
bellet: There seems to be some problems with this test:
- Currently there is no unlabeled points. On line 1142 I think you want `-np.ones(n)` and not `np.ones(n)`
- The reason why the test passes nevertheless is because RCA raises a warning "The inner covariance matrix is not invertible" and returns `rca.components_` with all nans in both cases. As suggested by the warning you may need to  first reduce the dimension using PCA

bellet:
> I wrote, the one on RCA in `tests/metric_learn_test.py` in the class `TestRCA`, and
> the one on the constraints in `tests/constraints.py`. I am interested in understanding why you chose these tests.

The test on RCA is a natural one given the change made in this PR: it checks that unlabeled points are ignored. Note also that the test would fail in the previous version (since RCA would crash in the presence of unlabeled points). Generally when fixing a bug one should always write a test that would fail before to make sure the bug is not introduced again later without notice.

The test to check that no unlabeled points are assigned to any chunks is related to the fact that the change in this PR keeps entries corresponding to unlabeled points in the chunk array. Therefore it makes sense to check that no such point was actually assigned to a chunk.

RobinVogel:
I spent some thinking to understand why I always have a covariance matrix that is composed of `NaN`'s in the context of my tests. The default value of `chunk_size` in `RCA_supervised` is 2, and at some point, one substract the average to the points in each chunks, e.g. if I have `x1` and `x_2` I replace them with `(x1-x2)/2` and `(x2-x1)/2`. It means that with chunks of size 2, the maximum rank of my covariance matrix is `num_chunks`, which might be below the number of features when doing small scale tests.

bellet:
> I spent some thinking to understand why I always have a covariance matrix that is composed of `NaN`'s in the context of my tests. This is caused by two points stacked up:
> 
>     * The default parameters for scikit-learn's `make_classification` involve a `n_redundant=2` parameter, which means that one of the features is a linear combination of the others. So in that case, the algorithm will fit the data, but throw a useless NaN matrix. Using nothing but informative parameters solves this problem, see [the documentation](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.make_classification.html) for more details.
> 
>     * The default value of `chunk_size` in `RCA_supervised` is 2, and at some point, one substract the average to the points in each chunks, e.g. if I have `x1` and `x_2` I replace them with `(x1-x2)/2` and `(x2-x1)/2`. It means that with chunks of size 2, the maximum rank of my covariance matrix is `num_chunks`, which might be below the number of features when doing small scale tests.
> 
> 
> I think both points should be addressed with proper errors. Should I raise an issue, or try and address those in this PR ?

Good observations.

Note that we already have a warning which is raised in this case, advising the user to first reduce the dimension of the data:
```
UserWarning: The inner covariance matrix is not invertible, so the transformation matrix may contain Nan values. You should reduce the dimensionality of your input,for instance using `sklearn.decomposition.PCA` as a preprocessing step.
```

Regarding to improving the error:
- IMO we should not check whether some features are linear combinations of others, this is too costly (and can be fixed by reducing the dimension). This is the responsability of the user.
- It seems that what you describe for `chunk_size=2` generalizes to larger chunk sizes, in the sense that we can have a general bound on the rank of the matrix as a function of `chunk_size` and `num_chunks`, i.e. when this is the case we know for sure it will fail based only on input parameters. If this is indeed the case then I agree we should throw a specific error and explain the bound to the user so she can adjust the parameters.

bellet:
For the first point, we could simply adapt the current error:
```
UserWarning: The inner covariance matrix is not invertible, so the transformation matrix may contain Nan values. You should remove any linearly dependent features and/or reduce the dimensionality of your input, for instance using `sklearn.decomposition.PCA` as a preprocessing step.
```

test/metric_learn_test.py
RobinVogel: Done

RobinVogel:
I added the remark with the bound on the rank.

Issue #262, opened 2019-11-13
[DOC] docs break with new sphinx


wdevazelhes:
See comment https://github.com/scikit-learn-contrib/metric-learn/issues/258#issuecomment-553482800. 

When building with the new sphinx version (2.1.1), references does not display prettily, though it works with some previous versions like 1.8, we should either make it work with the new version (preferably I guess), or add in the documentation which version to use

terrytangyuan:
I agree that making it work with the new version would be preferred since there is a major upgrade of sphinx.

bellet:
Fixed in #265

PR #265, opened 2019-11-19
Correct display of references problem in Sphinx 2.1.1, solves issue #262

RobinVogel:
I found some info [there](https://github.com/sphinx-doc/sphinx/issues/6705) and added the workaround parameter to the `doc/conf.py` file to solve the problem. The new displaying of references requires modifying the CSS. The modification fixes issue #262


perimosocordiae:
Thanks, merged. Once the theme we're using updates with better CSS we should remove this workaround.

Issue #255, opened 2019-10-29
More informative error when calling transform of unfitted metric learner


bellet:
We should throw an informative error/exception when one attempts to use an unfitted transformer to transform data (or any other method which requires the metric learner to be fitted). See below what happens in metric-learn versus sklearn. We can probably borrow some code from sklearn to handle this similarly.

Example for metric-learn:
```python
>>> from metric_learn import NCA
>>> from sklearn.datasets import make_classification
>>> X, y = make_classification()
>>> nca = NCA()
>>> nca.transform(X)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/aurelien/Documents/research/github/metric-learn/metric_learn/base_metric.py", line 244, in transform
    preprocessor=self.preprocessor_,
AttributeError: 'NCA' object has no attribute 'preprocessor_'
```

Example for sklearn:
```python
>>> from sklearn.decomposition import PCA
>>> from sklearn.datasets import make_classification
>>> X, y = make_classification()
>>> pca = PCA()
>>> pca.transform(X)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/aurelien/anaconda3/lib/python3.7/site-packages/sklearn/decomposition/base.py", line 125, in transform
    check_is_fitted(self, ['mean_', 'components_'], all_or_any=all)
  File "/home/aurelien/anaconda3/lib/python3.7/site-packages/sklearn/utils/validation.py", line 914, in check_is_fitted
    raise NotFittedError(msg % {'name': type(estimator).__name__})
sklearn.exceptions.NotFittedError: This PCA instance is not fitted yet. Call 'fit' with appropriate arguments before using this method.
```

bellet:
Done in #267

PR #267, opened 2019-11-28
More systematic checks that an estimator was fit before using its parameters

RobinVogel:
Solves #255


terrytangyuan:
@RobinVogel It seems like this PR might have caused master branch to fail. Could you take a look? https://travis-ci.org/scikit-learn-contrib/metric-learn/builds/620639370

RobinVogel:
Looking into it. The tests fail only on python 3 with skggm.
I had to install lapack before I could install skggm with pip with `conda install -c conda-forge lapack`,
followed by `pip install skggm`. For now I can't replicate the error of the travis log locally.

bellet:
It is strange because all tests passed on merge:
https://travis-ci.org/scikit-learn-contrib/metric-learn/builds/618537553?utm_source=github_status&utm_medium=notification

test/test_pairs_classifiers.py
```diff
@@ -133,10 +133,25 @@ def fit(self, pairs, y):
     pairs, y = self._prepare_inputs(pairs, y,
                                     type_of_inputs='tuples')
     self.components_ = np.atleast_2d(np.identity(pairs.shape[2]))
-    self.threshold_ = 'I am not set.'
+    # self.threshold_ is not set.
     return self
 
 
+def test_unset_threshold():
+  # test that set_threshold indeed sets the threshold
```
bellet: this description is wrong

test/test_pairs_classifiers.py
```diff
@@ -73,7 +73,7 @@ def test_raise_not_fitted_error_if_not_fitted(estimator, build_dataset,
   estimator.set_params(preprocessor=preprocessor)
   set_random_state(estimator)
   with pytest.raises(NotFittedError):
-    estimator.predict(input_data)
+    estimator.decision_function(input_data)
```
bellet: a few things here:
- description of the test still mentions "predict"
- maybe all relevant methods should be tested here? it would have detected that there was a problem for `score_pairs`

bellet:
I have re-run Travis to make sure it was not a glitch - but the errors are still there

test/test_pairs_classifiers.py
RobinVogel: Addressed in PR #269, sorry for the mishaps.

test/test_pairs_classifiers.py
RobinVogel: Addressed in PR #269, sorry for the mishap.

bellet:
Fixing the failing tests in #270 
The reason the tests passed in the PR but failed on master after merging is because the new version of sklearn was released in between the two! Pretty big coincidence

PR #269, opened 2019-12-10
Corrects the forgotten bits of PR #267

RobinVogel:
See title and PR #267.

### DIFF
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -215,10 +215,7 @@ class MahalanobisMixin(six.with_metaclass(ABCMeta, BaseMetricLearner,
     :ref:`mahalanobis_distances` : The section of the project documentation
       that describes Mahalanobis Distances.
     """
-    if sys.version_info.major < 3 or sys.version_info.minor < 5:
-        check_is_fitted(self, ['preprocessor_'])
-    else:
-        check_is_fitted(self)
+    check_is_fitted(self, ['preprocessor_'])
     pairs = check_input(pairs, type_of_inputs='tuples',
                         preprocessor=self.preprocessor_,
                         estimator=self, tuple_size=2)
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -215,6 +215,7 @@ class MahalanobisMixin(six.with_metaclass(ABCMeta, BaseMetricLearner,
     :ref:`mahalanobis_distances` : The section of the project documentation
       that describes Mahalanobis Distances.
     """
+    check_is_fitted(self, ['preprocessor_'])
     pairs = check_input(pairs, type_of_inputs='tuples',
                         preprocessor=self.preprocessor_,
                         estimator=self, tuple_size=2)
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -93,6 +93,8 @@ class BaseMetricLearner(six.with_metaclass(ABCMeta, BaseEstimator)):
       The checked input labels array.
     """
     self._check_preprocessor()
+
+    check_is_fitted(self, ['preprocessor_'])
     return check_input(X, y,
                        type_of_inputs=type_of_inputs,
                        preprocessor=self.preprocessor_,
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -216,7 +216,10 @@ class MahalanobisMixin(six.with_metaclass(ABCMeta, BaseMetricLearner,
     :ref:`mahalanobis_distances` : The section of the project documentation
       that describes Mahalanobis Distances.
     """
-    check_is_fitted(self, ['preprocessor_'])
+    if sys.version_info.major < 3 or sys.version_info.minor < 5:
+        check_is_fitted(self, ['preprocessor_'])
+    else:
+        check_is_fitted(self)
     pairs = check_input(pairs, type_of_inputs='tuples',
                         preprocessor=self.preprocessor_,
                         estimator=self, tuple_size=2)
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -340,11 +340,7 @@ class _PairsClassifierMixin(BaseMetricLearner):
     y_predicted : `numpy.ndarray` of floats, shape=(n_constraints,)
       The predicted learned metric value between samples in every pair.
     """
-    # TODO: remove when we stop supporting Python < 3.5
-    if sys.version_info.major < 3 or sys.version_info.minor < 5:
-      check_is_fitted(self, 'preprocessor_')
-    else:
-      check_is_fitted(self)
+    check_is_fitted(self, 'preprocessor_')
 
     if "threshold_" not in vars(self):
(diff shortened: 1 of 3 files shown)


bellet:
Thanks, LGTM.

Before merging we should fix the mysterious failure that arises. Note that in this PR the failed tests return "assert 3 == 0" (so I guess 3 warnings are raised) while I think before we had "assert 2 == 0"

One way to investigate this if you cannot reproduce the error could be to (temporarily) change the failing test so as to print the warnings in the Travis log?

test/test_pairs_classifiers.py
```diff
@@ -66,14 +66,24 @@ def test_predict_monotonous(estimator, build_dataset,
...
+  with pytest.raises(NotFittedError):
+    estimator.score_pairs(input_data)
   with pytest.raises(NotFittedError):
     estimator.decision_function(input_data)
+  with pytest.raises(NotFittedError):
+    estimator.get_metric()
+  with pytest.raises(NotFittedError):
+    estimator.transform(input_data)
+  with pytest.raises(NotFittedError):
+    estimator.get_mahalanobis_matrix()
```
bellet: actually:
- you should also add `calibrate_threshold` here
- shouldn't we also include `predict` at the end (after calling `set_threshold` so that the threshold is set)?
- also, isn't a single `with` enough?

RobinVogel: * I did not add `calibrate_threshold` because it calls `decision_function`, but I added it now.
* Ok, done.
* I don't agree, since with all the different `with`'s we test whether each test returns the right error, while one `with` would check whether an error is raised and if it's the right one. If any of the tested methods passes without an error, several `with`'s detect that, while only one doesn't.

bellet: Indeed, my mistake :-)

metric_learn/base_metric.py
```diff
@@ -215,6 +215,7 @@ def score_pairs(self, pairs):
     :ref:`mahalanobis_distances` : The section of the project documentation
       that describes Mahalanobis Distances.
     """
+    check_is_fitted(self, ['preprocessor_'])
```
bellet: to update as in #270

bellet:
Some tests are still failing..
If we cannot do otherwise maybe we can simply copy the former `check_is_fitted` function from sklearn so that we can enforce specific attributes to be set

RobinVogel:
> Some tests are still failing..
> If we cannot do otherwise maybe we can simply copy the former `check_is_fitted` function from sklearn so that we can enforce specific attributes to be set

After some investigation, I noticed that the problem was that the function `check_input` in `metric_learn/base_metric.py` will modify the result of the test with `sklearn.utils.validation.check_is_fitted`. I think that it stems from the fact that this function defines a parameter `preprocessor_` on the estimator which violates the assumption of scikit-learn that an object with parameters ending or starting with `_` indicate a fitted estimator.

In that context, I believe that the right fix is to correct `check_input`'s behavior and why it needs that `preprocessor_` element. I'll look into that.

bellet:
Thanks. I agree with your solution to prevent the user to set the threshold on an unfitted estimator - one additional reason is that fitting will anyway change the threshold through the call to `calibrate_threshold`.

@RobinVogel Could you quickly double check that all calls to `check_if_fitted` are actually needed? It now feels like some of them might be redundant

@wdevazelhes Would you mind providing an additional review to make sure these changes are consistent with the previous state of things?

RobinVogel:
I checked it, they are all necessary, there are three cases:
- a fitted parameter is called later in the function,
- we set the threshold or check that the estimator is fitted before using the threshold (predict, set_threshold),
- `calibrate_threshold`, which sets the attribute `preprocessor_` in the function. If I don't test those, the tests in the calls to `decision_function` pass and I get a `self.components_` doesn't exists error.

RobinVogel:
Ok, one needs to keep in mind that some tests check the number of warnings in scikit-learn, and since I  have a former version locally it said something of that type, which raised an `AssertionError` in the testing process:
```
{message : FutureWarning('Passing attributes to check_is_fitted is deprecated and will be removed in 0.23. The attributes argument is ignored.'), category : 'FutureWarning', filename : '/home/robin/anaconda3/lib/python3.7/site-packages/sklearn/utils/validation.py', lineno : 933, line : None}
```
But it's all good for travis.

bellet:
> Ok, one needs to keep in mind that some tests check the number of warnings in scikit-learn, and since I have a former version locally it said something of that type, which raised an `AssertionError` in the testing process:
> 
> ```
> {message : FutureWarning('Passing attributes to check_is_fitted is deprecated and will be removed in 0.23. The attributes argument is ignored.'), category : 'FutureWarning', filename : '/home/robin/anaconda3/lib/python3.7/site-packages/sklearn/utils/validation.py', lineno : 933, line : None}
> ```
> 
> But it's all good for travis.

Yes, I think this is fine as this problem only appears for version 0.22.0

bellet:
> I checked it, they are all necessary, there are three cases:
> 
>     * a fitted parameter is called later in the method,
> 
>     * we set the threshold or check that the estimator is fitted before using the threshold (predict, set_threshold),
> 
>     * `calibrate_threshold`, which sets the attribute `preprocessor_` in the function. If I don't test those, the tests in the calls to `decision_function` pass and I get a "`self.components_` doesn't exists error."

Naive question: since `check_is_fitted` accepts attributes again, can we reduce the number of calls by requiring the presence of more attributes in some of the checks?

bellet:
> On the other hand something that could be a good middle ground could be to `check_is_fitted` an attribute just before using it. It might still lead to redundancies but not too much (e.g.
> 
> ```python
> def m1(self):
>     self.m2()
>     self.m3() #v3 is checked inside m3 because it's used inside m3, so we don't need to check_is_fitted(self, v3_) here in m1 in the line above this comment, and therefore we save one use of check_is_fitted
>     check_is_fitted(self, v1_) #redundant: useless because already checked with m2, but it's easier to read to keep it here
>     self.v1 += 1
> 
>   def m2(self):
>     check_is_fitted(self, v1_)
>     self.v1_ += 1
> 
>   def m3(self):
>       check_is_fitted(self, v3_)
>       self.v3_ += 1
> ```
> 
> ). And it would be simple/readable in a way because each `check_is_fitted` would be done locally where we need it.

This sounds like a good idea actually. I agree it is better to prioritize readability over avoiding redundancies.

@RobinVogel maybe you can quickly go through the methods and adjust the calls to `check_if_fitted` accordingly whenever needed? Then we are good to merge

RobinVogel:
I agree with the convention and I added a missing `check_is_fitted` to reflect that.

However, I think there's a legitimate couple exceptions when ones needs to check that an estimator is fitted, or setting check that a threshold exists, before  a threshold:
in the file `metric_learn/base_metric.py`:
- L. 342 in `predict`: I want to check that the estimator is fitted before I check the existence of a threshold, since one has to fit before he thresholds, but I don't want to call `self.decision_function` if there's no threshold, since it's wasteful,
- L 422 in `set_threshold`: we forbid people to set the threshold to a unfitted estimator,
- L

PR #270, opened 2019-12-12
Fix failing tests in last build

bellet:
The problem is due to a deprecation warning for `check_is_fitted` in the new version of sklearn (see below)

This PR fixes this by not using the attributes argument when calling `check_is_fitted`.

### DIFF
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -240,14 +240,14 @@ class MahalanobisMixin(six.with_metaclass(ABCMeta, BaseMetricLearner,
     X_embedded : `numpy.ndarray`, shape=(n_samples, n_components)
       The embedded data points.
     """
-    check_is_fitted(self, ['preprocessor_', 'components_'])
+    check_is_fitted(self)
     X_checked = check_input(X, type_of_inputs='classic', estimator=self,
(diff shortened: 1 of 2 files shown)


bellet:
The warning raised is from sklearn (see https://github.com/scikit-learn/scikit-learn/pull/14545):
`Passing attributes to check_is_fitted is deprecated and will be removed in 0.23. The attributes argument is ignored.`

`check_is_fitted` now only checks the presence of any fitted attribute (ending with a trailing underscore)

It should be fine for us to simply remove the passed attributes. Will do

bellet:
Removing the passed attributes does make the Python > 3.4 tests pass but makes the other fail, as sklearn dropped support for Python < 3.5.

One way to fix this would be to keep the attributes and relax the tests to let this particular warning go through, but this will make the code break when sklearn releases its next version.

So I would favor a solution based on checking Python version with `sys.version_info` and call `check_is_fitted` accordingly. What do you think?

BTW this is an example of the problems we are likely to see more and more as we continue to support Python 2 while sklearn does not. I think we should also drop support for Python < 3.5 in the next release, see also discussion in #166

bellet:
I implemented the strategy proposed above and all tests are now passing, so I think it is ready to merge.

Issue #237, opened 2019-07-23
[DOC] RCA should maybe be in "Supervised algorithms"


wdevazelhes:
Currently `RCA` is in Pairs Learners in the doc, where it shouldn't. It should probably be in "Supervised"

bellet:
I agree it shouldn't be in pairs learners. It is not supervised either though (only need to know that groups of points share the same label, not the label itself). We could put it there if we do not want to create a new section in the doc for learning from chunklets (which will contain only RCA).

bellet:
Note also that we are currently inconsistent as RCA appears under Supervised Learning Algorithms in the Package Contents section f the doc:
http://contrib.scikit-learn.org/metric-learn/metric_learn.html#supervised-learning-algorithms

PR #274, opened 2020-01-14
Changes in documentation. Rephrasing, fixed examples, standarized notation, etc.

grudloff:


doc/supervised.rst
```diff
@@ -235,7 +235,7 @@ the sum of probability of being correctly classified:
 
 Local Fisher Discriminant Analysis (:py:class:`LFDA <metric_learn.LFDA>`)
 
-`LFDA` is a linear supervised dimensionality reduction method. It is
+`LFDA` is a linear supervised dimensionality reduction method witch effectively combines the ideas of `Linear Discriminant Analysis <https://en.wikipedia.org/wiki/Linear_discriminant_analysis>` and Locality-Preserving Projection . It is
```
perimosocordiae: witch -> which

doc/getting_started.rst
```diff
@@ -20,7 +20,8 @@ Alternately, download the source repository and run:
 **Optional dependencies**
 
 - For SDML, using skggm will allow the algorithm to solve problematic cases
-  (install from commit `a0ed406 <https://github.com/skggm/skggm/commit/a0ed406586c4364ea3297a658f415e13b5cbdaf8>`_).
+  (install from commit `a0ed406 <https://github.com/skggm/skggm/commit/a0ed406586c4364ea3297a658f415e13b5cbdaf8>`_). 
+- ``pip install ''git+https://github.com/skggm/skggm.git@a0ed406586c4364ea3297a658f415e13b5cbdaf8''`` to install the required version of skggm from PyPl.
```
perimosocordiae: Single quotes here, rather than `''`.

doc/index.rst
```diff
@@ -52,7 +52,7 @@ Documentation outline
 
    auto_examples/index
 
-:ref:`genindex` | :ref:`modindex` | :ref:`search`
+:ref:`genindex` | :doc:`Modules <./metric_learn>` | :ref:`search`
```
perimosocordiae: What does this do?

doc/getting_started.rst
```diff
@@ -10,7 +10,7 @@ Run ``pip install metric-learn`` to download and install from PyPI.
 Alternately, download the source repository and run:
 
 -  ``python setup.py install`` for default installation.
--  ``python setup.py test`` to run all tests.
+-  ``pytest test`` to run all tests.
```
bellet: Indeed it seems that `python setup.py test` has been deprecated:
https://setuptools.readthedocs.io/en/latest/setuptools.html#test-build-package-and-run-a-unittest-suite
They now recommend to use `tox` instead, which seems quite heavy

So I am happy with the proposed solution of simply suggesting to run `pytest`

doc/getting_started.rst
bellet: This could also be added to the README

doc/supervised.rst
```diff
@@ -261,18 +261,18 @@ where
     \,\,\mathbf{A}_{i,j}(1/n-1/n_l) \qquad y_i = y_j\end{aligned}\right.\\
 
 here :math:`\mathbf{A}_{i,j}` is the :math:`(i,j)`-th entry of the affinity
-matrix :math:`\mathbf{A}`:, which can be calculated with local scaling methods.
+matrix :math:`\mathbf{A}`:, which can be calculated with local scaling methods, `n` and `n_l` are the total number of points and the number of points per cluster `l`.
```
bellet: add "respectively"

metric_learn/lsml.py
```diff
@@ -290,6 +292,16 @@ class LSML_Supervised(_BaseLSML, TransformerMixin):
...
 
+  Examples
+  --------
+  >>> from metric_learn import ITML_Supervised
+  >>> from sklearn.datasets import load_iris
+  >>> iris_data = load_iris()
+  >>> X = iris_data['data']
+  >>> Y = iris_data['target']
+  >>> itml = ITML_Supervised(num_constraints=200)
+  >>> itml.fit(X, Y)
```
bellet: this should be LSML_Supervised, not ITML_Supervised

metric_learn/rca.py
```diff
@@ -62,13 +62,16 @@ class RCA(MahalanobisMixin, TransformerMixin):
...
+  >>> from metric_learn import RCA
+  >>> pairs = [[[1.2, 7.5], [1.3, 1.5]],
+  >>>         [[6.4, 2.6], [6.2, 9.7]],
+  >>>         [[1.3, 4.5], [3.2, 4.6]],
+  >>>         [[6.2, 5.5], [5.4, 5.4]]]
+  >>> y = [1, 1, -1, -1]
+  >>> # in this task we want points where the first feature is close to be
+  >>> # closer to each other, no matter how close the second feature is
+  >>> rca = RCA()
+  >>> rca.fit(pairs, y)
```
bellet: This example is wrong: RCA learns from chunks, not from pairs

grudloff: I took this example from the doc/weakly_supervised.rst RCA section. I will fix them both then.

grudloff: Then it would also be necessary to make a new section in Weakly Supervised Metric Learning called "Learning on chunks"? Right now RCA is under "Learning on pairs".

doc/index.rst
bellet: This page is already linked a bit above on the index page, so perhaps we could simply remove it?
Or, fix the problem preventing the generation of this file

metric_learn/rca.py
bellet: Yes, there is an issue on that, see #237 and comments therein
Since the idea of chunks is very specific to RCA, with a small abuse of language we were considering listing it under the supervised algorithms, even though the supervision is weaker

README.rst
```diff
@@ -26,6 +26,7 @@ metric-learn contains efficient Python implementations of several popular superv
 
 - For SDML, using skggm will allow the algorithm to solve problematic cases
   (install from commit `a0ed406 <https://github.com/skggm/skggm/commit/a0ed406586c4364ea3297a658f415e13b5cbdaf8>`_).
+- ``pip install 'git+https://github.com/skggm/skggm.git@a0ed406586c4364ea3297a658f415e13b5cbdaf8'`` to install the required version of skggm from PyPl.
```
bellet: - Please do not create a new list item as this is still about the skggm optional dependency
- This does not install from PyPI but from GitHub

doc/getting_started.rst
```diff
@@ -20,7 +20,8 @@ Alternately, download the source repository and run:
 **Optional dependencies**
 
 - For SDML, using skggm will allow the algorithm to solve problematic cases
-  (install from commit `a0ed406 <https://github.com/skggm/skggm/commit/a0ed406586c4364ea3297a658f415e13b5cbdaf8>`_).
+  (install from commit `a0ed406 <https://github.com/skggm/skggm/commit/a0ed406586c4364ea3297a658f415e13b5cbdaf8>`_). 
+- ``pip install 'git+https://github.com/skggm/skggm.git@a0ed406586c4364ea3297a658f415e13b5cbdaf8'`` to install the required version of skggm from PyPl.
```
bellet: same here

metric_learn/rca.py
grudloff: I think that it should remain in the weak learners section for consistency's sake, as there is a *_supervised version, even though the supervision is not so weak. Would it be worth it to make a section just for RCA learning on chunks?

doc/index.rst
grudloff: I am not knowledgeable in sphinx so I can't find a quick fix, in the meantime, I will remove it.

metric_learn/rca.py
bellet: Sure, why not, although it is unlikely to apply to another algorithm other than RCA in the future
If you are willing to do this, please go ahead. Probably a separate PR would be best for clarity

grudloff: Alright then, in the meantime I fixed the RCA examples as they are.

bellet:
LGTM! Thanks for the changes and congrats for your first contribution @grudloff :-)

PR #277, opened 2020-01-24
Fix covariance initialization when matrix is not invertible

### LINKED ISSUE #276: Error on covariance initialization when there are linearly dependent dimensions
#### Description
While fitting on MMC_Supervised with init="covariance" on the MNIST dataset, a warning is raised and an error is thrown. I tracked down the problem to the initialization which is done with the inverse of the covariance matrix, which isn't invertible in this case because there are some LD dimensions. 

This can be easily fixed by replacing by the (Moore-Penrose) pseudo-inverse as it is done for the random initialization. I am not sure this would be necessarily the desired behavior, but in any case, this issue should be addressed at least whit a more user-friendly error stating that there are some linearly dependent dimensions and that the input should be reduced to eliminate this redundancy.

Haven't checked but this issue should arise on any algorithm using the covariance initialization that doesn't require an SDP matrix.

#### Steps/Code to Reproduce
```python
from metric_learn import MMC_Supervised
from sklearn import datasets

# Load digits dataset
digits = datasets.load_digits()
X = digits.data
y = digits.target

mmc=MMC_Supervised(init='covariance')
mmc.fit(X,y)
```

#### Expected Results
No error or a more user-friendly error.

#### Actual Results
The following warnings are thrown during runtime.
```
/home/gabrielrudloff/anaconda3/lib/python3.7/site-packages/metric_learn/_util.py:718: RuntimeWarning: divide by zero encountered in true_divide
  M = np.dot(u / s, u.T)
/home/gabrielrudloff/anaconda3/lib/python3.7/site-packages/metric_learn/_util.py:718: RuntimeWarning: invalid value encountered in true_divide
  M = np.dot(u / s, u.T)
```
And this is the error thrown, which states that the matrix is not symetric. The computed matrix is full of nan/inf.
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-272efe10621c> in <module>
      9 
     10 mmc=MMC_Supervised(init='covariance')
---> 11 mmc.fit(X,y)

~/anaconda3/lib/python3.7/site-packages/metric_learn/mmc.py in fit(self, X, y, random_state)
    609                                         random_state=self.random_state)
    610     pairs, y = wrap_pairs(X, pos_neg)
--> 611     return _BaseMMC._fit(self, pairs, y)

~/anaconda3/lib/python3.7/site-packages/metric_learn/mmc.py in _fit(self, pairs, y)
     63       return self._fit_diag(pairs, y)
     64     else:
---> 65       return self._fit_full(pairs, y)
     66 
     67   def _fit_full(self, pairs, y):

~/anaconda3/lib/python3.7/site-packages/metric_learn/mmc.py in _fit_full(self, pairs, y)
    186     self.n_iter_ = cycle
    187 
--> 188     self.components_ = components_from_metric(self.A_)
    189     return self
    190 

~/anaconda3/lib/python3.7/site-packages/metric_learn/_util.py in components_from_metric(metric, tol)
    398   """
    399   if not np.allclose(metric, metric.T):
--> 400     raise ValueError("The input metric should be symmetric.")
    401   # If M is diagonal, we will just return the elementwise square root:
    402   if np.array_equal(metric, np.diag(np.diag(metric))):

ValueError: The input metric should be symmetric.
```

#### Versions

Linux-5.3.0-26-generic-x86_64-with-debian-buster-sid
Python 3.7.4 (default, Aug 13 2019, 20:35:49) 
[GCC 7.3.0]
NumPy 1.17.2
SciPy 1.3.1
Scikit-Learn 0.21.3
Metric-Learn 0.5.0

grudloff:
This PR fixes #276, an issue that arises in the context of covariance initialization on an algorithm that doesn't require a PD matrix. It can happen that the calculated matrix is singular and in consequence isn't invertible, leading to non-explicative warnings. This is fixed by the use of a pseudo-inverse when a PD matrix is not required.

### DIFF
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -719,9 +719,9 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
                         "require the `{}` to be strictly positive definite."
                         .format(*((matrix_name,) * 2)))
     elif not cov_is_definite:
-      warnings.warn('The initialization matrix is not invertible: '
+      warnings.warn('The covariance matrix is not invertible: '
                     'using the pseudo-inverse instead.'
-                    'To make the inverse covariance matrix invertible'
+                    'To make the covariance matrix invertible'
                     ' you can remove any linearly dependent features and/or '
                     'reduce the dimensionality of your input, '
                     'for instance using `sklearn.decomposition.PCA` as a '
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -707,15 +707,19 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
       X = input
     # atleast2d is necessary to deal with scalar covariance matrices
     M_inv = np.atleast_2d(np.cov(X, rowvar=False))
-    s, u = scipy.linalg.eigh(M_inv)
-    cov_is_definite = _check_sdp_from_eigen(s)
-    if strict_pd and not cov_is_definite:
-      raise LinAlgError("Unable to get a true inverse of the covariance "
-                        "matrix since it is not definite. Try another "
-                        "`{}`, or an algorithm that does not "
-                        "require the `{}` to be strictly positive definite."
-                        .format(*((matrix_name,) * 2)))
-    M = np.dot(u / s, u.T)
+    if strict_pd:
+      s, u = scipy.linalg.eigh(M_inv)
+      cov_is_definite = _check_sdp_from_eigen(s)
+      if not cov_is_definite:
+        raise LinAlgError("Unable to get a true inverse of the covariance "
+                          "matrix since it is not definite. Try another "
+                          "`{}`, or an algorithm that does not "
+                          "require the `{}` to be strictly positive definite."
+                          .format(*((matrix_name,) * 2)))
+      else:
+        M = np.dot(u / s, u.T)
+    else:
+      M = pinvh(M_inv)
     if return_inverse:
       return M, M_inv
     else:
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -1,5 +1,4 @@
 import numpy as np
-import scipy
 import six
 from numpy.linalg import LinAlgError
 from sklearn.datasets import make_spd_matrix
@@ -8,7 +7,7 @@ from sklearn.utils import check_array
 from sklearn.utils.validation import check_X_y, check_random_state
 from .exceptions import PreprocessorError, NonPSDError
 from sklearn.discriminant_analysis import LinearDiscriminantAnalysis
-from scipy.linalg import pinvh
+from scipy.linalg import pinvh, eigh
 import sys
 import time
 
@@ -679,7 +678,7 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
   random_state = check_random_state(random_state)
   M = init
   if isinstance(init, np.ndarray):
-    s, u = scipy.linalg.eigh(init)
+    s, u = eigh(init)
     init_is_definite = _check_sdp_from_eigen(s)
     if strict_pd and not init_is_definite:
       raise LinAlgError("You should provide a strictly positive definite "
@@ -708,7 +707,7 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
     # atleast2d is necessary to deal with scalar covariance matrices
     M_inv = np.atleast_2d(np.cov(X, rowvar=False))
     if strict_pd:
-      s, u = scipy.linalg.eigh(M_inv)
+      s, u = eigh(M_inv)
       cov_is_definite = _check_sdp_from_eigen(s)
       if not cov_is_definite:
         raise LinAlgError("Unable to get a true inverse of the covariance "
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -688,8 +688,8 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
                         "require the {} to be strictly positive definite."
                         .format(*((matrix_name,) * 3)))
     elif return_inverse and not init_is_definite:
-      warnings.warn('The initialization matrix is not invertible, '
-                    'but this isn"t an issue as the pseudo-inverse is used.')
+      warnings.warn('The initialization matrix is not invertible: '
+                    'using the pseudo-inverse instead.')
     if return_inverse:
       M_inv = _pseudo_inverse_from_eig(w, V)
       return M, M_inv
@@ -719,9 +719,10 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
                         "require the `{}` to be strictly positive definite."
                         .format(*((matrix_name,) * 2)))
     elif not cov_is_definite:
-      warnings.warn('The inverse covariance matrix is not invertible, '
-                    'but this isn"t an issue as the pseudo-inverse is used. '
-                    'You can remove any linearly dependent features and/or '
+      warnings.warn('The initialization matrix is not invertible: '
+                    'using the pseudo-inverse instead.'
+                    'To make the inverse covariance matrix invertible'
+                    ' you can remove any linearly dependent features and/or '
                     'reduce the dimensionality of your input, '
                     'for instance using `sklearn.decomposition.PCA` as a '
                     'preprocessing step.')
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -774,7 +774,7 @@ def _pseudo_inverse_from_eig(w, V, tol=None):
 
   Returns
   -------
-  output : (…, M, N) array_like
+  output : (..., M, N) array_like
     The pseudo-inverse given by the EVD.
   """
   if tol is None:
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -677,8 +677,8 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
 
   random_state = check_random_state(random_state)
   M = init
-  if isinstance(init, np.ndarray):
-    s, u = eigh(init)
+  if isinstance(M, np.ndarray):
+    s, u = eigh(M)
     init_is_definite = _check_sdp_from_eigen(s)
     if strict_pd and not init_is_definite:
       raise LinAlgError("You should provide a strictly positive definite "
@@ -687,7 +687,7 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
                         "require the {} to be strictly positive definite."
                         .format(*((matrix_name,) * 3)))
     if return_inverse:
-      M_inv = np.dot(u / s, u.T)
+      M_inv = pinvh(M)
       return M, M_inv
     else:
       return M
--- a/metric_learn/_util.py
+++ b/metric_learn/_util.py
@@ -7,7 +7,7 @@ from sklearn.utils import check_array
 from sklearn.utils.validation import check_X_y, check_random_state
 from .exceptions import PreprocessorError, NonPSDError
 from sklearn.discriminant_analysis import LinearDiscriminantAnalysis
-from scipy.linalg import pinvh
+from scipy.linalg import pinvh, eigh
 import sys
 import time
 import warnings
@@ -679,8 +679,8 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
   random_state = check_random_state(random_state)
   M = init
   if isinstance(M, np.ndarray):
-    U, s, Vh = np.linalg.svd(M, full_matrices=False, hermitian=True)
-    init_is_definite = _check_sdp_from_eigen(s)
+    w, V = eigh(M, check_finite=False)
+    init_is_definite = _check_sdp_from_eigen(w)
     if strict_pd and not init_is_definite:
       raise LinAlgError("You should provide a strictly positive definite "
                         "matrix as `{}`. This one is not definite. Try another"
@@ -691,7 +691,7 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
       warnings.warn('The initialization matrix is not invertible, '
                     'but this isn"t an issue as the pseudo-inverse is used.')
     if return_inverse:
-      M_inv = _pseudo_inverse_from_svd(U, s, Vh)
+      M_inv = _pseudo_inverse_from_eig(w, V)
       return M, M_inv
     else:
       return M
@@ -710,8 +710,8 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
       X = input
     # atleast2d is necessary to deal with scalar covariance matrices
     M_inv = np.atleast_2d(np.cov(X, rowvar=False))
-    U, s, Vh = np.linalg.svd(M_inv, full_matrices=False, hermitian=True)
-    cov_is_definite = _check_sdp_from_eigen(s)
+    w, V = eigh(M_inv, check_finite=False)
+    cov_is_definite = _check_sdp_from_eigen(w)
     if strict_pd and not cov_is_definite:
       raise LinAlgError("Unable to get a true inverse of the covariance "
                         "matrix since it is not definite. Try another "
@@ -725,7 +725,7 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
                     'reduce the dimensionality of your input, '
                     'for instance using `sklearn.decomposition.PCA` as a '
                     'preprocessing step.')
-    M = _pseudo_inverse_from_svd(U, s, Vh)
+    M = _pseudo_inverse_from_eig(w, V)
     if return_inverse:
       return M, M_inv
     else:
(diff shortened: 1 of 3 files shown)


bellet:
Thanks! Could you briefly describe the fix in the PR text, and confirm that it solves the broken piece of code in the issue?

metric_learn/_util.py
```diff
@@ -707,15 +707,19 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
...
-                        .format(*((matrix_name,) * 2)))
-    M = np.dot(u / s, u.T)
+    if strict_pd:
+      s, u = scipy.linalg.eigh(M_inv)
+      cov_is_definite = _check_sdp_from_eigen(s)
+      if not cov_is_definite:
+        raise LinAlgError("Unable to get a true inverse of the covariance "
+                          "matrix since it is not definite. Try another "
+                          "`{}`, or an algorithm that does not "
+                          "require the `{}` to be strictly positive definite."
```
wdevazelhes: Maybe we could also add a third option for the user here: to reduce the input to avoid linear dependencies, as stated in the description of #276 ? Similarly to https://github.com/scikit-learn-contrib/metric-learn/blob/1b40c3b6210a80408054f8b68fdca078906a804b/metric_learn/rca.py#L99-L104

wdevazelhes:
For the test, maybe you can extend/take inspiration from https://github.com/scikit-learn-contrib/metric-learn/blob/1e42acb64ddfbc1607b4917511615ae9c2bc93aa/test/test_mahalanobis_mixin.py#L572 (see comment https://github.com/scikit-learn-contrib/metric-learn/issues/276#issuecomment-578134260)

metric_learn/_util.py
```diff
@@ -707,15 +707,19 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
...
-    cov_is_definite = _check_sdp_from_eigen(s)
-    if strict_pd and not cov_is_definite:
-      raise LinAlgError("Unable to get a true inverse of the covariance "
-                        "matrix since it is not definite. Try another "
-                        "`{}`, or an algorithm that does not "
-                        "require the `{}` to be strictly positive definite."
-                        .format(*((matrix_name,) * 2)))
-    M = np.dot(u / s, u.T)
+    if strict_pd:
+      s, u = scipy.linalg.eigh(M_inv)
```
terrytangyuan: You can simply import `eigh` instead of the whole `scipy` library?

metric_learn/_util.py
```diff
@@ -707,15 +707,19 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
...
+      cov_is_definite = _check_sdp_from_eigen(s)
+      if not cov_is_definite:
```
terrytangyuan: Add a test case to exercise the issue you are trying to fix.

metric_learn/_util.py
grudloff: Yes, I totally agree!

metric_learn/_util.py
grudloff: yes sure, did it that way just because it was done previously like that.

metric_learn/_util.py
grudloff: Of course! I will build a test based on the suggestion of @wdevazelhes

grudloff:
Looking at `_utils.py` found that the same issue can arise from custom matrix initialization `init=M` as there is no check for invertibility. Probably can be handled whit the warning @wdevazelhes suggested.

bellet:
> Looking at `_utils.py` found that the same issue can arise from custom matrix initialization `init=M` as there is no check for invertibility. Probably can be handled with the warning @wdevazelhes suggested.

Good catch. We should return the pseudo-inverse there as well when `return_inverse=True`

grudloff:
> > Looking at `_utils.py` found that the same issue can arise from custom matrix initialization `init=M` as there is no check for invertibility. Probably can be handled with the warning @wdevazelhes suggested.
> 
> Good catch. We should return the pseudo-inverse there as well when `return_inverse=True`

Since we are not going to need the eigenvalue & eigenvectors to know positive definitiveness, would it be a good idea to replace it by the Cholesky Factorization which is a more efficient way to do so?

bellet:
Note that since we always compute `M = np.dot(u / s, u.T)` we could reuse `u` and `s` to compute the pseudo inverse directly instead of wasting this work by calling `pinvh(M)`

grudloff:
> Note that since we always compute `M = np.dot(u / s, u.T)` we could reuse `u` and `s` to compute the pseudo inverse directly instead of wasting this work by calling `pinvh(M)`

You are right! I can compute the SVD and with that I can check the eigenvalues and also calculate the pseudo inverse. I will implement it that way then.

grudloff:
This last commit broke two tests, but I suspect they are not well set as they ask for a `nonPSDError` from MMC and MMC_Supervised which shouldn't throw one as MMC doesn't ask for PSD.

bellet:
> > Note that since we always compute `M = np.dot(u / s, u.T)` we could reuse `u` and `s` to compute the pseudo inverse directly instead of wasting this work by calling `pinvh(M)`
> 
> You are right! I can compute the SVD and with that I can check the eigenvalues and also calculate the pseudo inverse. I will implement it that way then.

Why switch from `eigh` to `svd`? You should keep `eigh` (which will throw an error if the matrix is not symmetric) and the eigenvalue check, and simply replace the call to `pinv` by the computation of the pseudo-inverse from `s` and `u`

bellet:
> This last commit broke two tests, but I suspect they are not well set as they ask for a `nonPSDError` from MMC and MMC_Supervised which shouldn't throw one as MMC doesn't ask for PSD.

What do you mean? MMC does require PSD matrices (like any Mahalanobis metric learning algorithm). Maybe you meant it does not require strictly PD?

grudloff:
> > > Note that since we always compute `M = np.dot(u / s, u.T)` we could reuse `u` and `s` to compute the pseudo-inverse directly instead of wasting this work by calling `pinvh(M)`
> > 
> > 
> > You are right! I can compute the SVD and with that, I can check the eigenvalues and also calculate the pseudo inverse. I will implement it that way then.
> 
> Why switch from `eigh` to `svd`? You should keep `eigh` (which will throw an error if the matrix is not symmetric) and the eigenvalue check, and simply replace the call to `pinv` by the computation of the pseudo-inverse from `s` and `u`

Is it possible to get the pseudoinverse from the eigenvalue decomposition? I looked at the implementation of `pinv` and it was done with  'svd', and since it also outputs the (square roots of the) eigenvalues one can also check for positive definiteness.

grudloff:
> > This last commit broke two tests, but I suspect they are not well set as they ask for a `nonPSDError` from MMC and MMC_Supervised which shouldn't throw one as MMC doesn't ask for PSD.
> 
> What do you mean? MMC does require PSD matrices (like any Mahalanobis metric learning algorithm). Maybe you meant it does not require strictly PD?

My bad! That is what I meant. But I was wrong as the error raises when there are negative eigenvalues. I suspect the problem should be in the tolerance of `_check_sdp_from_eigen` as I am using the root squared eigenvalues from the `svd`, maybe I should modify it. The failed tests are the following:
```
FAILED test/test_mahalanobis_mixin.py::test_init_mahalanobis[MMC] - Failed: DID NOT RAISE <class 'metric_learn.exceptions.NonPSDError'>
FAILED test/test_mahalanobis_mixin.py::test_init_mahalanobis[MMC_Supervised] - Failed: DID NOT RAISE <class 'metric_learn.exceptions.NonPSDError'>
```
Any idea about the other failed test? Seems like an outdated version of numpy is being used.

grudloff:
SVD loses the sign information from the eigenvalues so PSD cannot be checked with them, this clearly explains the failed tests. I think I will have to roll back las commit then. Sorry for the confusion!

bellet:
> Is it possible to get the pseudoinverse from the eigenvalue decomposition? I looked at the implementation of `pinv` and it was done with `svd`, and since it also outputs the (square roots of the) eigenvalues one can also check for positive definiteness.

Yes since the SVD is the same as the eigenvalue decomposition for a symmetric matrix. Switching back to `eigh` and computing the pseudo-inverse from should fix the failures you are getting

grudloff:
> > Is it possible to get the pseudoinverse from the eigenvalue decomposition? I looked at the implementation of `pinv` and it was done with `svd`, and since it also outputs the (square roots of the) eigenvalues one can also check for positive definiteness.
> 
> Yes since the SVD is the same as the eigenvalue decomposition for a symmetric matrix. Switching back to `eigh` and computing the pseudo-inverse from should fix the failures you are getting

Yes, thank you! I was just looking at that.

bellet:
For computing the pseudo-inverse, you should simply follow what is done in `pinvh` (which actually calls `eigh`), including the tolerance strategy:
https://github.com/scipy/scipy/blob/v1.4.1/scipy/linalg/basic.py#L1394-L1470

metric_learn/_util.py
```diff
@@ -678,17 +678,20 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
...
+    init_is_definite = _check_sdp_from_eigen(w)
     if strict_pd and not init_is_definite:
       raise LinAlgError("You should provide a strictly positive definite "
                         "matrix as `{}`. This one is not definite. Try another"
                         " {}, or an algorithm that does not "
                         "require the {} to be strictly positive definite."
                         .format(*((matrix_name,) * 3)))
+    elif return_inverse and not init_is_definite:
+      warnings.warn('The initialization matrix is not invertible, '
+                    'but this isn"t an issue as the pseudo-inverse is used.')
```
bellet: instead of the second part maybe simply "using the pseudo-inverse instead"

metric_learn/_util.py
```diff
@@ -678,17 +678,20 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
 
   random_state = check_random_state(random_state)
   M = init
-  if isinstance(init, np.ndarray):
-    s, u = scipy.linalg.eigh(init)
-    init_is_definite = _check_sdp_from_eigen(s)
+  if isinstance(M, np.ndarray):
+    w, V = eigh(M, check_finite=False)
```
bellet: why did you turn off `check_finite`? I think it is better to keep it

metric_learn/_util.py
bellet: maybe simply "The initialization matrix is not invertible: using the pseudo-inverse instead"

metric_learn/_util.py
```diff
@@ -707,15 +710,22 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
...
     if strict_pd and not cov_is_definite:
       raise LinAlgError("Unable to get a true inverse of the covariance "
                         "matrix since it is not definite. Try another "
                         "`{}`, or an algorithm that does not "
                         "require the `{}` to be strictly positive definite."
                         .format(*((matrix_name,) * 2)))
-    M = np.dot(u / s, u.T)
+    elif not cov_is_definite:
+      warnings.warn('The inverse covariance matrix is not invertible, '
+                    'but this isn"t an issue as the pseudo-inverse is used. '
```
bellet: same here.

metric_learn/_util.py
```diff
@@ -707,15 +710,22 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
...
+                    'You can remove any linearly dependent features and/or '
```
bellet: for clarity, maybe start this sentence by "To make the inverse covariance matrix invertible, "

metric_learn/_util.py
grudloff: `check_array() ` does it previously, so it shouldn't be necessary to do it again.

metric_learn/_util.py
grudloff: Thanks! I was looking for a better way to phrase that.

grudloff:
> correctly



> Things look good - you mainly need to write tests and we're good to go. You should definitely write at least one test for your `_pseudo_inverse_from_eig` function, and tests for the non-invertible cases.
> 
> Also it would be nice to check whether we have one for when the initialized matrix is not symmetric. According to https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.eigh.html :
> "If eigenvalue computation does not converge, an error occurred, or b matrix is not definite positive. Note that if input matrices are not symmetric or hermitian, no error is reported but results will be wrong."
> 
> So it is not clear what happens for us and whether we correctly deal with this

Great! I am making the tests now, should be done soon.

I don't think there should be an issue with non-symmetric matrices as this is checked for the array init case and the covariance matrix should always be symmetric. I think there is a missing test for the symmetry checking of array init that should be implemented.

metric_learn/_util.py
bellet: Indeed!

bellet:
Indeed, I had missed the symmetry check for `init` - nevermind!

bellet:
(but in seems that we are not: https://codecov.io/gh/scikit-learn-contrib/metric-learn/src/master/metric_learn/_util.py#L671)

grudloff:
> (but in seems that we are not: https://codecov.io/gh/scikit-learn-contrib/metric-learn/src/master/metric_learn/_util.py#L671)

Great! With the other tests this should be ready then.

metric_learn/_util.py
```diff
@@ -719,9 +719,10 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
                         "require the `{}` to be strictly positive definite."
                         .format(*((matrix_name,) * 2)))
     elif not cov_is_definite:
-      warnings.warn('The inverse covariance matrix is not invertible, '
-                    'but this isn"t an issue as the pseudo-inverse is used. '
-                    'You can remove any linearly dependent features and/or '
+      warnings.warn('The initialization matrix is not invertible: '
```
bellet: initialization matrix --> covariance matrix

metric_learn/_util.py
```diff
@@ -719,9 +719,10 @@ def _initialize_metric_mahalanobis(input, init='identity', random_state=None,
                         "require the `{}` to be strictly positive definite."
                         .format(*((matrix_name,) * 2)))
     elif not cov_is_definite:
-      warnings.warn('The inverse covariance matrix is not invertible, '
-                    'but this isn"t an issue as the pseudo-inverse is used. '
-                    'You can remove any linearly dependent features and/or '
+      warnings.warn('The initialization matrix is not invertible: '
+                    'using the pseudo-inverse instead.'
+                    'To make the inverse covariance matrix invertible'
```
bellet: the inverse covariance --> the covariance

test/test_mahalanobis_mixin.py
```diff
@@ -603,6 +603,45 @@ def test_singular_covariance_init_or_prior(estimator, build_dataset):
...
+                          if idml[:3] in ['MMC']],
+                         ids=[idml for idml, (ml, _)
+                              in zip(ids_metric_learners,
+                                     metric_learners)
+                              if idml[:3] in ['MMC']])
+def test_singular_covariance_init_of_non_strict_pd(estimator, build_dataset):
+    """Tests that when using the 'covariance' init or prior, it returns the
+    appropriate warning if the covariance matrix is singular, for algorithms
+    that don't need a strictly PD init. Also checks that the returned
+    inverse matrix has finite values
```
bellet: This is not what you check, right? You seem to check the learned `M`.
Any reason you do this and not a call to `_initialize_metric_mahalanobis` like in the other test? Maybe you could check both things in both tests?

test/test_mahalanobis_mixin.py
```diff
@@ -654,6 +693,31 @@ def test_singular_array_init_or_prior(estimator, build_dataset, w0):
...
+    inverse matrix is finite. This isn't checked for model fitting as no
+    model curently uses this setting.
+    """
+    rng = np.random.RandomState(42)
+    X, y = shuffle(*make_blobs(random_state=rng),
+                   random_state=rng)
+    P = ortho_group.rvs(X.shape[1], random_state=rng)
+    w = np.abs(rng.randn(X.shape[1]))
+    w[0] = w0
+    M = P.dot(np.diag(w)).dot(P.T)
```
bellet: this seems a bit convoluted just to generated a singular matrix? You could just draw a rectangular matrix `L` and construct `M` as `L.T.dot(L)`?

test/test_mahalanobis_mixin.py
grudloff: Yes, you are totally right! Forgot to set the max iteration in order to make it stop before the first iteration.

I called `_initialize_metric_mahalanobis` directly in the other test because currently there is no algorithm that could reproduce that setting. But setting `max_iter = 0` should keep it from iterating and would result in the `M` from the initialization.

test/test_mahalanobis_mixin.py
grudloff: I agree with you, just did it that way because it was done like that on `test_singular_array_init_or_prior_strictpd`. But also this way you have only one zero eigenvalue and control its value directly with the parametrization.

bellet: Sounds good

test/test_utils.py
```diff
@@ -1146,3 +1147,25 @@ def test__auto_select_init(has_classes, n_features, n_samples, n_components,
   """Checks that the auto selection of the init works as expected"""
   assert (_auto_select_init(has_classes, n_features,
                             n_samples, n_components, n_classes) == result)
+
+
+@pytest.mark.parametrize('w0', [1e-20, 0., -1e-20])
+def test_pseudo_inverse_from_eig_and_pinvh_singular(w0):
+  """Checks that _pseudo_inverse_from_eig return the same result as
+  scipy.linalg.pinvh for a singular matrix"""
+  A = np.random.rand(100, 100)
```
bellet: You should fix the random seed

test/test_utils.py
```diff
@@ -1146,3 +1147,25 @@ def test__auto_select_init(has_classes, n_features, n_samples, n_components,
...
+  w[0] = w0
+  A = V.dot(np.diag(w)).dot(V.T)
+  np.testing.assert_allclose(_pseudo_inverse_from_eig(w, V), pinvh(A),
+                             rtol=1e-05)
+
+
+def test_pseudo_inverse_from_eig_and_pinvh_nonsingular():
+  """Checks that _pseudo_inverse_from_eig return the same result as
+  scipy.linalg.pinvh for a non singular matrix"""
+  A = np.random.rand(100, 100)
```
bellet: Same here

grudloff:
> LGTM, nice work @grudloff !

Thanks!

bellet:
Nitpick: calling `np.random.seed` modifies the global random state. To avoid affecting other parts of the code I I think it is better to create a separate pseudo-random number generator:
```
rng = np.random.RandomState(SEED)
A = rng.rand(100, 100)
```

bellet:
Thanks @grudloff ! Merging

PR #278, opened 2020-02-14
SCML : Sparse Compositional Metric Learning

grudloff:
Sparse Compositional Metric Learning (SCML) allows scalable learning of global, multi-task and multiple local Mahalanobis metrics for multi-class data under a unified framework based on sparse combinations of rank-one basis metrics. For this initial merge, only the global setting will be implemented.

The algorithm learns on triplets, so it is necessary to add the base class for the addition of this kind of algorithm. For the sake of clarity, this will be added in a separate concurrent PR.

This implementation follows closely the [matlab implementation](https://github.com/bellet/SCML) of Y. Shi, A. Bellet and F. Sha. Sparse Compositional Metric Learning. AAAI Conference on Artificial Intelligence (AAAI), 2014. [SCML paper](http://researchers.lille.inria.fr/abellet/papers/aaai14.pdf)

**Theory - Global setting**

The Mahalanobis Matrix is constructed as a sum of rank-1 PSD matrices:

The basis are intended to be locally discriminative. In the original paper and in this implementation they are constructed with LDA of several local regions. There are other options to construct this basis that will be added later.

The constrains are a set of triplets `C` that are inforced through the minimization problem:

**Advantages** 

- No need to project to SPD cone as M is constructed as a nonegative sum of SPD matrices.
- Only `n_basis` parameters must be learned and stored, without taking into account the basis.
- Uses stochastic composite optimization and in consequence, big datasets can be handled by this algorithm. An efficient implementation of regularized dual averaging is used for the optimization procedure.
- Faster to Train.
- Good Performance on standard datasets (see [paper](http://researchers.lille.inria.fr/abellet/papers/aaai14.pdf) for detailed information) especially on multi-task and local settings. A Benchmark against the other algorithms of the package will be added to this PR later. 
- First algorithm to learn on triplets to be implemented in this package.

**Tests on Vehicles dataset**
The results for the vehicles dataset found in the [matlab implementation github](https://github.com/bellet/SCML) are used to validate the consistency and the correctness of the current implementation.
On each of the tests, the algorithm was run 100 times, the mean and std of the resulted accuracy in test and train, as well as the time, are shown below each test link.
For the batch versions, the numbers of iterations are reduced to have the same amount of gradient computations for each method.

[Test Vanilla](https://gist.github.com/grudloff/29fcb5d529706893c6c33dcd1e4c2aee)
```
Time. Mean:  5.7989197754859925 std:  0.05547399180223441
Train - Accuracy. Mean:  79.3214990138067 std:  0.7612083562308343
Test - Accuracy. Mean:  77.71176470588235 std:  2.056124280977084
```
[Test Batch](https://gist.github.com/grudloff/67208cbc86b9bbc484ba5ac7c64dc07f)
```
Time. Mean:  6.090397052764892 std:  0.048224950065408285
Train - Accuracy. Mean:  81.41222879684418 std:  0.9372967012148159
Test - Accuracy. Mean:  78.0 std:  1.6946894459868151
```
[Test Adagrad](https://gist.github.com/grudloff/73862709ae0a43020cc0a8a02beed9f5)
```
Time. Mean:  6.090397052764892 std:  0.048224950065408285
Train - Accuracy. Mean:  81.41222879684418 std:  0.9372967012148159
Test - Accuracy. Mean:  78.0 std:  1.6946894459868151
```
[Test Batch+Adagrad](https://gist.github.com/grudloff/dc43bf78c7daeecaca8ab4cfc5184ba4)
```
Time. Mean:  1.7653330016136168 std:  0.015989019692575546
Train - Accuracy. Mean:  81.31558185404339 std:  0.9613659886786882
Test - Accuracy. Mean:  78.03529411764706 std:  1.641208095836975
```
We can see that the result are almost the same and even with little variance improvements over the test accuracy. Also the use of mini-batches yields almost a 4-fold improvement on the time used. 

Furthermore, the adagrad version has a faster convergence than the "vanilla" version as it can be observed of the results obtained with 1/20 the number of iterations, as observed on the achieved train acuracy. But this comes with an apparent tradeoff as the test accuracy of the vanilla version is a little bit better, this suggest that maybe it is a good idea to allow both options.
[Test Vanilla - 1/20 iterations](https://gist.github.com/grudloff/ec4bddf365a72e4489c1ac9a6845cf7e)
```
Time. Mean:  1.2541892004013062 std:  0.02197305793469499
Train - Accuracy. Mean:  79.0 std:  0.8888682213828464
Test - Accuracy. Mean:  77.37647058823529 std:  1.5683284028668087
```
[Test Adagrad - 1/20 iterations](https://gist.github.com/grudloff/c5c2fb559c245ad860fa3de9ff0c1941)
```
Time. Mean:  1.3263736057281494 std:  0.020214300581072035
Train - Accuracy. Mean:  80.65483234714003 std:  1.1301895624652614
Test - Accuracy. Mean:  77.0764705882353 std:  2.0688077237684213
```

TODO: 
- [x] Code Comentaries and docstrings
- [x] Documentation
- [x] Basic Tests
- [ ] Specific Tests
- [ ] Benchmark

### DIFF
--- a/doc/weakly_supervised.rst
+++ b/doc/weakly_supervised.rst
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
 Algorithms
 ----------
 
+.. _scml:
+
+:py:class:`SCML <metric_learn.SCML>`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Sparse Compositional Metric Learning
+(:py:class:`SCML <metric_learn.SCML>`)
+
+`SCML` learns an squared mahalanobis distance from triplet constraints by 
+optimizing sparse positive weights assigned to a set of :math:`K` locally discriminative 
+rank-one PSD bases. This can be formulated as an optimization problem with only :math:`K`
+parameters, that can be solved with an efficient stochastic composite scheme.
+
+The Mahalanobis Matrix :math:`M` is built from a basis set :math:`B = \{b_i\}_{i=\{1,...,K\}}`
+weighted by a :math:`K` dimensional vector :math:`w = \{w_i\}_{i=\{1,...,K\}}` as:
+
+.. math::
+
+    M = \sum_{i=1}^K w_i b_i b_i^T = B \cdot diag(w) \cdot B^T \quad w_i \geq 0
+
+Learning :math:`M` in this form makes it PSD by design, as it is a nonnegative sum of PSD matrices.
+The optimization problem of :math:`w` over the triplets constraints :math:`C` is formulated as a
+classic margin-based hinge loss function over the relative constrains, a regularization :math:`\ell_1`
+is added to yield an sparse representation. The formulation is the following:
+
+.. math::
+
+    \min_{w} \sum_{(x_a,x_b,x_c)\in C} [1 + d_w(x_a,x_b)-d_w(x_a,x_c)]_+ + \beta||w||_1
+
+Where :math:`[\cdot]_+` is the hinge loss. 
+ 
+.. topic:: Example Code:
+
+::
+
+    from metric_learn import SCML
+
+    triplets = [[[1.2, 7.5], [1.3, 1.5], [6.2, 9.7]],
+                [[1.3, 4.5], [3.2, 4.6], [5.4, 5.4]],
+                [[3.2, 7.5], [3.3, 1.5], [8.2, 9.7]],
+                [[3.3, 4.5], [5.2, 4.6], [7.4, 5.4]]]
+
+    scml = SCML()
+    scml.fit(triplets)
+
+.. topic:: References:
+
+  .. [1] Y. Shi, A. Bellet and F. Sha. `Sparse Compositional Metric Learning.
+         <http://researchers.lille.inria.fr/abellet/papers/aaai14.pdf>`_. \
+         (AAAI), 2014.
+
+  .. [2] Adapted from original \
+         `Matlab implementation.<https://github.com/bellet/SCML>`_.
+
 
 .. _learning_on_quadruplets:
 
--- a/doc/weakly_supervised.rst
+++ b/doc/weakly_supervised.rst
@@ -708,28 +708,31 @@ Algorithms
 Sparse Compositional Metric Learning
 (:py:class:`SCML <metric_learn.SCML>`)
 
-`SCML` learns an squared mahalanobis distance from triplet constraints by 
-optimizing sparse positive weights assigned to a set of :math:`K` locally discriminative 
-rank-one PSD bases. This can be formulated as an optimization problem with only :math:`K`
-parameters, that can be solved with an efficient stochastic composite scheme.
+`SCML` learns an squared Mahalanobis distance from triplet constraints by
+optimizing sparse positive weights assigned to a set of :math:`K` rank-one
+PSD bases. This can be formulated as an optimization problem with only
+:math:`K` parameters, that can be solved with an efficient stochastic
+composite scheme.
 
-The Mahalanobis Matrix :math:`M` is built from a basis set :math:`B = \{b_i\}_{i=\{1,...,K\}}`
+The Mahalanobis matrix :math:`M` is built from a basis set :math:`B = \{b_i\}_{i=\{1,...,K\}}`
 weighted by a :math:`K` dimensional vector :math:`w = \{w_i\}_{i=\{1,...,K\}}` as:
 
 .. math::
 
     M = \sum_{i=1}^K w_i b_i b_i^T = B \cdot diag(w) \cdot B^T \quad w_i \geq 0
 
-Learning :math:`M` in this form makes it PSD by design, as it is a nonnegative sum of PSD matrices.
-The optimization problem of :math:`w` over the triplets constraints :math:`C` is formulated as a
-classic margin-based hinge loss function over the relative constrains, a regularization :math:`\ell_1`
-is added to yield an sparse representation. The formulation is the following:
+Learning :math:`M` in this form makes it PSD by design, as it is a
+nonnegative sum of PSD matrices. The basis set :math:`B` is fixed on advance
+and it is possible to construct it from the data. The optimization problem
+over :math:`w` is formulated as a classic margin-based hinge loss function
+involving the set :math:`C` of triplets. A regularization :math:`\ell_1`
+is added to yield a sparse combination. The formulation is the following:
 
 .. math::
 
-    \min_{w} \sum_{(x_a,x_b,x_c)\in C} [1 + d_w(x_a,x_b)-d_w(x_a,x_c)]_+ + \beta||w||_1
+    \min_{w\geq 0} \sum_{(x_i,x_j,x_k)\in C} [1 + d_w(x_i,x_j)-d_w(x_i,x_k)]_+ + \beta||w||_1
 
-Where :math:`[\cdot]_+` is the hinge loss. 
+where :math:`[\cdot]_+` is the hinge loss. 
  
 .. topic:: Example Code:
 
@@ -883,13 +886,13 @@ extension leads to more stable estimation when the dimension is high and
 only a small amount of constraints is given.
 
 The loss function of each constraint 
-:math:`d(\mathbf{x}_a, \mathbf{x}_b) < d(\mathbf{x}_c, \mathbf{x}_d)` is 
+:math:`d(\mathbf{x}_i, \mathbf{x}_j) < d(\mathbf{x}_k, \mathbf{x}_l)` is 
 denoted as:
 
 .. math::
 
-    H(d_\mathbf{M}(\mathbf{x}_a, \mathbf{x}_b) 
-    - d_\mathbf{M}(\mathbf{x}_c, \mathbf{x}_d))
+    H(d_\mathbf{M}(\mathbf{x}_i, \mathbf{x}_j) 
+    - d_\mathbf{M}(\mathbf{x}_k, \mathbf{x}_l))
 
 where :math:`H(\cdot)` is the squared Hinge loss function defined as:
 
@@ -899,8 +902,8 @@ where :math:`H(\cdot)` is the squared Hinge loss function defined as:
     \,\,x^2 \qquad x>0\end{aligned}\right.\\
 
 The summed loss function :math:`L(C)` is the simple sum over all constraints 
-:math:`C = \{(\mathbf{x}_a , \mathbf{x}_b , \mathbf{x}_c , \mathbf{x}_d) 
-: d(\mathbf{x}_a , \mathbf{x}_b) < d(\mathbf{x}_c , \mathbf{x}_d)\}`. The 
+:math:`C = \{(\mathbf{x}_i , \mathbf{x}_j , \mathbf{x}_k , \mathbf{x}_l) 
+: d(\mathbf{x}_i , \mathbf{x}_j) < d(\mathbf{x}_k , \mathbf{x}_l)\}`. The 
 original paper suggested here should be a weighted sum since the confidence 
 or probability of each constraint might differ. However, for the sake of 
 simplicity and assumption of no extra knowledge provided, we just deploy 
@@ -912,9 +915,9 @@ knowledge:
 
 .. math::
 
-    \min_\mathbf{M}(D_{ld}(\mathbf{M, M_0}) + \sum_{(\mathbf{x}_a, 
-    \mathbf{x}_b, \mathbf{x}_c, \mathbf{x}_d)\in C}H(d_\mathbf{M}(
-    \mathbf{x}_a, \mathbf{x}_b) - d_\mathbf{M}(\mathbf{x}_c, \mathbf{x}_c))\\
+    \min_\mathbf{M}(D_{ld}(\mathbf{M, M_0}) + \sum_{(\mathbf{x}_i, 
+    \mathbf{x}_j, \mathbf{x}_k, \mathbf{x}_l)\in C}H(d_\mathbf{M}(
+    \mathbf{x}_i, \mathbf{x}_j) - d_\mathbf{M}(\mathbf{x}_k, \mathbf{x}_l))\\
 
 where :math:`\mathbf{M}_0` is the prior metric matrix, set as identity 
 by default, :math:`D_{ld}(\mathbf{\cdot, \cdot})` is the LogDet divergence:
--- a/doc/weakly_supervised.rst
+++ b/doc/weakly_supervised.rst
@@ -708,7 +708,7 @@ Algorithms
 Sparse Compositional Metric Learning
 (:py:class:`SCML <metric_learn.SCML>`)
 
-`SCML` learns an squared Mahalanobis distance from triplet constraints by
+`SCML` learns a squared Mahalanobis distance from triplet constraints by
 optimizing sparse positive weights assigned to a set of :math:`K` rank-one
 PSD bases. This can be formulated as an optimization problem with only
 :math:`K` parameters, that can be solved with an efficient stochastic
@@ -722,7 +722,7 @@ weighted by a :math:`K` dimensional vector :math:`w = \{w_i\}_{i=\{1,...,K\}}` a
     M = \sum_{i=1}^K w_i b_i b_i^T = B \cdot diag(w) \cdot B^T \quad w_i \geq 0
 
 Learning :math:`M` in this form makes it PSD by design, as it is a
-nonnegative sum of PSD matrices. The basis set :math:`B` is fixed on advance
+nonnegative sum of PSD matrices. The basis set :math:`B` is fixed in advance
 and it is possible to construct it from the data. The optimization problem
 over :math:`w` is formulated as a classic margin-based hinge loss function
 involving the set :math:`C` of triplets. A regularization :math:`\ell_1`
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -554,9 +554,9 @@ class SCML_Supervised(_BaseSCML, TransformerMixin):
       warnings.warn("The number of basis is less than the number of classes, "
                     "which may lead to poor discriminative performance.")
     elif n_basis >= X.shape[0]*2*num_eig:
-      raise ValueError("The needed number of clusters to generate the selected"
-                       "number of basis is unfeasible to achieve as it is "
-                       "greater than the number of available samples")
+      raise ValueError("Not enough samples to generate %d LDA bases, n_basis"
+                       "should be smaller than %d" %
+                       (n_basis, X.shape[0]*2*num_eig))
 
     kmeans = KMeans(n_clusters=n_clusters, random_state=self.random_state,
                     algorithm='elkan').fit(X)
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -222,9 +222,17 @@ class _BaseSCML_global(MahalanobisMixin):
     uniqPairs = np.unique(T_pairs_sorted, axis=0)
 
     if self.n_basis is None:
+      # TODO: Get a good default n_basis directive
       n_basis = uniqPairs.shape[0]
+      warnings.warn('The number of basis will be set to n_basis= %d' % n_basis)
 
-    elif self.n_basis > uniqPairs.shape[0]:
+    elif isinstance(self.n_basis, int):
+      n_basis = self.n_basis
+    else:
+      raise ValueError("n_basis should be an integer, instead it is of type %s"
+                       % type(self.n_basis))
+
+    if n_basis > uniqPairs.shape[0]:
       n_basis = uniqPairs.shape[0]
       warnings.warn("The selected number of basis is greater than the number "
                     "of points, only n_basis = %d will be generated" %
@@ -452,7 +460,7 @@ class SCML_global_Supervised(_BaseSCML_global, TransformerMixin):
     # Add other options passed as string
     authorized_basis = ['triplet_diffs']
     supervised_basis = ['LDA']
-    authorized_basis += supervised_basis
+    authorized_basis = supervised_basis + authorized_basis
 
     if not(isinstance(self.basis, np.ndarray)) \
        and self.basis not in authorized_basis:
@@ -485,18 +493,22 @@ class SCML_global_Supervised(_BaseSCML_global, TransformerMixin):
     num_eig = min(n_class-1, n_features)
 
     if self.n_basis is None:
+      # TODO: Get a good default n_basis directive
       n_basis = min(20*n_features, X.shape[0]*2*num_eig)
       warnings.warn('The number of basis will be set to n_basis= %d' % n_basis)
 
-    # n_basis must be greater or equal to n_class
-    elif self.n_basis < n_class:
-      raise ValueError("The number of basis should be greater than the"
-                       " number of classes")
     elif isinstance(self.n_basis, int):
       n_basis = self.n_basis
     else:
       raise ValueError("n_basis should be an integer, instead it is of type %s"
                        % type(self.n_basis))
+
+    if n_basis <= n_class:
+      raise ValueError("The number of basis should be greater than the"
+                       " number of classes")
+    elif n_basis >= X.shape[0]*2*num_eig:
+      raise ValueError("The selected number of basis needs a greater number of"
+                       " clusters than the number of available samples")
 
     # Number of clusters needed for 2 scales given the number of basis
     # yielded by every LDA
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -556,24 +556,21 @@ class SCML_Supervised(_BaseSCML, TransformerMixin):
 
     idx_set = np.zeros((n_clusters, sum(k_class)), dtype=np.int)
 
-    # TODO: It may be better to precompute this similarly to how it is done
-    # with the triplets generator
-    start = 0
-    finish = 0
+    start_finish_indices = np.hstack((0, k_class)).cumsum()
 
     neigh = NearestNeighbors()
 
     for c in range(n_class):
         sel_c = np.where(y == labels[c])
         kc = k_class[c]
+
         # get k_class same class neighbours
         neigh.fit(X=X[sel_c])
 
-        finish += kc
+        start, finish = start_finish_indices[c:c+2]
         idx_set[:, start:finish] = np.take(sel_c, neigh.kneighbors(X=cX,
                                            n_neighbors=kc,
                                            return_distance=False))
-        start = finish
 
     # Compute basis for every cluster in first scale
     basis = np.zeros((n_basis, n_features))
@@ -592,36 +589,35 @@ class SCML_Supervised(_BaseSCML, TransformerMixin):
 
     idx_set = np.zeros((n_clusters, sum(k_class)), dtype=np.int)
 
-    # TODO: It may be better to precompute this similarly to how it is done
-    # with the triplets generator
-    start = 0
-    finish = 0
+    start_finish_indices = np.hstack((0, k_class)).cumsum()
 
     for c in range(n_class):
-        sel_c = np.where(y == labels[c])
-        kc = k_class[c]
+      sel_c = np.where(y == labels[c])
+      kc = k_class[c]
 
-        # get k_class genuine neighbours
-        neigh.fit(X=X[sel_c])
-        finish += kc
-        idx_set[:, start:finish] = np.take(sel_c, neigh.kneighbors(X=cX,
-                                           n_neighbors=kc,
-                                           return_distance=False))
-        start = finish
+      # get k_class genuine neighbours
+      neigh.fit(X=X[sel_c])
+
+      start, finish = start_finish_indices[c:c+2]
+      idx_set[:, start:finish] = np.take(sel_c, neigh.kneighbors(X=cX,
+                                         n_neighbors=kc,
+                                         return_distance=False))
 
     # Compute basis for every cluster in second scale
     finish = num_eig * n_clusters
 
+    start_finish_indices = np.arange(num_eig * n_clusters, n_basis, num_eig)
+    start_finish_indices = np.append(start_finish_indices, n_basis)
+
     for i in range(n_clusters):
-        start = finish
-        finish += num_eig
+      try:
+        start, finish = start_finish_indices[i:i+2]
+      except ValueError:
+        # No more clusters to be yielded
+        break
 
-        # handle tail, as n_basis != n_clusters*2*n_eig
-        if finish > n_basis:
-          finish = n_basis
+      lda.fit(X[idx_set[i, :]], y[idx_set[i, :]])
 
-        lda.fit(X[idx_set[i, :]], y[idx_set[i, :]])
-
-        basis[start:finish, :] = normalize(lda.scalings_.T[:finish-start])
+      basis[start:finish, :] = normalize(lda.scalings_.T[:finish-start])
 
     return basis, n_basis
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -73,10 +73,10 @@ class _BaseSCML(MahalanobisMixin):
         # regularization part of obj function
         obj1 = np.sum(w)*self.beta
 
-      # Every triplet distance difference in the space given by L
-      # plus a slack of one
+        # Every triplet distance difference in the space given by L
+        # plus a slack of one
         slack_val = 1 + np.matmul(dist_diff, w.T)
-      # Mask of places with positive slack
+        # Mask of places with positive slack
         slack_mask = slack_val > 0
 
         # loss function of learning task part of obj function
@@ -96,13 +96,13 @@ class _BaseSCML(MahalanobisMixin):
       idx = rand_int[iter]
 
       slack_val = 1 + np.matmul(dist_diff[idx, :], w.T)
-
       slack_mask = np.squeeze(slack_val > 0, axis=1)
-      avg_grad_w = ((iter * avg_grad_w + np.sum(dist_diff[idx[slack_mask], :],
-                                                axis=0, keepdims=True))
-                    / (iter+1))
 
-      scale_f = -np.sqrt(iter+1) / (self.gamma*self.batch_size)
+      grad_w = np.sum(dist_diff[idx[slack_mask], :],
+                      axis=0, keepdims=True)/self.batch_size
+      avg_grad_w = (iter * avg_grad_w + grad_w) / (iter+1)
+
+      scale_f = -np.sqrt(iter+1) / self.gamma
 
       # proximal operator with negative trimming equivalent
       w = scale_f * np.minimum(avg_grad_w + self.beta, 0)
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -18,13 +18,15 @@ class _BaseSCML_global(MahalanobisMixin):
 
   _tuple_size = 3   # constraints are triplets
 
-  def __init__(self, beta=1e-5, basis=None, n_basis=None,
-               max_iter=100000, verbose=False,  preprocessor=None,
-               random_state=None):
+  def __init__(self, beta=1e-5, basis=None, n_basis=None, gamma=5e-3,
+               max_iter=100000, output_iter=5000, verbose=False,
+               preprocessor=None, random_state=None):
     self.beta = beta
     self.basis = basis
     self.n_basis = n_basis
+    self.gamma = gamma
     self.max_iter = max_iter
+    self.output_iter = output_iter
     self.verbose = verbose
     self.preprocessor = preprocessor
     self.random_state = random_state
@@ -47,14 +49,12 @@ class _BaseSCML_global(MahalanobisMixin):
     triplets = self._prepare_inputs(triplets, type_of_inputs='tuples')
 
     # TODO:
-    # This algorithm is build to work with indeces, but in order to be
+    # This algorithm is built to work with indices, but in order to be
     # compliant with the current handling of inputs it is converted
     # back to indices by the following function. This should be improved
     # in the future.
     triplets, X = self._to_index_points(triplets)
 
-    # TODO: should be given access to gamma?
-    gamma = 5e-3
     dist_diff = self._compute_dist_diff(triplets, X)
 
     sizeT = triplets.shape[0]
@@ -62,12 +62,10 @@ class _BaseSCML_global(MahalanobisMixin):
     w = np.zeros((1, self.n_basis))
     avg_grad_w = np.zeros((1, self.n_basis))
 
-    output_iter = 5000     # output every output_iter iterations
-
     best_obj = np.inf
 
     for iter in range(self.max_iter):
-      if (iter % output_iter == 0):
+      if (iter % self.output_iter == 0):
         # regularization part of obj function
         obj1 = np.sum(w)*self.beta
 
@@ -103,7 +101,7 @@ class _BaseSCML_global(MahalanobisMixin):
       else:
         avg_grad_w = iter * avg_grad_w / (iter+1)
 
-      scale_f = -np.sqrt(iter+1) / gamma
+      scale_f = -np.sqrt(iter+1) / self.gamma
 
       # proximal operator with negative trimming equivalent
       w = scale_f * np.minimum(avg_grad_w + self.beta, 0)
@@ -129,15 +127,15 @@ class _BaseSCML_global(MahalanobisMixin):
     # np.array (2*lenT,2)
     T_pairs_sorted = np.sort(np.vstack((T[:, [0, 1]], T[:, [0, 2]])),
                              kind='stable')
-    # calculate all unique pairs and their indeces
-    uniqPairs, indeces = np.unique(T_pairs_sorted, return_inverse=True,
+    # calculate all unique pairs and their indices
+    uniqPairs, indices = np.unique(T_pairs_sorted, return_inverse=True,
                                    axis=0)
     # calculate L2 distance acording to bases only for unique pairs
     dist = np.square(XB[uniqPairs[:, 0], :]-XB[uniqPairs[:, 1], :])
 
     # return the diference of distances between all positive and negative
     # pairs
-    return dist[indeces[:lenT]]-dist[indeces[lenT:]]
+    return dist[indices[:lenT]]-dist[indices[lenT:]]
 
   def _get_components(self, w):
     """
@@ -197,6 +195,8 @@ class SCML_global(_BaseSCML_global, _TripletsClassifierMixin):
   Read more in the :ref:`User Guide <scml>`.
   Parameters
   ----------
+  beta: float (default=1e-5)
+      L1 regularization parameter.
   basis : None, string or numpy array, optional (default=None)
        Prior to set for the metric. Possible options are
        '', and a numpy array of shape (n_basis, n_features). If
@@ -209,7 +209,13 @@ class SCML_global(_BaseSCML_global, _TripletsClassifierMixin):
       Number of basis to be yielded. In case it is not set it will be set based
       on the basis numpy array. If an string option is pased to basis an error
       wild be raised as this value will be needed.
-  max_iter : int, optional
+  gamma: float (default = 5e-3)
+      Learning rate
+  max_iter : int (default = 100000)
+      Number of iterations for the algorithm
+  output_iter : int (default = 5000)
+      Number of iterations to check current weights performance and output this
+      information in case verbose is True.
   verbose : bool, optional
       if True, prints information while learning
   preprocessor : array-like, shape=(n_samples, n_features) or callable
@@ -276,6 +282,8 @@ class SCML_global_Supervised(_BaseSCML_global, TransformerMixin):
   Read more in the :ref:`User Guide <scml>`.
   Parameters
   ----------
+  beta: float (default=1e-5)
+      L1 regularization parameter.
   basis : None, string or numpy array, optional (default=None)
       Prior to set for the metric. Possible options are
       'LDA', and a numpy array of shape (n_basis, n_features). If
@@ -293,7 +301,13 @@ class SCML_global_Supervised(_BaseSCML_global, TransformerMixin):
       Number of basis to be yielded. In case it is not set it will be set based
       on the basis numpy array. If an string option is pased to basis an error
       wild be raised as this value will be needed.
-  max_iter : int, optional
+  gamma: float (default = 5e-3)
+      Learning rate
+  max_iter : int (default = 100000)
+      Number of iterations for the algorithm
+  output_iter : int (default = 5000)
+      Number of iterations to check current weights performance and output this
+      information in case verbose is True.
   verbose : bool, optional
       if True, prints information while learning
   preprocessor : array-like, shape=(n_samples, n_features) or callable
@@ -331,8 +345,8 @@ class SCML_global_Supervised(_BaseSCML_global, TransformerMixin):
   """
 
   def __init__(self, k_genuine=3, k_impostor=10, beta=1e-5, basis=None,
-               n_basis=None, max_iter=100000, verbose=False,
-               preprocessor=None, random_state=None):
+               n_basis=None, gamma=5e-3, max_iter=100000, output_iter=5000,
+               verbose=False, preprocessor=None, random_state=None):
     self.k_genuine = k_genuine
     self.k_impostor = k_impostor
     _BaseSCML_global.__init__(self, beta=beta, basis=basis, n_basis=n_basis,
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -170,16 +170,9 @@ class _BaseSCML_global(MahalanobisMixin):
 
   def _to_index_points(self, triplets, X=None):
     shape = triplets.shape
-    if len(shape) == 3:
-      X, triplets = np.unique(np.vstack(triplets), return_inverse=True, axis=0)
-      triplets = triplets.reshape(shape[:2])
-      return triplets, X
-    elif(len(shape) == 2 and X is not None):
-      return triplets, X
-    elif(self.preprocessor is not None):
-      return triplets, self.preprocessor
-    else:
-      raise ValueError('A preprocessor is needed when triplets are indices')
+    X, triplets = np.unique(np.vstack(triplets), return_inverse=True, axis=0)
+    triplets = triplets.reshape(shape[:2])
+    return triplets, X
 
   def _initialize_basis(self, triplets, X):
     """ TODO: complete function description
@@ -232,19 +225,15 @@ class _BaseSCML_global(MahalanobisMixin):
       n_basis = uniqPairs.shape[0]
 
     elif self.n_basis > uniqPairs.shape[0]:
-      print("n_basis too big")
       n_basis = uniqPairs.shape[0]
+      warnings.warn("The selected number of basis is greater than the number "
+                    "of points, only n_basis = %d will be generated" %
+                    n_basis)
 
     else:
       n_basis = self.n_basis
 
-    if len(triplets.shape) == 3:
-      pass
-    elif X is not None:
-      uniqPairs = X[uniqPairs]
-    else:
-      raise ValueError('The processor must be set if indices are used for the'
-                       'triplets construction')
+    uniqPairs = X[uniqPairs]
 
     rng = check_random_state(self.random_state)
 
@@ -447,16 +436,37 @@ class SCML_global_Supervised(_BaseSCML_global, TransformerMixin):
     """
     X, y = self._prepare_inputs(X, y, ensure_min_samples=2)
 
-    # TODO: this should be replaced by a _initialize_bases_supervised
-    # for future adding of other approaches of basis generation
-    if(self.basis == "LDA"):
-      basis, n_basis = self._generate_bases_LDA(X, y)
+    basis, n_basis = self._initialize_basis_supervised(X, y)
 
     constraints = Constraints(y)
     triplets = constraints.generate_knntriplets(X, self.k_genuine,
                                                 self.k_impostor)
 
     return self._fit(triplets, X, basis, n_basis)
+
+  def _initialize_basis_supervised(self, X, y):
+    """ TODO: complete function description
+    """
+
+    # TODO:
+    # Add other options passed as string
+    authorized_basis = ['triplet_diffs']
+    supervised_basis = ['LDA']
+    authorized_basis += supervised_basis
+
+    if not(isinstance(self.basis, np.ndarray)) \
+       and self.basis not in authorized_basis:
+      raise ValueError(
+          "`basis` must be '{}' "
+          "or a numpy array of shape (n_basis, n_features)."
+          .format("', '".join(authorized_basis)))
+
+    if self.basis is supervised_basis[0]:
+      basis, n_basis = self._generate_bases_LDA(X, y)
+    else:
+      basis, n_basis = None, None
+
+    return basis, n_basis
 
   def _generate_bases_LDA(self, X, y):
     """
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -546,17 +546,17 @@ class SCML_Supervised(_BaseSCML, TransformerMixin):
       raise ValueError("n_basis should be an integer, instead it is of type %s"
                        % type(self.n_basis))
 
-    if n_basis < n_class:
-      warnings.warn("The number of basis is less than the number of classes,"
-                    " this will lead to less basis than the amount yielded by"
-                    " LDA")
-    elif n_basis >= X.shape[0]*2*num_eig:
-      raise ValueError("The selected number of basis needs a greater number of"
-                       " clusters than the number of available samples")
-
     # Number of clusters needed for 2 scales given the number of basis
     # yielded by every LDA
     n_clusters = int(np.ceil(n_basis/(2 * num_eig)))
+
+    if n_basis < n_class:
+      warnings.warn("The number of basis is less than the number of classes, "
+                    "which may lead to poor discriminative performance.")
+    elif n_basis >= X.shape[0]*2*num_eig:
+      raise ValueError("The needed number of clusters to generate the selected"
+                       "number of basis is unfeasible to achieve as it is "
+                       "greater than the number of available samples")
 
     kmeans = KMeans(n_clusters=n_clusters, random_state=self.random_state,
                     algorithm='elkan').fit(X)
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -22,14 +22,15 @@ class _BaseSCML(MahalanobisMixin):
   _authorized_basis = ['triplet_diffs']
 
   def __init__(self, beta=1e-5, basis='triplet_diffs', n_basis=None,
-               gamma=5e-3, max_iter=100000, output_iter=5000, verbose=False,
-               preprocessor=None, random_state=None):
+               gamma=5e-3, max_iter=100000, output_iter=5000, batch_size=10,
+               verbose=False, preprocessor=None, random_state=None):
     self.beta = beta
     self.basis = basis
     self.n_basis = n_basis
     self.gamma = gamma
     self.max_iter = max_iter
     self.output_iter = output_iter
+    self.batch_size = batch_size
     self.verbose = verbose
     self.preprocessor = preprocessor
     self.random_state = random_state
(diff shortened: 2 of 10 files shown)


metric_learn/scml.py
```diff
@@ -0,0 +1,487 @@
...
+      n_features = self.preprocessor.shape[1]
+    else:
+      n_features = self.triplets.shape[1]
+
+    self._initialize_basis(n_features)
+
+    triplets = self._prepare_inputs(triplets, type_of_inputs='tuples')
+
+    # TODO:
+    # This algorithm is build to work with indeces, but in order to be
```
terrytangyuan: build -> built

metric_learn/scml.py
```diff
@@ -0,0 +1,487 @@
...
+    # compliant with the current handling of inputs it is converted
+    # back to indices by the following function. This should be improved
+    # in the future.
+    triplets, X = self._to_index_points(triplets)
+
+    # TODO: should be given access to gamma?
+    gamma = 5e-3
```
terrytangyuan: Yes would be better to expose this than hard-coding it here

metric_learn/scml.py
```diff
@@ -0,0 +1,487 @@
...
+    dist_diff = self._compute_dist_diff(triplets, X)
+
+    sizeT = triplets.shape[0]
+
+    w = np.zeros((1, self.n_basis))
+    avg_grad_w = np.zeros((1, self.n_basis))
+
+    output_iter = 5000     # output every output_iter iterations
```
terrytangyuan: This should probably be exposed as well

perimosocordiae:
I suspect this will need a rebase now that the triplets PR is merged.

grudloff:
I edited the PR and added a test on the current implementation of this branch of the Vehicles dataset from the Matlab implementation repository. I will be adding tests on more datasets soon!

grudloff:
I added a simple basis construction for the unsupervised version in the meantime. The implementation is now consistent with other algorithm implementations, with the new basis construction for the unsupervised version it is possible to check both the normal and supervised version. The following step is to add specific tests and to add some benchmarks on other datasets!

I think the PR is now in shape for some feedback!

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+    lenT = len(T)
+    # get all positive and negative pairs with lowest index first
+    # np.array (2*lenT,2)
+    T_pairs_sorted = np.sort(np.vstack((T[:, [0, 1]], T[:, [0, 2]])),
+                             kind='stable')
+    # calculate all unique pairs and their indices
+    uniqPairs, indices = np.unique(T_pairs_sorted, return_inverse=True,
+                                   axis=0)
+    # calculate L2 distance acording to bases only for unique pairs
+    dist = np.square(XB[uniqPairs[:, 0], :]-XB[uniqPairs[:, 1], :])
```
bellet: spaces around `-` would be better?

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+
+    # return the diference of distances between all positive and negative
+    # pairs
+    return dist[indices[:lenT]]-dist[indices[lenT:]]
```
bellet: spaces around `-` would be better?

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+
+    # get rid of inactive bases
+    active_idx, = w > 0
+    w = w[..., active_idx]
+    basis = basis[active_idx, :]
+
+    K, d = basis.shape
+
+    if(K < d):  # if metric is low-rank
+      return np.sqrt(w.T)*basis  # equivalent to np.diag(np.sqrt(w)).dot(B)
```
bellet: update comment with new variable name

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+    """
+
+    # get rid of inactive bases
+    active_idx, = w > 0
+    w = w[..., active_idx]
+    basis = basis[active_idx, :]
+
+    K, d = basis.shape
+
+    if(K < d):  # if metric is low-rank
```
bellet: in this case we are currently returning a rectangular `L` of dimension `(K, d)`, correct? If so we should consider printing something to let the user know that the learned transformation reduces dimension

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+    # TODO:
+    # Add other options passed as string
+    authorized_basis = ['triplet_diffs']
+    if isinstance(self.basis, np.ndarray):
+      # TODO: should copy?
+      basis = check_array(self.basis, copy=True)
+      n_basis = basis.shape[0]
+      if basis.shape[1] != n_features:
+        raise ValueError('The input dimensionality ({}) of the given '
+                         'linear transformation `init` must match the '
```
bellet: wrong message. Should refer to the dimensionality of the provided bases

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+                         'dimensionality of the given inputs `X` ({}).'
+                         .format(basis.shape[1], n_features))
+    elif self.basis not in authorized_basis:
+      raise ValueError(
+          "`basis` must be '{}' "
+          "or a numpy array of shape (n_basis, n_features)."
+          .format("', '".join(authorized_basis)))
+    if self.basis is authorized_basis[0]:
```
bellet: not great for readability

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+
+    basis = selected_pairs[:, 0]-selected_pairs[:, 1]
+
+    return basis, n_basis
+
+
+class SCML_global(_BaseSCML_global, _TripletsClassifierMixin):
+  """Sparse Compositional Metric Learning (SCML)
+
+  `SCML` builds a metric as the sparse positive combination of a set of locally
```
bellet: builds a metric as the sparse positive combination of --> learns a metric from triplet constraints by optimizing sparse positive weights assigned to
of locally --> of `K` locally

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  discriminative rank-one PSD basis. This allows an optimization scheme with
```
bellet: basis --> bases
This allows an optimization scheme --> This can be formulated as an optimization problem

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  only `K` parameters, that can be yielded with an efficient stochastic
```
bellet: yielded --> solved

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  composite optimization over a set of triplets constraints. Each triplet is
```
bellet: optimization --> scheme
and you can remove "over a set of triplets constraints" (this is about the formulation, not the solver)

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  constructed as a relative distance comparison with respect to the first
+  element so that the second element is closer than the last.
```
bellet: I am not sure it is worth explaining what triplets are here? It does not look like we do this for other algorithms (e.g., ITML)

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  Read more in the :ref:`User Guide <scml>`.
+  Parameters
+  ----------
+  beta: float (default=1e-5)
+      L1 regularization parameter.
+  basis : None, string or numpy array, optional (default=None)
+       Prior to set for the metric. Possible options are
+       '', and a numpy array of shape (n_basis, n_features). If
+        None an error will be raised as the basis set is esential
+        to SCML.
```
bellet: ??? I guess this is temporary. I thought the `None` value was to allow the algorithm to choose a good rule-of-thumb value

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+       numpy array
+           A matrix of shape (n_basis, n_features), that will be used as
+           the basis set for the metric construction.
+  n_basis : int, optional
+      Number of basis to be yielded. In case it is not set it will be set based
+      on the basis numpy array. If an string option is pased to basis an error
+      wild be raised as this value will be needed.
```
bellet: again this is confusing but we can sort this out later

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  gamma: float (default = 5e-3)
+      Learning rate
```
bellet: add "for the optimization algorithm."

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  max_iter : int (default = 100000)
+      Number of iterations for the algorithm
```
bellet: Missing trailing dot

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  output_iter : int (default = 5000)
+      Number of iterations to check current weights performance and output this
+      information in case verbose is True.
```
bellet: not sure we should uncover this. maybe we could fix it, or choose it based on `max_iter`

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  verbose : bool, optional
+      if True, prints information while learning
```
bellet: if --> If
trailing dot

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  preprocessor : array-like, shape=(n_samples, n_features) or callable
+      The preprocessor to call to get triplets from indices. If array-like,
+      triplets will be formed like this: X[indices].
+  random_state : int or numpy.RandomState or None, optional (default=None)
+      A pseudo random number generator object or a seed for it if int.
+  Attributes
+  ----------
+  components_ : `numpy.ndarray`, shape=(n_features, n_features)
+      The linear transformation ``L`` deduced from the learned Mahalanobis
+      metric (See function `components_from_metric`.)
```
bellet: for SCML it is `_get_components`
(maybe choose a more consistent and explicit name, like `_components_from_basis_weights`)

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  Examples
+  --------
+  >>> from metric_learn import SCLM_global_Supervised
+  >>> from sklearn.datasets import load_iris
+  >>> iris_data = load_iris()
+  >>> X = iris_data['data']
+  >>> Y = iris_data['target']
+  >>> scml = SCML_global_Supervised(basis='LDA', n_basis=400)
+  >>> scml.fit(X, Y)
```
bellet: this example is for the supervised version

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+         <http://researchers.lille.inria.fr/abellet/papers/aaai14.pdf>`_. \
+         (AAAI), 2014.
+  .. [2] Adapted from original \
+         `Matlab implementation.<https://github.com/bellet/SCML>`_.
+  See Also
+  --------
+  metric_learn.SCML_global : The original weakly-supervised algorithm.
+
+  :ref:`supervised_version` : The section of the project documentation
+    that describes the supervised version of weakly supervised estimators.
```
bellet: this is for the supervised version

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  def fit(self, triplets):
+    """Learn the SCML model.
+
+    Parameters
+    ----------
+    triplets : array-like, shape=(n_constraints, 3, n_features) or \
+          (n_constraints, 3)
+      3D array-like of triplets of points or 2D array of triplets of
+      indicators. In order to supervise the algorithm in the right way, we
+      should have the three samples ordered in a way such that:
```
bellet: Be more concise, e.g., "Triplets are assumed to be ordered such that"

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+    Returns
+    -------
+    self : object
+      Returns the instance.
+    """
+
+    return self._fit(triplets)
+
+
+class SCML_global_Supervised(_BaseSCML_global, TransformerMixin):
```
bellet: Following previous comment, this should just be `SCML_Supervised`

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  """Supervised version of Sparse Compositional Metric Learning (SCML)
+
+  `SCML_global_Supervised` creates triplets by taking `k_genuine` neighbours
+  of the same class and `k_impostor` neighbours from diferent classes for each
+  point and then runs the SCML algorithm on these triplets.
+  Read more in the :ref:`User Guide <scml>`.
+  Parameters
```
bellet: Please use same line breaks convention as in docstring for other algorithms/functions
And most comments above apply here too

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  """Sparse Compositional Metric Learning (SCML)
+
+  `SCML` builds a metric as the sparse positive combination of a set of locally
+  discriminative rank-one PSD basis. This allows an optimization scheme with
+  only `K` parameters, that can be yielded with an efficient stochastic
+  composite optimization over a set of triplets constraints. Each triplet is
+  constructed as a relative distance comparison with respect to the first
+  element so that the second element is closer than the last.
+  Read more in the :ref:`User Guide <scml>`.
+  Parameters
```
bellet: Please use same line breaks convention as in docstring for other algorithms/functions

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+  np.sum([[0, 1], [1, 1]], where=[False, True], axis=1)
+except TypeError:
+  def sum_where(X, where):
+    return np.sum(X[where])
+else:
+  def sum_where(X, where):
+    return np.sum(X, where=where)
+
+
+class _BaseSCML_global(MahalanobisMixin):
```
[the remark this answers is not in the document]
grudloff: Great idea!

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+    self.n_basis = n_basis
+    self.gamma = gamma
+    self.max_iter = max_iter
+    self.output_iter = output_iter
+    self.verbose = verbose
+    self.preprocessor = preprocessor
+    self.random_state = random_state
+    super(_BaseSCML_global, self).__init__(preprocessor)
+
+  def _fit(self, triplets, X=None, basis=None, n_basis=None):
```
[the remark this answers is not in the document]
grudloff: I know, it's a bit confusing, I will try to explain it. 

This is for the case of the supervised `fit()`. In the beginning, my implementation modified `self.preprocessor` to be equal to `X`, but then I learned that `self.preprocessor`  should not be modified and that they can be different. Because of this `_prepare_inputs` doesn't have the right reference, as `triplets` are formed with reference to `X`, not `self.preprocessor`. To tackle this I choose to just pass `X` down the line so that in the future the following lines:

```
    if X is not None:
      triplets = X[triplets]
    triplets = self._prepare_inputs(triplets, type_of_inputs='tuples')
    triplets, X = self._to_index_points(triplets)
```
can be condenced in a new `_prepare_input` option like:
```
triplets, X = self._prepare_inputs(triplets, X , type_of_inputs='tuples_reference')
```
There is also another option to solve this, which is to change the reference of the triplets back to the `self.preprocessor` but for this, the reference `X` should be saved and checked before `_prepare_input` converts it to the corresponding points of the `self.preprocessor`. I found this option to be more convoluted, so I choose the other option.

I am not completely sure if this is the best way to go, so I am open for suggestions on a better way to handle this situation!

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+
+    # Currently prepare_inputs makes triplets contain points and not indices
+    triplets = self._prepare_inputs(triplets, type_of_inputs='tuples')
+
+    # TODO:
+    # This algorithm is built to work with indices, but in order to be
+    # compliant with the current handling of inputs it is converted
+    # back to indices by the following function. This should be improved
+    # in the future.
+    triplets, X = self._to_index_points(triplets, X)
```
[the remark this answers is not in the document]
grudloff: I am sorry, this is actually not the reason! That's actually what is happening inside `_to_index_points`, the `X` input is just an artifact I forgot to take out.

metric_learn/scml.py
```diff
@@ -0,0 +1,596 @@
...
+
+  def _compute_dist_diff(self, T, X, basis):
+    """
+    Helper function to compute the distance difference of every triplet in the
+    space yielded by the basis set.
+    """
+    # Transformation of data by the basis set
+    XB = np.matmul(X, basis.T)
+
+    lenT = len(T)
```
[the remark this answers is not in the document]
grudloff: yes, you are totally right!

metric_learn/scml.py
grudloff: Yes, you are correct! Should this be a warning? Also, do you think its a good idea to change `K, d` to `n_basis, n_features`?

metric_learn/scml.py
grudloff: Agreed, I was following what was done in the `init` handling on `_util.py`. I will improve it!

metric_learn/scml.py
grudloff: My bad!

metric_learn/scml.py
grudloff: My bad, haven't updated the description of `basis` yet!

metric_learn/scml.py
grudloff: I uncovered it based on @terrytangyuan suggestion. A fixed value based on `max_iter` sounds good though.

metric_learn/scml.py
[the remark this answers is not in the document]
bellet: I still think you don't need any `X`.

Why should it be different from the case of learning from pairs of quadruplets? The only thing you need in your implementation is to recover a data matrix along with triplets in the form of indices to this data matrix. While `_prepare_input` will always give you triplets in the form of features (not indices), you can easily reconstruct an data matrix and triplet indices. Even if they do not match what was given as preprocessor, who cares? This is simply for the purpose of training.

metric_learn/scml.py
bellet: Definitely. Yes, a warning sounds good I guess?

metric_learn/scml.py
bellet: We do not expose this for other algorithms: typically the verbose mode outputs each iteration. Of course these are batch algorithms - for stochastic algorithms with low per-iteration complexity, it is a waste of resources to print at each iteration. On the other hand, the purpose of verbose mode is to check what is happening, so we could consider printing every iteration, or say every 10 iter if the computational burden is too high.

metric_learn/scml.py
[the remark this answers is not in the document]
grudloff: There is no issue with that, actually `_to_index_points` doesn't recover necessarily the same `X`. The issue is that in the current way it is done if the triplets in the form of indices are passed to `_prepare_input` the reference is wrong and in consequence, the formed triplets will also be wrong!

It is perfectly possible to input the triplets already in the form of features but I am trying to think in some future new option of `_prepare_input` that will allow for a more efficient input check that doesn't require this back and forth conversion.

metric_learn/scml.py
grudloff: I agree that we should try to be consistent with other algorithms. Although I think the decision is not so trivial as on every output iteration the objective is computed with all the constraints, so there is a high computational cost compared to each normal iteration. Also, on every output iteration the best weights are updated, so reducing this value could lead to overfitting as more instances (with possible overfitting) are tested.

Maybe a reasonable default value in relation to `max_iter` sounds good?

metric_learn/scml.py
grudloff: Maybe I misunderstood you, you mean the message or the use of `authorized_basis[0]`?

bellet: I meant the latter. `authorized_basis` is convenient to check if the value is allowed, but then when dealing with each case it is more readable to use the string values directly

metric_learn/scml.py
bellet: Good point, you also use this to keep the best weight, not only for verbose purposes.
Maybe it is good to expose it then. As we may also use it to adapt the learning rate automatically

metric_learn/scml.py
grudloff: Agreed, already changed to that in the last commit!

metric_learn/scml.py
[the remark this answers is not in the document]
bellet: Do you mean that the reason for this is to allow the user to give as input triplet indices which do not refer to the `X` given as preprocessor?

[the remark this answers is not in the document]
bellet: Because otherwise I do not understand why you say "if the triplets in the form of indices are passed to `_prepare_input` the reference is wrong".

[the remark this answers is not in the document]
grudloff: No! In the supervised case, the `triplets` are constructed as indices references to `X` but `_prepare_input` rebuilds the triplets with the preprocessor, so if `X` and the preprocessor are not equal the reference is completely wrong.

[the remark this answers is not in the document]
bellet: OK, this is for the supervised case. If I'm not mistaken, in the supervised versions of other algorithms, the tuples are always passed as a 3D array (i.e., features, not indices) to the `_fit`. This avoids the above issue, and I think this is what you should do.

Of course in the future we should find a nice way to avoid materializing the 3D array and store only indices, so that algorithms that do not need to have access to all data at once (such as this one) can benefit from lower memory complexity. But this should be done in a separate PR.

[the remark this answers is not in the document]
grudloff: Ok, sounds good! I will just pass the 3D array then.

metric_learn/scml.py
```diff
@@ -0,0 +1,627 @@
...
+
+    if self.verbose:
+      print("max iteration reached.")
+
+    # return L matrix yielded from best weights
+    self.components_ = self._components_from_basis_weights(basis, best_w)
+
+    return self
+
+  def _compute_dist_diff(self, T, X, basis):
```
wdevazelhes: Maybe rename "T" to "triplets" or "trip" or something like this ? (so as not to have the same name as the transpose operator)

metric_learn/scml.py
```diff
@@ -0,0 +1,627 @@
...
+    # calculate L2 distance acording to bases only for unique pairs
+    dist = np.square(XB[uniqPairs[:, 0], :] - XB[uniqPairs[:, 1], :])
+
+    # return the diference of distances between all positive and negative
+    # pairs
+    return dist[indices[:lenT]] - dist[indices[lenT:]]
+
+  def _components_from_basis_weights(self, basis, w):
+    """
+    get components matrix (L) from computed mahalanobis matrix
```
wdevazelhes:
```suggestion
    Get components matrix (L) from computed mahalanobis matrix.
```

metric_learn/scml.py
```diff
@@ -0,0 +1,627 @@
...
+
+    # get rid of inactive bases
+    active_idx, = w > 0
+    w = w[..., active_idx]
+    basis = basis[active_idx, :]
+
+    n_basis, n_features = basis.shape
+
+    if n_basis < n_features:  # if metric is low-rank
+      warnings.warn("The number of effective basis is less than the numbert of"
```
wdevazelhes:
```suggestion
      warnings.warn("The number of effective basis is less than the number of"
```

test/test_mahalanobis_mixin.py
```diff
@@ -298,12 +299,18 @@ def test_components_is_2D(estimator, build_dataset):
       to_keep = np.where(np.abs(diffs.ravel()) > 1e-9)
       trunc_data = trunc_data[to_keep]
       labels = labels[to_keep]
+  if isinstance(estimator, _TripletsClassifierMixin):
+    for slice_idx in [[0, 1], [0, 2]]:
+      pairs = trunc_data[:, slice_idx, :]
+      diffs = pairs[:, 1, :] - pairs[:, 0, :]
+      to_keep = np.abs(diffs.ravel()) > 1e-9
```
wdevazelhes: you're right, `np.where` is not needed, so I'm thinking we could remove it too for the other two cases (`_PairsClassifierMixin` and `_QuadrupletsClassifierMixin`), also, now I realize we could probably factorize the three cases together (the only thing that change is the slice(s) to take)

grudloff: Yes! I was thinking the same. It shouldn't be difficult to put them all together.

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+    # get rid of inactive bases
+    # TODO: Maybe have a tolerance over zero?
+    active_idx, = w > 0
+    w = w[..., active_idx]
+    basis = basis[active_idx, :]
+
+    n_basis, n_features = basis.shape
+
+    if n_basis < n_features:  # if metric is low-rank
+      warnings.warn("The number of effective basis is less than the number of"
```
bellet: maybe "bases with nonzero weight" instead of "effective" is more informative

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+    on one of the available options.
+    """
+    n_features = X.shape[1]
+
+    if isinstance(self.basis, np.ndarray):
+      # TODO: should copy?
+      basis = check_array(self.basis, copy=True)
+      if basis.shape[1] != n_features:
+        raise ValueError('The dimensionality ({}) of the provided bases must'
+                         ' match the dimensionality of the given inputs `X` '
```
bellet: maybe remove the reference to `X` as in some cases the user does not give `X` but triplets
you could just say "the dimensionality of the data"

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+
+    return basis, n_basis
+
+  def _generate_bases_dist_diff(self, triplets, X):
+    """ Constructs the basis set from the differences of positive and negative
+    pairs from the triplets constraints.
+
+    The basis set is constructed iteratively by taking n_features positive and
+    n_features negative pairs differences, then adding and substracting
+    respectively all the outerproducts, and finally adding the eigenvectors
```
bellet: "adding the eigenvectors" is confusing (they are not added together)
selecting?

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+    of this matrix with positive eigenvalue.
+    """
+    n_features = X.shape[1]
+    n_triplets = triplets.shape[0]
+
+    if self.n_basis is None:
+      # TODO: Get a good default n_basis directive
+      n_basis = int(n_triplets/10)
```
bellet: I know you have not worked on this yet, but just to mention that this default value should probably be based on `n_features` as this is the dimensionality of the space

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+      n_basis = self.n_basis
+    else:
+      raise ValueError("n_basis should be an integer, instead it is of type %s"
+                       % type(self.n_basis))
+
+    if n_basis > n_triplets:
+      n_basis = n_triplets
+      warnings.warn("The selected number of basis is greater than the number "
+                    "of points, only n_basis = %d will be generated" %
+                    n_basis)
```
bellet: Is `n_basis > n_triplets` an issue?

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+
+    basis = np.zeros((n_basis, n_features))
+
+    # get all positive and negative pairs with lowest index first
+    # np.array (2*n_triplets,2)
+    triplets_pairs_sorted = np.sort(np.vstack((triplets[:, [0, 1]],
```
bellet: same question as above

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+
+    while(finish != n_basis):
+
+      # select n_features positive differences
+      d_pos = diff_pos[rng.choice(n_triplets,
+                                  size=n_features, replace=False), :]
+
+      # select n_features negative differences
+      d_neg = diff_neg[rng.choice(n_triplets,
+                                  size=n_features, replace=False), :]
```
bellet: A simpler alternative is to extract the positive and the negative pair from the same triplet (hence drawing only `n_triplets` rather than `2*n_triplets`). This probably works equally good?

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+    # get all positive and negative pairs with lowest index first
+    # np.array (2*n_triplets,2)
+    triplets_pairs_sorted = np.sort(np.vstack((triplets[:, [0, 1]],
+                                               triplets[:, [0, 2]])),
+                                    kind='stable')
+    # calculate all unique pairs and their indices
+    uniqPairs, indices = np.unique(triplets_pairs_sorted, return_inverse=True,
+                                   axis=0)
+    # calculate L2 distance according to bases only for unique pairs
+    diff = X[uniqPairs[:, 0], :] - X[uniqPairs[:, 1], :]
```
bellet: this was already computed in `_compute_dist_diff`, maybe there is a way to avoid recomputing it here?

metric_learn/scml.py
bellet: Also indent conventions (i.e., 2 spaces) ;-)

metric_learn/scml.py
bellet: +indent

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+         `Matlab implementation.<https://github.com/bellet/SCML>`_.
+
+  See Also
+  --------
+  metric_learn.SCML_Supervised : The supervised version of this
+    algorithm, which construct the triplets from the labels.
+
+  :ref:`supervised_version` : The section of the project documentation
+    that describes the supervised version of weakly supervised estimators.
+  """
```
bellet: replace with reference to the weakly supervised one

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+                       " number of classes")
+    elif n_basis >= X.shape[0]*2*num_eig:
+      raise ValueError("The selected number of basis needs a greater number of"
+                       " clusters than the number of available samples")
+
+    # Number of clusters needed for 2 scales given the number of basis
+    # yielded by every LDA
+    n_clusters = int(np.ceil(n_basis/(2 * num_eig)))
+
+    # TODO: maybe give acces to Kmeans jobs for faster computation?
```
bellet: Probably not worth it. The expert user can set the backend context so that several cores are used, see
https://scikit-learn.org/stable/glossary.html#term-n-jobs

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+    idx_set = np.zeros((n_clusters, sum(k_class)), dtype=np.int)
+
+    start_finish_indices = np.hstack((0, k_class)).cumsum()
+
+    for c in range(n_class):
+      sel_c = np.where(y == labels[c])
+      kc = k_class[c]
+
+      # get k_class genuine neighbourss
+      neigh.fit(X=X[sel_c])
```
bellet: Calls to `kneighbors` should be needed only once regardless of the number of scales, no? Using the max scale

metric_learn/scml.py
```diff
@@ -0,0 +1,647 @@
...
+
+    n_triplets = triplets.shape[0]
+    # get all positive and negative pairs with lowest index first
+    # np.array (2*n_triplets,2)
+    triplets_pairs_sorted = np.sort(np.vstack((triplets[:, [0, 1]],
+                                               triplets[:, [0, 2]])),
+                                    kind='stable')
+    # calculate all unique pairs and their indices
+    uniqPairs, indices = np.unique(triplets_pairs_sorted, return_inverse=True,
+                                   axis=0)
```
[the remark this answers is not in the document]
grudloff: That sort is atually just to have all the pairs with the lowest index first!

doc/weakly_supervised.rst
```diff
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
...
 
+.. _scml:
+
+:py:class:`SCML <metric_learn.SCML>`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Sparse Compositional Metric Learning
+(:py:class:`SCML <metric_learn.SCML>`)
+
+`SCML` learns an squared mahalanobis distance from triplet constraints by
```
bellet: Mahalanobis

doc/weakly_supervised.rst
```diff
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
...
+
+Sparse Compositional Metric Learning
+(:py:class:`SCML <metric_learn.SCML>`)
+
+`SCML` learns an squared mahalanobis distance from triplet constraints by 
+optimizing sparse positive weights assigned to a set of :math:`K` locally discriminative 
+rank-one PSD bases. This can be formulated as an optimization problem with only :math:`K`
+parameters, that can be solved with an efficient stochastic composite scheme.
+
+The Mahalanobis Matrix :math:`M` is built from a basis set :math:`B = \{b_i\}_{i=\{1,...,K\}}`
```
bellet: matrix

doc/weakly_supervised.rst
```diff
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
...
+.. _scml:
+
+:py:class:`SCML <metric_learn.SCML>`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Sparse Compositional Metric Learning
+(:py:class:`SCML <metric_learn.SCML>`)
+
+`SCML` learns an squared mahalanobis distance from triplet constraints by 
+optimizing sparse positive weights assigned to a set of :math:`K` locally discriminative
```
bellet: remove "locally discriminative", which refers to LDA bases (the basis set can be arbitrary)

doc/weakly_supervised.rst
```diff
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
...
+
+The Mahalanobis Matrix :math:`M` is built from a basis set :math:`B = \{b_i\}_{i=\{1,...,K\}}`
+weighted by a :math:`K` dimensional vector :math:`w = \{w_i\}_{i=\{1,...,K\}}` as:
+
+.. math::
+
+    M = \sum_{i=1}^K w_i b_i b_i^T = B \cdot diag(w) \cdot B^T \quad w_i \geq 0
+
+Learning :math:`M` in this form makes it PSD by design, as it is a nonnegative sum of PSD matrices.
+The optimization problem of :math:`w` over the triplets constraints :math:`C` is formulated as a
```
bellet: "over" usually refers to the variables being optimized, so I would reformulate as follows:
The optimization problem over :math:`w` is formulated as a classic margin-based hinge loss function involving :math:`C` triplets.

doc/weakly_supervised.rst
```diff
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
...
+classic margin-based hinge loss function over the relative constrains, a regularization :math:`\ell_1`
+is added to yield an sparse representation. The formulation is the following:
```
bellet: an --> a
sparse weights or sparse combination, instead of sparse representation?

doc/weakly_supervised.rst
```diff
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
...
+
+.. math::
+
+    \min_{w} \sum_{(x_a,x_b,x_c)\in C} [1 + d_w(x_a,x_b)-d_w(x_a,x_c)]_+ + \beta||w||_1
+
+Where :math:`[\cdot]_+` is the hinge loss.
```
bellet: where

doc/weakly_supervised.rst
```diff
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
...
+parameters, that can be solved with an efficient stochastic composite scheme.
+
+The Mahalanobis Matrix :math:`M` is built from a basis set :math:`B = \{b_i\}_{i=\{1,...,K\}}`
+weighted by a :math:`K` dimensional vector :math:`w = \{w_i\}_{i=\{1,...,K\}}` as:
+
+.. math::
+
+    M = \sum_{i=1}^K w_i b_i b_i^T = B \cdot diag(w) \cdot B^T \quad w_i \geq 0
+
+Learning :math:`M` in this form makes it PSD by design, as it is a nonnegative sum of PSD matrices.
```
bellet: Before "The optimization problem", add a sentence to clarify that the set of bases B is fixed in advance (typically generated from the input data).

doc/weakly_supervised.rst
```diff
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
...
+The optimization problem of :math:`w` over the triplets constraints :math:`C` is formulated as a
+classic margin-based hinge loss function over the relative constrains, a regularization :math:`\ell_1`
```
bellet: Make the regularization part a separate sentence?

doc/weakly_supervised.rst
```diff
@@ -700,6 +700,60 @@ of triplets that have the right predicted ordering.
...
+is added to yield an sparse representation. The formulation is the following:
+
+.. math::
+
+    \min_{w} \sum_{(x_a,x_b,x_c)\in C} [1 + d_w(x_a,x_b)-d_w(x_a,x_c)]_+ + \beta||w||_1
```
bellet: I would use i, j, k instead of a, b, c as common in the literature

Turns out we use i,j for pairs and a,b,c,d for quadruplets in LSML description. Maybe a good idea to use i,j,k,l there as well to make things more uniform across the doc

bellet: \min_{w\geq 0}

metric_learn/scml.py
grudloff: Not really! I think that may have stuck around from the previous `_generate_bases_dist_diff`.

doc/weakly_supervised.rst
bellet: (this applies also to the docstring somewhere I think)

metric_learn/scml.py
grudloff: Actually that also makes more sense!

metric_learn/scml.py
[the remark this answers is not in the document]
bellet: Ah, yes. Nevermind :-)
Have you checked that doing the extra work of identifying the unique pairs provides a speed up in practice?

test/test_sklearn_compat.py
```diff
@@ -359,20 +351,17 @@ def make_random_state(estimator, in_pipeline):
...
+    pipeline.fit(input_data, y)
 
     if hasattr(estimator, 'score'):
-      result = estimator.score(*remove_y_quadruplets(estimator,
-                                                     input_data,
-                                                     y))
-      result_pipe = pipeline.score(*remove_y_quadruplets(estimator,
-                                                         input_data,
-                                                         y))
+      result = estimator.score(input_data, y)
```
bellet: here too

metric_learn/scml.py
grudloff: Ok, great!

metric_learn/scml.py
grudloff: Great idea! This is a very good way to improve this!

metric_learn/scml.py
bellet: Right, I forgot for a second that the `diff` array was not the same distances, so nevermind :-)

perimosocordiae:
I'm seeing the "Unable to find commit in GitHub" entries as well, but the reports themselves are accessible for me. I'll see if there's some access control that needs tweaking.

perimosocordiae:
@bellet Are you logged in to codecov with your Github identity? If so, I don't see what could be blocking your access.

bellet:
> @bellet Are you logged in to codecov with your Github identity? If so, I don't see what could be blocking your access.

I wasn't. I can indeed access them now, thanks. I think I was able to see these without being logged in before though? Any reason why everything is not public?

bellet:
@grudloff for instance looks like you don't test the verbose mode

test/test_sklearn_compat.py
```diff
@@ -79,6 +80,9 @@ def test_sdml(self):
   def test_rca(self):
     check_estimator(Stable_RCA_Supervised)
 
+  def test_scml(self):
+    check_estimator(SCML_Supervised)
+
```
grudloff: I added this test but it is currently failing. The dataset used has very close values for the number of samples and of features, this leads to a large number of basis, which is currently allowed but yields multiple bases that are the same. I am not yet seeing why, but this yields the same values for the weights instead of just keeping one, which should be the behaviour under _l1_ regularization. If anyone has any insight on this it would be of great help.

bellet: Why is the test failing exactly?

doc/weakly_supervised.rst
grudloff: Good idea! I was aware I was missing mentioning that somewhere

doc/weakly_supervised.rst
grudloff: I actually was following the quadruplets convention! Yes, making all follow that convention sound good!

test/test_sklearn_compat.py
grudloff: On the Cholesky decomposition for the computation of `L`, it fails as the yielded matrix is not PSD. It caught my attention that all the bases were the same and the weights too. Very weird behaviour.

bellet: - *All* bases are the same, or some of them? If it is all of them it is very weird indeed, you should look into it to find out what happens
- Even if all bases are the same, as long as the weights are positive the final matrix should be PSD? Did you look into it more deeply to understand what happens? It is probably some numerical issue, e.g. some eigenvalue super close to zero but negative? This makes sense, and in fact in the case where `n_basis >= n_features` you should probably fall back on `components_from_metric` for a more robust behavior

test/metric_learn_test.py
```diff
@@ -76,6 +77,124 @@ def test_singular_returns_pseudo_inverse(self):
...
+    n_basis = 4.0
+
+    scml = SCML(n_basis=n_basis)
+    msg = ("n_basis should be an integer, instead it is of type %s"
+           % type(n_basis))
+    with pytest.raises(ValueError) as raised_error:
+      scml.fit(triplets)
+    assert msg == raised_error.value.args[0]
+
+  def test_bad_basis_supervised(self):
```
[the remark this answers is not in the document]
grudloff: Ok! Tried to parametrize it but without any success, it looks like it's not possible to parametrize inside a `unittest.TestCase` class! [https://stackoverflow.com/questions/35561843/py-test-parametrize-test-cases-from-classes](url)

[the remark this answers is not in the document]
bellet: True. This is why some ITML tests (e.g., `test_bounds_parameters_invalid`) are outside the class. You could do the same here.

IMO, ultimately we should remove this class structure (which we inherited from the initial use of `unittest` before we moved mostly to `pytest`) or make it a subclass of `object`as indicated in the stackoverflow topic you mentioned

test/metric_learn_test.py
```diff
@@ -76,6 +77,124 @@ def test_singular_returns_pseudo_inverse(self):
...
+    labels, class_count = np.unique(y, return_counts=True)
+    n_class = len(labels)
+
+    scml = SCML_Supervised(n_basis=n_class)
+    msg = ("The number of basis should be greater than the number of classes")
+    with pytest.raises(ValueError) as raised_error:
+      scml.fit(X, y)
+    assert msg == raised_error.value.args[0]
+
+  def test_big_n_basis_supervised(self):
```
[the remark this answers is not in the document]
grudloff: That actually would work ok but as LDA builds the scalings for the maximum separation of each class, having less than the number of classes would leave classes with "no representation" in the basis set. So the error is merely because it would be a really bad decision, maybe I could replace it for a warning.

test/metric_learn_test.py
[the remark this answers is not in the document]
grudloff: Making the SCML test inherit from `object` would be a pretty straightforward solution as there is only one assert from `unittest` and the rest are normal asserts. 

Maybe for the other classes, it may require a little bit more work. But I agree that it makes sense to move in that direction!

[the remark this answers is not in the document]
grudloff: Another alternative would be to make `MetricTestCase` inherit from object and build the required assertion methods that are used.

test/metric_learn_test.py
```diff
@@ -76,6 +77,124 @@ def test_singular_returns_pseudo_inverse(self):
...
+    n_basis = 4.0
+
+    scml = SCML_Supervised(n_basis=n_basis)
+    msg = ("n_basis should be an integer, instead it is of type %s"
+           % type(n_basis))
+    with pytest.raises(ValueError) as raised_error:
+      scml.fit(X, y)
+    assert msg == raised_error.value.args[0]
+
+  def test_array_basis_supervised(self):
```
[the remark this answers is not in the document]
grudloff: yes, of course! I don't know why I didn't.

test/test_base_metric.py
```diff
@@ -135,12 +135,12 @@ def test_get_metric_is_independent_from_metric_learner(estimator,
 
   # we fit the metric learner on it and then we compute the metric on some
   # points
-  model.fit(input_data, labels)
+  model.fit(*remove_y(model, input_data, labels))
```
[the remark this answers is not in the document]
grudloff: I don't know if it was a looked over or intentional, but `LSML` worked because it allows a second `weights` input. But in other tests, a `remove_y_quadruplet` was used.

test/test_sklearn_compat.py
```diff
@@ -359,20 +351,17 @@ def make_random_state(estimator, in_pipeline):
...
     estimator = clone(estimator)
-    estimator.set_params(preprocessor=preprocessor)
+    estimator.set_params(preprocessor=preprocessor,
+                         **make_random_state(estimator, False))
     pipeline = make_pipeline(estimator)
-    estimator.fit(*remove_y_quadruplets(estimator, input_data, y),
-                  **make_random_state(estimator, False))
-    pipeline.fit(*remove_y_quadruplets(estimator, input_data, y),
-                 **make_random_state(estimator, True))
+    estimator.fit(input_data, y)
```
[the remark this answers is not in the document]
grudloff: yeah, I was too when dealing with this at the beginning. It looks like the handling of LSML, which is a quadruplet learner, evolved as the tests for it were added. 

Here the quadruplets and triplets are excluded with the `if` on line 338 as they don't have a `y`. Anyways, I think this could be handled more elegantly by getting a list of the allowed estimators for this.

test/test_utils.py
```diff
@@ -911,35 +897,21 @@ def test_same_with_or_without_preprocessor(estimator, build_dataset):
                                    dataset_formed.data,
                                    random_state=SEED)
 
-  def make_random_state(estimator):
-    rs = {}
-    if estimator.__class__.__name__[-11:] == '_Supervised':
-      rs['random_state'] = check_random_state(SEED)
-    return rs
-
```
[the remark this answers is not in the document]
grudloff: This was used to set the `random_state` on `fit`, which is now deprecated. So now it is not necessary and actually was a problem since SCML doesn't have a `random_state` on `fit`.

test/test_sklearn_compat.py
grudloff: I tried to change the `n_basis` and I could achieve different escenarios with repeating basis, and with the same shared weigths on each set of repeated basis. I have to take a deeper look to see what is really going on. 

Yes, that is exactly what happens. If I compute the `np.linalg.eig` of `M` it yields some `e-17` negative values. That would definitely fix this really quickly!

test/metric_learn_test.py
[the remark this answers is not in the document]
bellet: Ok, makes sense. You need to make the message more informative (not clear what "clusters" are)

test/test_base_metric.py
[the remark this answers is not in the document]
bellet: Nice catch - it was probably not intentional but it is good to fix it!

test/test_sklearn_compat.py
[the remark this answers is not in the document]
bellet: OK, fair enough! It is actually explained in a comment on line 335: it would be nice to update it to mention also triplets

test/test_utils.py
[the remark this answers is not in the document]
bellet: Nice catch again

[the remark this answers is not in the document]
bellet: But maybe it is good to fix the random state for everyone at init then?

test/metric_learn_test.py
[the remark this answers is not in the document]
bellet: You should go for the easiest fix, and avoid making too many changes unrelated with this PR

But of course if you prefer you are very welcome to first remove the last remaining bits of `unittest` in a separate PR ;-)

metric_learn/scml.py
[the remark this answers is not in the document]
grudloff: Not really! I should benchmark this for sure!

doc/weakly_supervised.rst
grudloff: I changed the end a little bit to " involving the set :math:`C` of triplets."

test/metric_learn_test.py
```diff
@@ -76,6 +77,124 @@ def test_singular_returns_pseudo_inverse(self):
...
+
+  def test_big_n_basis(self):
+    scml = SCML(n_basis=4)
+    triplets = np.random.rand(3, 3, 3)
+    n_basis = 3
+    msg = ("The selected number of basis is greater than the number of points"
+           ", only n_basis = %d will be generated" % n_basis)
+    with pytest.warns(UserWarning) as raised_warning:
+      scml.fit(triplets)
+    assert msg == raised_warning[0].message.args[0]
```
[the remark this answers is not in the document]
grudloff: I replaced it for the test of the correct warning when there is dimensionality reduction!

test/test_utils.py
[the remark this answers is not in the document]
grudloff: I am not sure what you mean, the init of the estimator? The seed is already fixed by `set_random_state`!

test/test_sklearn_compat.py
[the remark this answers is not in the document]
grudloff: Yes, for sure! I thought I already did!

test/test_utils.py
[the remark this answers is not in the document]
bellet: Indeed, nevermind :-)

test/test_base_metric.py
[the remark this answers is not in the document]
wdevazelhes: I confirm, it was not intentional :p Good catch !

test/metric_learn_test.py
```diff
@@ -76,6 +76,111 @@ def test_singular_returns_pseudo_inverse(self):
                     pseudo_inverse)
 
 
+class TestSCML():
```
bellet: maybe better to inherit from `object` explicitly, see
https://stackoverflow.com/questions/15374857/should-all-python-classes-extend-object

test/metric_learn_test.py
```diff
@@ -76,6 +76,111 @@ def test_singular_returns_pseudo_inverse(self):
...
+  @pytest.mark.parametrize(('estimator', 'data', 'authorized_basis'),
+                           [(SCML, (np.ones((3, 3, 3)),), ['triplet_diffs']),
+                            (SCML_Supervised, (np.array([[0, 0], [0, 1],
+                                                         [2, 0], [2, 1]]),
+                                               np.array([1, 0, 1, 0])),
+                             ['triplet_diffs', 'lda'])])
+  def test_bad_basis(self, estimator, data, authorized_basis):
+    model = estimator(basis='bad_basis')
+    msg = ("`basis` must be one of the options '{}' or an array of shape "
+           "(n_basis, n_features).".format("', '".join(authorized_basis)))
```
bellet: you could maybe just reuse the attribute `_authorized_basis` from `SCML` and `SCML_Supervised`?

grudloff: Great idea!

metric_learn/scml.py
```diff
@@ -0,0 +1,624 @@
...
+      pos_eig_mask = w > 0
+      start = finish
+      finish += pos_eig_mask.sum()
+
+      try:
+        basis[start:finish, :] = v[pos_eig_mask]
+      except ValueError:
+        # if finish is greater than n_basis
+        basis[start:, :] = v[pos_eig_mask][:n_basis-start]
+        break
```
bellet: Related to previous comment: would be good that one of the test covers this case

metric_learn/scml.py
```diff
@@ -0,0 +1,624 @@
...
+      raise ValueError(
+          "`basis` must be one of the options '{}' "
+          "or an array of shape (n_basis, n_features)."
+          .format("', '".join(self._authorized_basis)))
+    if self.basis == 'triplet_diffs':
+      basis, n_basis = self._generate_bases_dist_diff(triplets, X)
+
+    return basis, n_basis
+
+  def _generate_bases_dist_diff(self, triplets, X):
```
bellet: It would be good to write a few tests to check that the number of bases generated is as expected in various cases

metric_learn/scml.py
```diff
@@ -0,0 +1,624 @@
...
+    """
+
+    if self.basis == 'lda':
+      basis, n_basis = self._generate_bases_LDA(X, y)
+    else:
+      basis, n_basis = None, None
+
+    return basis, n_basis
+
+  def _generate_bases_LDA(self, X, y):
```
bellet: Same here about testing

metric_learn/scml.py
```diff
@@ -0,0 +1,624 @@
...
+    # Number of clusters needed for 2 scales given the number of basis
+    # yielded by every LDA
+    n_clusters = int(np.ceil(n_basis/(2 * num_eig)))
+
+    kmeans = KMeans(n_clusters=n_clusters, random_state=self.random_state,
+                    algorithm='elkan').fit(X)
+    cX = kmeans.cluster_centers_
+
+    n_scales = 2
+    if n_features > 50:
```
bellet: again here would be good to have a test that covers this

test/test_sklearn_compat.py
grudloff: As the bases are exactly the same and the weights are all initialized on zero the effect of each iteration on the weights is shared.

I think this could be either solved by initializing the weights by random values or running an `np.unique` over the basis set. I think the first option would be better as the second would be very costly, but I don't know what the effect of changing the initialization would be.

metric_learn/scml.py
```diff
@@ -0,0 +1,624 @@
...
+
+# hack around lack of where in older numpy versions
+try:
+  np.sum([[0, 1], [1, 1]], where=[False, True], axis=1)
+except TypeError:
+  def sum_where(X, where):
+    return np.sum(X[where])
+else:
+  def sum_where(X, where):
+    return np.sum(X, where=where)
```
bellet: Do you think this is worth it? You only use it to compute the objective function, and I don't think the complexity is wildly different (according to quick timing experiment, `np.sum(X[where])` is slightly slower for dense `where` but slightly faster in the sparse case). Considering that the where option was just introduced in the last version of numpy, I think that simply using `np.sum(X[where])` is good enough for now.

grudloff: ok, sounds good!

test/test_sklearn_compat.py
bellet: When initializing with uniform weights it is indeed normal to get the same weight for bases that are the same. For *identical bases*, any allocation of a fixed amount of weight magnitude has the same l1 norm (w1=2 and w2=2, or w1=1 and w2=3, or w1=0 and w2=4, etc). The l1 norm eliminates linearly related features only when they are not exactly equal.

Of course it is very surprising that if you initialize to random weights you also get equal weights at the end !?

test/metric_learn_test.py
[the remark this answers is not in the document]
grudloff: Not sure what you meant there by clusters. But I just changed the Error section for the following:
```
    if n_basis < n_class:
      warnings.warn("The number of basis is less than the number of classes,"
                    " this will lead to less basis than the amount yielded by"
                    " LDA")
```
Let me know what you think about this alternative.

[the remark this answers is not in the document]
bellet: It is not easy for the user to understand the consequences I think ("less basis than the amount yielded by LDA" - so what?). Maybe "The number of basis is less than the number of classes, which may lead to poor discriminative performance."? Or something of the same fashion to be more concrete

Regarding clusters, I guess I got confused with the next test, which has this rather confusing message
```
msg = ("The selected number of basis needs a greater number of clusters"
           " than the number of available samples")
```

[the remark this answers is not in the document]
grudloff: How you phrase it is indeed way more clearer! Yes, I get that the other message is kind of confusing but I didn't know how to phrase it better without having to explain fully how the basis are generated. I will give it another try now!

test/metric_learn_test.py
```diff
@@ -86,6 +87,16 @@ def test_iris(self, basis):
     csep = class_separation(scml.transform(X), y)
     assert csep < 0.24
 
+  def test_big_n_features(self):
+    X, y = make_classification(n_samples=100, n_classes=3, n_features=60,
+                               n_informative=60, n_redundant=0, n_repeated=0,
+                               random_state=42)
```
grudloff: While making this test, I tried for `n_informative = 45` and `n_redundant = 15 `. But got as a result an empty `components` matrix, because the objective function never went below 1 which is the value it takes on `iter==1`, so the weights stayed with the initialization value. 

Watching at the change in the weights it seems like it is learning ok, just the values for the objective function are bigger (reaches a minimum of 2 or so) while in general values below 1 are achieved very fast.

I haven't looked deeply into it yet, but it seems like a situation that should be addressed.

test/metric_learn_test.py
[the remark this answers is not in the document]
bellet: I have looked at the new attempt. I think it is still very difficult to understand for the average user who does not even know what cluster refers too. Maybe you should just switch to something like
"Not enough samples to generate {} LDA bases: `n_basis` should be at most {}"
This way the user knows that there is a problem and how to fix it

test/metric_learn_test.py
bellet: Testing class separation is nice (although it is not good here probably due to dimension compared to number of data points), but I mostly meant making a test to check that the number of bases one obtains is as expected (check that the returned `n_basis` is equal to the expected one, and that the shape of the returned `basis` is also what we expect), parameterized for a few values of `n_samples`, `n_features` and `n_classes`

(of course you should keep the two tests with toy data where you also know exactly what the basis should be, which is nice)

grudloff: I actually thought this test only to test the `n_features > 50` case!

bellet: OK, then you could keep it and just extend it to also check the number of basis and basis shape, and parametrize it to cover a few values of the above mentioned parameters

grudloff: I was already on it! Just added it in the last commit.

bellet: Perfect!
For `n_samples`, maybe remove 1000 to avoid slowing down too much the test suite?

metric_learn/scml.py
```diff
@@ -65,9 +66,12 @@ def _fit(self, triplets, basis=None, n_basis=None):
     best_obj = np.inf
 
     rng = check_random_state(self.random_state)
-    rand_int = rng.randint(low=0, high=n_triplets, size=self.max_iter)
-    for iter in range(self.max_iter):
-      if (iter + 1) % self.output_iter == 0:
+    max_iter = int(self.max_iter/self.batch_size)
+    output_iter = int(self.output_iter/self.batch_size)
```
bellet: I think this is not consistent with the common semantic of "iteration": one iteration is one update of the parameter, regardless of the mini-batch size used to estimate the gradient

bellet:
Things are looking pretty good at this point. I think the priority now should be to have a benchmark which shows the benefit compared to existing algorithms in the package and allows to decide on decent default parameter values.

bellet:
Some additional small things to fix:
- The algorithm crashes when `output_iter` is larger than `max_iter` because `best_w` is not defined
- The algorithm crashes when some arguments are given with the wrong type, e.g. a float for `max_iter`. It would be a good idea to check that all arguments have the desired type and throw an error if it is not the case, along with tests

metric_learn/scml.py
```diff
@@ -0,0 +1,646 @@
...
+    if basis is None:
+      basis, n_basis = self._initialize_basis(triplets, X)
+
+    dist_diff = self._compute_dist_diff(triplets, X, basis)
+
+    n_triplets = triplets.shape[0]
+
+    # weight vector
+    w = np.zeros((1, n_basis))
+    # avarage obj gradient wrt weights
```
bellet, line 79: average

perimosocordiae:
Merged. Thanks for the huge contribution, @grudloff !

grudloff:
This is great news, thanks a lot! I had to leave this hanging but I definitely had in mind giving it the last push to reach completion once I had a bit of time. Glad you decided to merge as-is

bellet:
Congrats @grudloff!

wdevazelhes:
Congrats @grudloff !

PR #279, opened 2020-02-14
[MRG] Learning on Triplets

grudloff:
[WIP] Learning on Triplets

This PR is intended to allow the implementation of algorithms that learn on triplets by adding the base class for this kind of learner. The semantic of triplets is similar to quadruplets but the positive and negative pairs are constructed with respect to the first element of the triplet, for this the added code follows closely the one for quadruplets. 

The main reason for the creation of this is the implementation of SCML (PR  #278) an algorithm that learns on triplets. This will also enable the implementation of future algorithms that learn on triplets.

TODO:

- [x] add `_TripletsClassifierMixin` to `base_metric.py`
- [x] Tests
- [x] Documentation

### DIFF
--- a/doc/weakly_supervised.rst
+++ b/doc/weakly_supervised.rst
@@ -592,6 +592,119 @@ points, while constrains the sum of distances between dissimilar points:
         -with-side-information.pdf>`_. NIPS 2002
   .. [2] Adapted from Matlab code http://www.cs.cmu.edu/%7Eepxing/papers/Old_papers/code_Metric_online.tar.gz
 
+.. _learning_on_triplets:
+
+Learning on Triplets
+====================
+
+Some metric learning algorithms learn on triplets of samples. In this case,
+one should provide the algorithm with `n_samples` triplets of points. The
+semantic of each triplet is that the first two points should be closer
+together than the first and the last.
+
+Fitting
+-------
+Here is an example for fitting on triplets (see :ref:`fit_ws` for more
+details on the input data format and how to fit, in the general case of
+learning on tuples).
+
+>>> from metric_learn import SCML
+>>> triplets = np.array([[[1.2, 3.2], [2.3, 5.5], [2.1, 0.6]],
+>>>                      [[4.5, 2.3], [2.1, 2.3], [7.3, 3.4]]])
+>>> scml = SCML(random_state=42)
+>>> scml.fit(triplets)
+SCML(beta=1e-5, B=None, max_iter=100000, verbose=False,
+    preprocessor=None, random_state=None)
+
+Or alternatively (using a preprocessor):
+
+>>> X = np.array([[[1.2, 3.2], 
+>>>                [2.3, 5.5],
+>>>                [2.1, 0.6],
+>>>                [4.5, 2.3],
+>>>                [2.1, 2.3],
+>>>                [7.3, 3.4]])
+>>> triplets_indices = np.array([[0, 1, 2], [3, 4, 5]])
+>>> scml = SCML(preprocessor=X, random_state=42)
+>>> scml.fit(triplets_indices)
+SCML(beta=1e-5, B=None, max_iter=100000, verbose=False,
+   preprocessor=array([[1.2, 3.2],
+       [2.3, 5.5],
+       [2.4, 6.7],
+       [2.1, 0.6],
+       [4.5, 2.3],
+       [2.1, 2.3],
+       [0.6, 1.2],
+       [7.3, 3.4]]),
+    random_state=None)
+
+
+Here, we want to learn a metric that, for each of the two
+`triplets`, will put the two first points closer together than the first and the last.
+
+.. _triplets_predicting:
+
+Prediction
+----------
+
+When a triplets learner is fitted, it is also able to predict, for an
+upcoming triplet, whether the two first points are more similar than the
+first and the last (+1), or not (-1).
+
+>>> triplets_test = np.array(
+... [[[5.6, 5.3], [2.2, 2.1], [1.2, 3.4]],
+...  [[6.0, 4.2], [4.3, 1.2], [0.1, 7.8]]])
+>>> scml.predict(triplets_test)
+array([-1.,  1.])
+
+.. _triplets_scoring:
+
+Scoring
+-------
+
+Triplet metric learners can also
+return a `decision_function` for a set of pairs. This is basically the "score"
+which sign will be taken to find the prediction for the pair, which
+corresponds to the difference between the distance between the first two points,
+and the distance between the first and last points of the triplet (higher
+score means the first and last points are more likely to be more dissimilar than
+the two first points (i.e. more likely to have a +1 prediction since it's
+the right ordering)).
+
+>>> scml.decision_function(triplets_test)
+array([-1.75700306,  4.98982131])
+
+In the above example, for the first triplet in `triplets_test`, the
+two first points are predicted less similar than the first and last points (they
+are further away in the transformed space).
+
+Unlike for pairs learners, triplets learners don't allow to give a `y`
+when fitting, which does not allow to use scikit-learn scoring functions
+like:
+
+>>> from sklearn.model_selection import cross_val_score
+>>> cross_val_score(scml, triplets, scoring='f1_score')  # this won't work
+
+(This is actually intentional)
+
+However, triplets learners do have a default scoring function, which will
+basically return the accuracy score on a given test set, i.e. the proportion
+of triplets have the right predicted ordering.
+
+>>> scml.score(triplets_test)
+0.5
+
+.. note::
+   See :ref:`fit_ws` for more details on metric learners functions that are
+   not specific to learning on pairs, like `transform`, `score_pairs`,
+   `get_metric` and `get_mahalanobis_matrix`.
+
+
+
+
+Algorithms
+----------
+
 
 .. _learning_on_quadruplets:
 
--- a/doc/weakly_supervised.rst
+++ b/doc/weakly_supervised.rst
@@ -599,8 +599,8 @@ Learning on Triplets
 
 Some metric learning algorithms learn on triplets of samples. In this case,
 one should provide the algorithm with `n_samples` triplets of points. The
-semantic of each triplet is that the first two points should be closer
-together than the first and the last.
+semantic of each triplet is that the first point should be closer to the
+second point than to the third one.
 
 Fitting
 -------
@@ -640,7 +640,8 @@ SCML(beta=1e-5, B=None, max_iter=100000, verbose=False,
 
 
 Here, we want to learn a metric that, for each of the two
-`triplets`, will put the two first points closer together than the first and the last.
+`triplets`, will make the first point closer to the
+second point than to the third one.
 
 .. _triplets_predicting:
 
@@ -648,8 +649,8 @@ Prediction
 ----------
 
 When a triplets learner is fitted, it is also able to predict, for an
-upcoming triplet, whether the two first points are more similar than the
-first and the last (+1), or not (-1).
+upcoming triplet, whether the first point is closer to the second point 
+than to the third one. (+1), or not (-1).
 
 >>> triplets_test = np.array(
 ... [[[5.6, 5.3], [2.2, 2.1], [1.2, 3.4]],
@@ -662,34 +663,28 @@ array([-1.,  1.])
 Scoring
 -------
 
-Triplet metric learners can also
-return a `decision_function` for a set of pairs. This is basically the "score"
-which sign will be taken to find the prediction for the pair, which
-corresponds to the difference between the distance between the first two points,
-and the distance between the first and last points of the triplet (higher
-score means the first and last points are more likely to be more dissimilar than
-the two first points (i.e. more likely to have a +1 prediction since it's
-the right ordering)).
+Triplet metric learners can also return a `decision_function` for a set of triplets,
+which correspond to the distance between the first two points minus the distance
+between the first and last points of the triplet (the higher the value, the more
+similar the first point to the second point compared to the last one). This "score"
+can be interpreted as a measure of likeliness of having a +1 prediction for this 
+triplet.
 
 >>> scml.decision_function(triplets_test)
 array([-1.75700306,  4.98982131])
 
-In the above example, for the first triplet in `triplets_test`, the
-two first points are predicted less similar than the first and last points (they
-are further away in the transformed space).
+In the above example, for the first triplet in `triplets_test`, the first 
+point is predicted less similar to the second point than to the last point
+(they are further away in the transformed space).
 
-Unlike for pairs learners, triplets learners don't allow to give a `y`
-when fitting, which does not allow to use scikit-learn scoring functions
-like:
-
->>> from sklearn.model_selection import cross_val_score
->>> cross_val_score(scml, triplets, scoring='f1_score')  # this won't work
-
-(This is actually intentional)
+Unlike pairs learners, triplets learners do not allow to give a `y` when fitting: we
+assume that the ordering of points within triplets is such that the training triplets
+are all positive. Therefore, it is not possible to use scikit-learn scoring functions
+(such as 'f1_score') for triplets learners.
 
 However, triplets learners do have a default scoring function, which will
 basically return the accuracy score on a given test set, i.e. the proportion
-of triplets have the right predicted ordering.
+of triplets that have the right predicted ordering.
 
 >>> scml.score(triplets_test)
 0.5
@@ -777,16 +772,14 @@ array([-1.,  1.])
 .. _quadruplets_scoring:
 
 Scoring
--------
+-------W
 
-Quadruplet metric learners can also
-return a `decision_function` for a set of pairs. This is basically the "score"
-which sign will be taken to find the prediction for the pair, which
-corresponds to the difference between the distance between the two last points,
-and the distance between the two last points of the quadruplet (higher
-score means the two last points are more likely to be more dissimilar than
-the two first points (i.e. more likely to have a +1 prediction since it's
-the right ordering)).
+Quadruplet metric learners can also return a `decision_function` for a set of
+quadruplets, which correspond to the distance between the first pair of points minus 
+the distance between the second pair of points of the triplet (the higher the value,
+the more similar the first pair is than the last pair). 
+This "score" can be interpreted as a measure of likeliness of having a +1 prediction 
+for this quadruplet.
 
 >>> lsml.decision_function(quadruplets_test)
 array([-1.75700306,  4.98982131])
@@ -795,17 +788,10 @@ In the above example, for the first quadruplet in `quadruplets_test`, the
 two first points are predicted less similar than the two last points (they
 are further away in the transformed space).
 
-Unlike for pairs learners, quadruplets learners don't allow to give a `y`
-when fitting, which does not allow to use scikit-learn scoring functions
-like:
-
->>> from sklearn.model_selection import cross_val_score
->>> cross_val_score(lsml, quadruplets, scoring='f1_score')  # this won't work
-
-(This is actually intentional, for more details
-about that, see
-`this comment <https://github.com/scikit-learn-contrib/metric-learn/pull/168#pullrequestreview-203730742>`_
-on github.)
+Like triplet learners, quadruplets learners do not allow to give a `y` when fitting: we
+assume that the ordering of points within triplets is such that the training triplets
+are all positive. Therefore, it is not possible to use scikit-learn scoring functions
+(such as 'f1_score') for triplets learners.
 
 However, quadruplets learners do have a default scoring function, which will
 basically return the accuracy score on a given test set, i.e. the proportion
--- a/doc/weakly_supervised.rst
+++ b/doc/weakly_supervised.rst
@@ -594,7 +594,7 @@ points, while constrains the sum of distances between dissimilar points:
 
 .. _learning_on_triplets:
 
-Learning on Triplets
+Learning on triplets
 ====================
 
 Some metric learning algorithms learn on triplets of samples. In this case,
@@ -650,7 +650,7 @@ Prediction
 
 When a triplets learner is fitted, it is also able to predict, for an
 upcoming triplet, whether the first point is closer to the second point 
-than to the third one. (+1), or not (-1).
+than to the third one (+1), or not (-1).
 
 >>> triplets_test = np.array(
 ... [[[5.6, 5.3], [2.2, 2.1], [1.2, 3.4]],
@@ -664,7 +664,7 @@ Scoring
 -------
 
 Triplet metric learners can also return a `decision_function` for a set of triplets,
-which correspond to the distance between the first two points minus the distance
+which corresponds to the distance between the first two points minus the distance
 between the first and last points of the triplet (the higher the value, the more
 similar the first point to the second point compared to the last one). This "score"
 can be interpreted as a measure of likeliness of having a +1 prediction for this 
@@ -707,7 +707,7 @@ Learning on quadruplets
 =======================
 
 Some metric learning algorithms learn on quadruplets of samples. In this case,
-one should provide the algorithm with `n_samples` quadruplets of points. Th
+one should provide the algorithm with `n_samples` quadruplets of points. The
 semantic of each quadruplet is that the first two points should be closer
 together than the last two points.
 
@@ -772,10 +772,10 @@ array([-1.,  1.])
 .. _quadruplets_scoring:
 
 Scoring
--------W
+-------
 
 Quadruplet metric learners can also return a `decision_function` for a set of
-quadruplets, which correspond to the distance between the first pair of points minus 
+quadruplets, which corresponds to the distance between the first pair of points minus 
 the distance between the second pair of points of the triplet (the higher the value,
 the more similar the first pair is than the last pair). 
 This "score" can be interpreted as a measure of likeliness of having a +1 prediction 
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -589,6 +589,97 @@ class _PairsClassifierMixin(BaseMetricLearner):
                          'Got {} instead.'.format(type(beta)))
 
 
+class _TripletsClassifierMixin(BaseMetricLearner):
+  """Base class for triplets learners.
+  """
+
+  _tuple_size = 3  # number of points in a tuple, 3 for triplets
+
+  def predict(self, triplets):
+    """Predicts the ordering between sample distances in input triplets.
+
+    For each triplets, returns 1 if the first two elements are closer than the
+    first and last and -1 if not.
+
+    Parameters
+    ----------
+    triplets : array-like, shape=(n_triplets, 3, n_features) or (n_triplets, 3)
+      3D Array of triplets to predict, with each row corresponding to three
+      points, or 2D array of indices of triplets if the metric learner
+      uses a preprocessor.
+
+    Returns
+    -------
+    prediction : `numpy.ndarray` of floats, shape=(n_constraints,)
+      Predictions of the ordering of pairs, for each triplet.
+    """
+
+    # Aren't the following lines redundant as they are called on
+    # decision_function (as in quadruplets predict) ???
+    # check_is_fitted(self, 'preprocessor_')
+    # triplets = check_input(triplets, type_of_inputs='tuples',
+    #                       preprocessor=self.preprocessor_,
+    #                       estimator=self, tuple_size=self._tuple_size)
+    return np.sign(self.decision_function(triplets))
+
+  def decision_function(self, triplets):
+    """Predicts differences between sample distances in input triplets.
+
+    For each triplets in the samples, computes the difference between the
+    learned metric of the second pair minus the learned metric of the first
+    pair. The higher it is, the more probable it is that the pairs in the
+    triplets are presented in the right order, i.e. that the label of the
+    triplet is 1. The lower it is, the more probable it is that the label of
+    the triplet is -1.
+
+    Parameters
+    ----------
+    triplet : array-like, shape=(n_triplets, 4, n_features) or \
+                  (n_triplets, 4)
+      3D Array of triplets to predict, with each row corresponding to four
+      points, or 2D array of indices of triplets if the metric learner
+      uses a preprocessor.
+
+    Returns
+    -------
+    decision_function : `numpy.ndarray` of floats, shape=(n_constraints,)
+      Metric differences.
+    """
+    check_is_fitted(self, 'preprocessor_')
+    triplets = check_input(triplets, type_of_inputs='tuples',
+                           preprocessor=self.preprocessor_,
+                           estimator=self, tuple_size=self._tuple_size)
+    return (self.score_pairs(triplets[:, [0, 2]]) -
+            self.score_pairs(triplets[:, :2]))
+
+  def score(self, triplets):
+    """Computes score on input triplets
+
+    Returns the accuracy score of the following classification task: a record
+    is correctly classified if the predicted similarity between the first two
+    samples is higher than that of the first and last element.
+
+    Parameters
+    ----------
+    triplets : array-like, shape=(n_triplets, 4, n_features) or \
+                  (n_triplets, 4)
+      3D Array of triplets to score, with each row corresponding to four
+      points, or 2D array of indices of triplets if the metric learner
+      uses a preprocessor.
+
+    Returns
+    -------
+    score : float
+      The triplets score.
+    """
+    # Since the prediction is a vector of values in {-1, +1}, we need to
+    # rescale them to {0, 1} to compute the accuracy using the mean (because
+    # then 1 means a correctly classified result (pairs are in the right
+    # order), and a 0 an incorrectly classified result (pairs are in the
+    # wrong order).
+    return self.predict(triplets).mean() / 2 + 0.5
+
+
 class _QuadrupletsClassifierMixin(BaseMetricLearner):
   """Base class for quadruplets learners.
   """
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -598,8 +598,8 @@ class _TripletsClassifierMixin(BaseMetricLearner):
   def predict(self, triplets):
     """Predicts the ordering between sample distances in input triplets.
 
-    For each triplets, returns 1 if the first two elements are closer than the
-    first and last and -1 if not.
+    For each triplets, returns 1 if the first element is closer to the second than to the
+    last and -1 if not.
 
     Parameters
     ----------
--- a/metric_learn/base_metric.py
+++ b/metric_learn/base_metric.py
@@ -619,11 +619,11 @@ class _TripletsClassifierMixin(BaseMetricLearner):
     """Predicts differences between sample distances in input triplets.
 
     For each triplet (X_a, X_b, X_c) in the samples, computes the difference
-    between the learned metric of the second pair (X_a, X_c) minus the learned
-    metric of the first pair (X_a, X_b). The higher it is, the more probable it
-    is that the pairs in the triplets are presented in the right order, i.e.
-    that the label of the triplet is 1. The lower it is, the more probable it
-    is that the label of the triplet is -1.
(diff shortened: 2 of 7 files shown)


metric_learn/base_metric.py
```diff
@@ -589,6 +589,97 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+      uses a preprocessor.
+
+    Returns
+    -------
+    prediction : `numpy.ndarray` of floats, shape=(n_constraints,)
+      Predictions of the ordering of pairs, for each triplet.
+    """
+
+    # Aren't the following lines redundant as they are called on
+    # decision_function (as in quadruplets predict) ???
```
perimosocordiae: Yes, we should leave the check out here and in the quadruplets case as well.

grudloff: I think it is better to leave the check in `decision_function`, as it can be called separately.

bellet: Yes, this is indeed not needed, it looks like you can safely remove the calls to `check_is_fitted` and `check_input` here and in the quadruplet version

grudloff:
I think the PR is ready for review now.

doc/weakly_supervised.rst
```diff
@@ -592,6 +592,119 @@ points, while constrains the sum of distances between dissimilar points:
...
+are further away in the transformed space).
+
+Unlike for pairs learners, triplets learners don't allow to give a `y`
+when fitting, which does not allow to use scikit-learn scoring functions
+like:
+
+>>> from sklearn.model_selection import cross_val_score
+>>> cross_val_score(scml, triplets, scoring='f1_score')  # this won't work
+
+(This is actually intentional)
```
bellet: I think we can simplify this and remove the "example", simply writing:

> Unlike pairs learners, triplets learners do not allow to give a `y` when fitting: we assume that the ordering of points within triplets is such that the training triplets are all positive. Therefore, it is not possible to use scikit-learn scoring functions (such as 'f1_score') for triplets learners.

And modify the quadruplets part accordingly, replacing "Unlike pairs learners" by "Like triplets learners"

metric_learn/base_metric.py
```diff
@@ -589,6 +589,90 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+class _TripletsClassifierMixin(BaseMetricLearner):
+  """Base class for triplets learners.
+  """
+
+  _tuple_size = 3  # number of points in a tuple, 3 for triplets
+
+  def predict(self, triplets):
+    """Predicts the ordering between sample distances in input triplets.
+
+    For each triplets, returns 1 if the first two elements are closer than the
```

metric_learn/base_metric.py
```diff
@@ -589,6 +589,90 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+    first and last and -1 if not.
+
+    Parameters
+    ----------
+    triplets : array-like, shape=(n_triplets, 3, n_features) or (n_triplets, 3)
+      3D Array of triplets to predict, with each row corresponding to three
```
bellet: Array --> array

metric_learn/base_metric.py
```diff
@@ -589,6 +589,90 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+    """
+    return np.sign(self.decision_function(triplets))
+
+  def decision_function(self, triplets):
+    """Predicts differences between sample distances in input triplets.
+
+    For each triplets in the samples, computes the difference between the
+    learned metric of the second pair minus the learned metric of the first
+    pair. The higher it is, the more probable it is that the pairs in the
+    triplets are presented in the right order, i.e. that the label of the
```
bellet: make it more explicit what you call "the first pair" and "the second pair". It is not as obvious as in the quadruplet case

metric_learn/base_metric.py
```diff
@@ -589,6 +589,90 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+    triplet is 1. The lower it is, the more probable it is that the label of
+    the triplet is -1.
+
+    Parameters
+    ----------
+    triplet : array-like, shape=(n_triplets, 4, n_features) or \
+                  (n_triplets, 4)
+      3D Array of triplets to predict, with each row corresponding to four
```
bellet: Array --> array, four --> three

metric_learn/base_metric.py
```diff
@@ -589,6 +589,90 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+    learned metric of the second pair minus the learned metric of the first
+    pair. The higher it is, the more probable it is that the pairs in the
+    triplets are presented in the right order, i.e. that the label of the
+    triplet is 1. The lower it is, the more probable it is that the label of
+    the triplet is -1.
+
+    Parameters
+    ----------
+    triplet : array-like, shape=(n_triplets, 4, n_features) or \
+                  (n_triplets, 4)
```
bellet: 4 --> 3

metric_learn/base_metric.py
```diff
@@ -589,6 +589,90 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+                           estimator=self, tuple_size=self._tuple_size)
+    return (self.score_pairs(triplets[:, [0, 2]]) -
+            self.score_pairs(triplets[:, :2]))
+
+  def score(self, triplets):
+    """Computes score on input triplets
+
+    Returns the accuracy score of the following classification task: a record
+    is correctly classified if the predicted similarity between the first two
+    samples is higher than that of the first and last element.
```

metric_learn/base_metric.py
```diff
@@ -589,6 +589,90 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+    triplets = check_input(triplets, type_of_inputs='tuples',
+                           preprocessor=self.preprocessor_,
+                           estimator=self, tuple_size=self._tuple_size)
+    return (self.score_pairs(triplets[:, [0, 2]]) -
+            self.score_pairs(triplets[:, :2]))
+
+  def score(self, triplets):
+    """Computes score on input triplets
+
+    Returns the accuracy score of the following classification task: a record
```
bellet: record --> triplet

metric_learn/base_metric.py
```diff
@@ -589,6 +589,90 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+    is correctly classified if the predicted similarity between the first two
+    samples is higher than that of the first and last element.
+
+    Parameters
+    ----------
+    triplets : array-like, shape=(n_triplets, 4, n_features) or \
+                  (n_triplets, 4)
```
bellet: 4 --> 3

metric_learn/base_metric.py
```diff
@@ -589,6 +589,90 @@ def _validate_calibration_params(strategy='accuracy', min_rate=None,
...
+      3D Array of triplets to score, with each row corresponding to four
```
bellet: four --> three

metric_learn/constraints.py
```diff
@@ -33,6 +35,56 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
       return a[:n], b[:n], c[:n], d[:n]
     return a, b, c, d
 
+  def generate_knntriplets(self, X, k_genuine, k_impostor):
+
```
bellet: docstring would be nice

metric_learn/constraints.py
```diff
@@ -33,6 +35,56 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
       return a[:n], b[:n], c[:n], d[:n]
     return a, b, c, d
 
+  def generate_knntriplets(self, X, k_genuine, k_impostor):
```
bellet: In the future we might want to rethink a bit the Constraints class, because this now requires access to `X`, not just the labels given at init. But let's keep things like this for now

metric_learn/constraints.py
```diff
@@ -33,6 +35,56 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+
+    labels = np.unique(self.partial_labels)
+    L = len(labels)
```
bellet: avoid using cap symbol for something which is not a matrix. `n_labels` seems more appropriate

metric_learn/constraints.py
```diff
@@ -33,6 +35,56 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+
+    for i in range(L):
+
+        # generate mask for current label
+        gen_mask = self.partial_labels == labels[i]
+        gen_indx = np.where(gen_mask)
+
+        # get k_genuine genuine neighbours
+        neigh.fit(X=X[gen_indx])
+        gen_neigh = np.take(gen_indx, neigh.kneighbors(n_neighbors=k_genuine,
```
bellet: I think here you are including the point itself as its own neighbor

metric_learn/constraints.py
```diff
@@ -33,6 +35,56 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+        imp_indx = np.where(np.invert(gen_mask))
+
+        # get k_impostor impostor neighbours
+        neigh.fit(X=X[imp_indx])
+        imp_neigh = np.take(imp_indx, neigh.kneighbors(
+                            n_neighbors=k_impostor,
+                            X=X[gen_mask],
+                            return_distance=False))
+
+        # lenght = len_label*k_genuine*k_impostor
```
bellet: length

metric_learn/constraints.py
```diff
@@ -33,6 +35,56 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+        finish += np.sum(gen_mask)*k_genuine*k_impostor
+
+        triplets[start:finish, :] = self._comb(gen_indx, gen_neigh, imp_neigh,
+                                               k_genuine, k_impostor)
+        start = finish
+
+        # TODO: deal with too litle elements for k neighbors to be yielded
```
bellet: yes, definitely need this, and some tests

test/test_triplets_classifiers.py
```diff
@@ -0,0 +1,71 @@
...
+def test_accuracy_toy_example(estimator, build_dataset):
+  """Test that the default scoring for triplets (accuracy) works on some
+  toy example"""
+  triplets, X = build_dataset(with_preprocessor=True)
+  triplets = X[triplets]
+  estimator = clone(estimator)
+  set_random_state(estimator)
+  estimator.fit(triplets)
+  # We take the two first points and we build 4 regularly spaced points on the
+  # line they define, so that it's easy to build quadruplets of different
```
bellet: quadruplets --> triplets

test/test_utils.py
```diff
@@ -83,6 +85,34 @@ def build_pairs(with_preprocessor=False):
     return Dataset(X[c], target, None, X[c[:, 0]])
 
 
+def build_triplets(with_preprocessor=False):
+  input_data, labels = load_iris(return_X_y=True)
+  X, y = shuffle(input_data, labels, random_state=SEED)
+  constraints = Constraints(y)
+  triplets = constraints.generate_knntriplets(X, 3, 10)
```
bellet: maybe 3 or 4 impostors is sufficient? to make the tests faster

test/test_utils.py
```diff
@@ -83,6 +85,34 @@ def build_pairs(with_preprocessor=False):
...
+  if with_preprocessor:
+    # if preprocessor, we build a 2D array of triplets of indices
+    return triplets, X
+  else:
+    # if not, we build a 3D array of triplets of samples
+    return X[triplets], None
+
+
+class mock_triplet_LSML(_BaseLSML, _TripletsClassifierMixin):
```
bellet: this is fine for now. we can remove it in #278 and replace it with SCML

doc/weakly_supervised.rst
```diff
@@ -592,6 +592,119 @@ points, while constrains the sum of distances between dissimilar points:
...
+-------
+
+Triplet metric learners can also
+return a `decision_function` for a set of pairs. This is basically the "score"
+which sign will be taken to find the prediction for the pair, which
+corresponds to the difference between the distance between the first two points,
+and the distance between the first and last points of the triplet (higher
+score means the first and last points are more likely to be more dissimilar than
+the two first points (i.e. more likely to have a +1 prediction since it's
+the right ordering)).
```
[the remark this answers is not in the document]
grudloff: Yes, thanks! I wasn't quite happy with the original phrasing either.

doc/weakly_supervised.rst
grudloff: Great, I agree it's way more clear this way!

metric_learn/base_metric.py
grudloff: I added the referenced points following the notation used in the documentation of `<a,b,c>` subindices for elements of every triplet. Let me know what you think of the new phrasing.

metric_learn/constraints.py
grudloff: Done!

metric_learn/constraints.py
grudloff: Agreed. It would be nicer to be able to pass `X` to `Constrains()`, maybe as an option for cases like this when it's needed.

metric_learn/constraints.py
grudloff: Done

metric_learn/constraints.py
grudloff: No! The same point is not considered when `X` is not provided, as stated in `sklearn.neighbors.NearestNeighbors` documentation:
> (...) If not provided, neighbors of each indexed point are returned. In this case, the query point is not considered its own neighbor.

test/test_utils.py
grudloff: Yes, I think you are right! Done.

metric_learn/constraints.py
grudloff: How should this be handled? Should there be a k_vector = np.minimum(k, labels_count) like it is handled on the LDA basis construction? or should an error be raised for the user to select a lower number for k?

I am more inclined to the second option.

grudloff: I decided to implement this similarly as it is done with the LDA basis construction, by creating a k-vector that adjusts k to the classes that need a lower value. Also, an informative warning is thrown for every class with this issue in order to allow the user to know this is the underlying behaviour. Let me know if you think this a good way to go, or if there is a better alternative!

metric_learn/constraints.py
bellet: OK, my bad!

metric_learn/constraints.py
```diff
@@ -33,6 +35,104 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+    k_genuine_vec = np.ones(n_labels, dtype=np.intp)*k_genuine
+    k_impostor_vec = np.ones(n_labels, dtype=np.intp)*k_impostor
+
+    for i in range(n_labels):
+      if (k_genuine + 1 > labels_count[i]):
+        k_genuine_vec[i] = labels_count[i]-1
+        warnings.warn("The class {} has {} elements but a minimum of {},"
+                      " which corresponds to k_genuine+1, is expected. "
+                      "A lower number of k_genuine will be used for this"
+                      "class.\n"
```
bellet: This could be more direct and informative:
The class {} has {} elements, which is not sufficient to generate {} genuine neighbors as specified by k_genuine. Will generate {} genuine neighbors instead.

metric_learn/constraints.py
```diff
@@ -33,6 +35,104 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+                      .format(labels[i], labels_count[i], k_genuine+1))
+      if (k_impostor > len_input - labels_count[i]):
+        k_impostor_vec[i] = len_input - labels_count[i]
+        warnings.warn("The class {} has {} elements of other classes but a "
+                      "minimum of {}, which corresponds to k_impostor, is"
+                      " expected. A lower number of k_impostor will be used"
+                      " for this class.\n"
+                      .format(labels[i], len_input - labels_count[i],
+                              k_impostor))
```
bellet: same here

test/test_constraints.py
```diff
@@ -69,3 +70,78 @@ def test_unknown_labels_not_in_chunks(num_chunks, chunk_size):
...
+  k = 1
+  X = np.array([[0, 0],
+               [1, 1],
+               [2, 2],
+               [3, 3],
+               [4, 4],
+               [5, 5],
+               [6, 6],
+               [7, 7],
+               [8, 8]])
```
bellet: maybe you can more than one on each line to take up less space

test/test_constraints.py
```diff
@@ -69,3 +70,78 @@ def test_unknown_labels_not_in_chunks(num_chunks, chunk_size):
...
+  y = np.array([1, 1, 1, 2, 2, 2, 3, 3, 3])
+  T_test = np.array([[0, 1, 3],
+                    [1, 0, 3],
+                    [2, 1, 3],
+                    [3, 4, 2],
+                    [4, 3, 2],
+                    [5, 4, 6],
+                    [6, 7, 5],
+                    [7, 6, 5],
+                    [8, 7, 5]])
```
bellet: same here

test/test_constraints.py
```diff
@@ -69,3 +70,78 @@ def test_unknown_labels_not_in_chunks(num_chunks, chunk_size):
...
+  T = Constraints(y).generate_knntriplets(X, k, k)
+
+  assert np.array_equal(T, T_test)
```
bellet: this example is not ideal as there are ties (e.g., data point 1 is at equal distance from 0 and from 2), so this equality check assumes a specific tie breaking strategy. It would be better to modify the example so that it does not involve ties.

this check also assumes a particular ordering in the triplet generation. it is probably better to sort the rows so that the check is independent of the ordering

test/test_constraints.py
```diff
@@ -69,3 +70,78 @@ def test_unknown_labels_not_in_chunks(num_chunks, chunk_size):
                               random_state=SEED)
 
   assert np.all(chunks[labels < 0] < 0)
+
+
+def test_generate_knntriplets():
```
bellet: I think you should expand this test (manually or through `parametrize`) to check that everything goes well for several values of `k_genuine` and `k_impostor`, testing the normal case (the one you test already) but also the edge cases (when there is exactly the right amount of data points, and when one point is missing for either genuine or impostor neighbors).

doc/weakly_supervised.rst
```diff
@@ -592,6 +592,114 @@ points, while constrains the sum of distances between dissimilar points:
         -with-side-information.pdf>`_. NIPS 2002
   .. [2] Adapted from Matlab code http://www.cs.cmu.edu/%7Eepxing/papers/Old_papers/code_Metric_online.tar.gz
 
+.. _learning_on_triplets:
+
+Learning on Triplets
+====================
+
+Some metric learning algorithms learn on triplets of samples. In this case,
+one should provide the algorithm with `n_samples` triplets of points. The
```
bellet: nitpick: correct the typo "Th" --> "The" in the quadruplet version of this (line 710)

metric_learn/constraints.py
```diff
@@ -36,17 +36,39 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+        Input data, where each row corresponds to a single instance.
+      k_genuine : int
+        Number of neighbors of the same class to be taken into account.
+      k_impostor : int
+        Number of neighbors of different classes to be taken into account.
+
+    Returns
+    -------
+    triplets : array-like, shape=(n_constraints, 3)
+      2D array of triplets of indicators.
```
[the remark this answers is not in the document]
grudloff: They are synonyms and it's phrased like that in many parts of the repository!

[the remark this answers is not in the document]
bellet: To me, they are not synonyms and I find this a bit confusing as in ML indicator variables are typically binary variables that specify presence/absence of an attribute.
But if we use this terminology in many places, let's keep it like that

test/test_constraints.py
grudloff: Agreed! I added a test on those edge cases and the result is compared to the result of a "naive" version of the knn generator.

test/test_constraints.py
grudloff: Totally agree with you. Fixed these two issues in the las commit!

test/test_constraints.py
bellet: Why do you need to define a naive version? (which is then itself not tested...)
You could simply do everything in `test_generate_knntriplets` on the data you defined manually. Just like you did for the normal case, you can manually create the correct set of triplets for the edge cases and do the check

test/test_constraints.py
```diff
@@ -69,3 +71,149 @@ def test_unknown_labels_not_in_chunks(num_chunks, chunk_size):
...
+  k = 1
+  X = np.array([[0, 0], [2, 2], [4, 4], [8, 8], [16, 16], [32, 32], [64, 64],
+                [128, 128], [256, 256]])
+  y = np.array([1, 1, 1, 2, 2, 2, 3, 3, 3])
+
+  T_test = np.array([[0, 1, 3], [1, 0, 3], [2, 1, 3], [3, 4, 2], [4, 3, 2],
+                     [5, 4, 2], [6, 7, 5], [7, 6, 5], [8, 7, 5]])
+  T = Constraints(y).generate_knntriplets(X, k, k)
+
+  assert len(list(set(map(tuple, T)) - set(map(tuple, T_test)))) == 0
```
bellet: this seems more complicated than sorting the two arrays and checking for equality?

test/test_constraints.py
grudloff: Done!

test/test_constraints.py
grudloff: Agreed! Changed it to just sorting the triplets

test/test_constraints.py
```diff
@@ -69,3 +70,106 @@ def test_unknown_labels_not_in_chunks(num_chunks, chunk_size):
...
+                          (1, 3,
+                          [[0, 1, 3], [0, 1, 4], [0, 1, 5], [1, 0, 3],
+                           [1, 0, 4], [1, 0, 5], [2, 1, 3], [2, 1, 4],
+                           [2, 1, 5], [3, 4, 0], [3, 4, 1], [3, 4, 2],
+                           [4, 3, 0], [4, 3, 1], [4, 3, 2], [5, 4, 0],
+                           [5, 4, 1], [5, 4, 2]]),
+                          (1, 2,
+                          [[0, 1, 3], [0, 1, 4], [1, 0, 3], [1, 0, 4],
+                           [2, 1, 3], [2, 1, 4], [3, 4, 1], [3, 4, 2],
+                           [4, 3, 1], [4, 3, 2], [5, 4, 1], [5, 4, 2]])])
```
bellet: Could you also small test edges cases where there is not enough points? typically (1, 4) and (3, 1)?
the outputs should be the same as for (1, 3) and (2, 1) respectively

(you could remove (2,3), which is lengthy)

test/test_constraints.py
```diff
@@ -69,3 +70,106 @@ def test_unknown_labels_not_in_chunks(num_chunks, chunk_size):
...
+  k = 1
+  X = np.array([[0, 0], [2, 2], [4, 4], [8, 8], [16, 16], [32, 32], [64, 64],
+                [128, 128], [256, 256]])
+  y = np.array([1, 1, 1, 2, 2, 2, 3, 3, 3])
+
+  T_test = np.array([[0, 1, 3], [1, 0, 3], [2, 1, 3], [3, 4, 2], [4, 3, 2],
+                     [5, 4, 2], [6, 7, 5], [7, 6, 5], [8, 7, 5]])
+  T = Constraints(y).generate_knntriplets(X, k, k)
+
+  assert np.array_equal(sorted(T.tolist()), sorted(T_test.tolist()))
```
bellet: Not sure what is the added value of this compared to what you have now in `test_generate_knntriplets_k`?

test/test_constraints.py
bellet: Nevermind about (2,3), it is useful at it corresponds to the case where both values of k are maximum

test/test_constraints.py
grudloff: yes, I was thinking the same! I will just leave the other test.

test/test_constraints.py
grudloff: No problem! I will put (2,3) with over the edge cases (3, 3), (2, 4), (3, 4) which should all have the same result.

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+    """
+
+    labels, labels_count = np.unique(self.partial_labels, return_counts=True)
+    n_labels = len(labels)
+    len_input = np.size(self.partial_labels, 0)
```
perimosocordiae: `self.partial_labels.shape[0]` seems more idiomatic here

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+
+    # Handle the case where there are too few elements to yield k_genuine or
+    # k_impostor neighbors for every class.
+
+    k_genuine_vec = np.ones(n_labels, dtype=np.intp)*k_genuine
```
perimosocordiae: `np.full_like(labels, k_genuine)`

Same for the next line

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+    k_impostor_vec = np.ones(n_labels, dtype=np.intp)*k_impostor
+
+    for i in range(n_labels):
+      if (k_genuine + 1 > labels_count[i]):
```
perimosocordiae: Drop the extra parens here, same with the other if statements below.

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+    n_labels = len(labels)
+    len_input = np.size(self.partial_labels, 0)
+
+    # Handle the case where there are too few elements to yield k_genuine or
+    # k_impostor neighbors for every class.
+
+    k_genuine_vec = np.ones(n_labels, dtype=np.intp)*k_genuine
+    k_impostor_vec = np.ones(n_labels, dtype=np.intp)*k_impostor
+
+    for i in range(n_labels):
```
perimosocordiae: `for i, count in enumerate(labels_count):`

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+      if (k_impostor > len_input - labels_count[i]):
+        k_impostor_vec[i] = len_input - labels_count[i]
+        warnings.warn("The class {} has {} elements of other classes, which is"
+                      " not sufficient to generate {} impostor neighbors as "
+                      "specified by k_impostor. Will generate {} impostor "
+                      "neighbors instead.\n"
+                      .format(labels[i], k_impostor_vec[i], k_impostor,
+                              k_impostor_vec[i]))
+
+    triplets = np.empty((np.dot(k_genuine_vec*k_impostor_vec, labels_count),
```
perimosocordiae: This is confusing. Let's split it up:

```
num_triplets = labels_count.dot(k_genuine_vec * k_impostor_vec)
triplets = np.empty((num_triplets, 3), dtype=np.intp)
```

Also a comment explaining the math would be useful.

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+                         3), dtype=np.intp)
+
+    start = 0
+    finish = 0
+    neigh = NearestNeighbors()
+
+    for i in range(n_labels):
```
perimosocordiae: `for i, label in enumerate(labels):`

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+
+        # generate mask for current label
+        gen_mask = self.partial_labels == labels[i]
+        gen_indx = np.where(gen_mask)
```
perimosocordiae: Getting indices isn't necessary, you can index with the mask directly.

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+
+        # get k_genuine genuine neighbours
+        neigh.fit(X=X[gen_indx])
+        gen_neigh = np.take(gen_indx, neigh.kneighbors(
+                            n_neighbors=k_genuine_vec[i],
+                            return_distance=False))
+
+        # generate mask for impostors of current label
+        imp_indx = np.where(np.invert(gen_mask))
```
perimosocordiae: `imp_mask = ~gen_mask`

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+
+        # get k_impostor impostor neighbours
+        neigh.fit(X=X[imp_indx])
+        imp_neigh = np.take(imp_indx, neigh.kneighbors(
+                            n_neighbors=k_impostor_vec[i],
+                            X=X[gen_mask],
+                            return_distance=False))
+
+        # length = len_label*k_genuine*k_impostor
+        finish += labels_count[i]*k_genuine_vec[i]*k_impostor_vec[i]
```
perimosocordiae: Let's compute this outside the loop: `finish_indices = labels_count * k_genuine_vec * k_impostor_vec`

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+
+        triplets[start:finish, :] = self._comb(gen_indx, gen_neigh, imp_neigh,
+                                               k_genuine_vec[i],
+                                               k_impostor_vec[i])
+        start = finish
+
+    return triplets
+
+  def _comb(self, A, B, C, sizeB, sizeC):
```
perimosocordiae: This should be a free function, not a method.

metric_learn/constraints.py
```diff
@@ -33,6 +35,110 @@ def positive_negative_pairs(self, num_constraints, same_length=False,
...
+    # generate_knntriplets helper function
+    # generate an array will all combinations of choosing
```
perimosocordiae: will -> with

test/test_constraints.py
```diff
@@ -69,3 +70,104 @@ def test_unknown_labels_not_in_chunks(num_chunks, chunk_size):
...
+
+def test_generate_knntriplets_k_impostor():
+  """Checks the correct error raised when k_impostor is too big """
+  X, y = shuffle(*make_blobs(random_state=SEED),
+                 random_state=SEED)
+
+  length = len(y)
+  label, labels_count = np.unique(y, return_counts=True)
+  labels_count_max = np.max(labels_count)
+  idx_biggest_label = np.where(labels_count == labels_count_max)
```
perimosocordiae: `idx_biggest_label, = np.where(...)`

test/test_triplets_classifiers.py
```diff
@@ -0,0 +1,71 @@
...
+  input_data, preprocessor = build_dataset(with_preprocessor)
+  estimator = clone(estimator)
+  estimator.set_params(preprocessor=preprocessor)
+  set_random_state(estimator)
+  triplets_train, triplets_test = train_test_split(input_data)
+  estimator.fit(triplets_train)
+  predictions = estimator.predict(triplets_test)
+  for i in range(len(input_data)):
+    prediction = estimator.predict(input_data[None, i])
+    if(prediction == 0):
```
perimosocordiae: drop extra parens

test/test_utils.py
```diff
@@ -83,6 +85,34 @@ def build_pairs(with_preprocessor=False):
     return Dataset(X[c], target, None, X[c[:, 0]])
 
 
+def build_triplets(with_preprocessor=False):
+  input_data, labels = load_iris(return_X_y=True)
+  X, y = shuffle(input_data, labels, random_state=SEED)
+  constraints = Constraints(y)
+  triplets = constraints.generate_knntriplets(X, 3, 4)
```
perimosocordiae: `constraints.generate_knntriplets(X, k_genuine=3, k_impostor=4)`

metric_learn/constraints.py
grudloff: maybe `np.full(n_labels, k_genuine)` is better?

metric_learn/constraints.py
grudloff: I did it this way because it is useful later and it is faster to access X from the indices! Please let me know if I am wrong about this last bit.

metric_learn/constraints.py
grudloff: You mean to separate it in two lines for clarity or just replacing the `np.invert` by the `~`?

metric_learn/constraints.py
grudloff: I calculated directly the cumulative sum outside the for loop, take a look at it and let me know what you think.

metric_learn/constraints.py
perimosocordiae: I prefer `full_like` as it's more explicit that we're making it the same shape as the `labels` array. Ideally we wouldn't need to store `n_labels` anywhere.

You can also pass a `dtype=np.intp` to `full_like`, though I suspect the labels array is already of integer type.

metric_learn/constraints.py
perimosocordiae: I see now that you need the explicit indices for the `_comb` call, so I retract my comment.

test/test_triplets_classifiers.py
grudloff: actually, I forgot to take those lines out previously! I will erase them now.

metric_learn/constraints.py
perimosocordiae: I think you can avoid creating `imp_indx` entirely, as it's only used for indexing as far as I can see.

metric_learn/constraints.py
grudloff: Alright! You are correct `n_labels` isn't needed anymore. The `self.partial_labels` is indeed initialized as `int` in the `Constraints.__init__()` 

Should we handle unknown labels? (`label = -1`) I am thinking this could be easily handled by restricting `labels, labels_count` to the mask `labels!=-1`. This way the unknown labeled will be used for the impostor neighbors.

bellet: I think we should just ignore the unlabeled points (all points with negative label) as done in the `pairs` method. A point with unknown label may have the same label as the target point, so I don't think we should allow them as candidates for impostors.

grudloff: You are right! I will do that then.

bellet: Maybe you can add one unlabeled point to your manually defined dataset (one that is closer to some other point) so that the test verified that it is ignored

metric_learn/constraints.py
grudloff: It is used later to create `imp_neigh` on line 131:
```
imp_neigh = np.take(imp_indx, neigh.kneighbors(n_neighbors=k_impostor_vec[i], X=X[gen_mask], return_distance=False))
```
This maps the indeces of the impostors neighbors yielded by `niegh.kneighbors() ` that reference `X[imp_indx]` back to the indeces of X.

Maybe this is a bit unclear, should I refactor it to be more understandable?

perimosocordiae: Ooh, I misread that line at first. I do think it would be clearer if written as a plain indexing expression:

```
neighbor_inds = imp_indx[neigh.kneighbors(n_neighbors=k_impostor_vec[i], X=X[gen_mask], return_distance=False)]
```

grudloff: Sadly it cannot be indexed like that! as the `neigh.kneighbors()` yields a vector of indices of n_input times k. I changed it a little bit for clarity and added a comment, let me know what you think!

metric_learn/constraints.py
grudloff: Done!

bellet:
@perimosocordiae are you happy with the changes? If so I think we can merge

perimosocordiae:
LGTM, merged. This is a very nice improvement, thanks @grudloff !

grudloff:
Great! :) Thanks!

PR #283, opened 2020-03-30
[MRG] Allow installation from conda

bellet:
Anaconda is widely used so it would be nice to be able to install `metric-learn` with `conda install`. After a bit of research I think the best option is to add `metric-learn` to the popular `conda-forge` channel instead of creating a specific channel

I have created the recipe and PR there: https://github.com/conda-forge/staged-recipes/pull/11168

@perimosocordiae @terrytangyuan @wdevazelhes I put you in the list of maintainers, so you have to post a comment on the above PR to let them know that you are willing to be listed there

Note: I also noticed that the PyPi entry is somewhat outdated, for instance the badges for travis and codecov do not work probably due to outdated URL. metric-learn appears in the "My projects" section but the "manage" button is greyed out. Can someone give me the management rights and/or fix the small problems in the PyPi entry? Thanks!


perimosocordiae:
I commented on the conda PR. Thanks for setting this up!

terrytangyuan:
Thanks. This looks great!

bellet:
As requested in the conda-forge PR, I modified `setup.cfg` to include the license file to future source distributions. I checked locally and it works when running `python setup.py sdist`

synapticarbors:
@bellet If you checked that the entry in `setup.cfg` works, that's great. I had a hard time finding explicit documentation. In the past I've always suggested including a `MANIFEST.in` with the proper entry:

https://packaging.python.org/guides/using-manifest-in/#using-manifest-in

but if the `setup.cfg` route works, that's obviously fine too.

bellet:
> @bellet If you checked that the entry in `setup.cfg` works, that's great. I had a hard time finding explicit documentation. In the past I've always suggested including a `MANIFEST.in` with the proper entry:
> 
> https://packaging.python.org/guides/using-manifest-in/#using-manifest-in
> 
> but if the `setup.cfg` route works, that's obviously fine too.

Yes, I read about both solutions. I checked on my local machine and the added line in `setup.cfg` works (the tarball does contain the license file after the change). But I don't know how robust this behavior is. @perimosocordiae @terrytangyuan @wdevazelhes maybe you guys can give it a try on your local machine? Otherwise we can also add a `MANIFEST.in` file with the appropriate entry

terrytangyuan:
@bellet @synapticarbors I think adding it in `setup.cfg` has the same effect ([reference](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files)). It works for me locally.

README.rst
```diff
@@ -31,12 +31,12 @@ metric-learn contains efficient Python implementations of several popular superv
 
 **Installation/Setup**
 
-Run ``pip install metric-learn`` to download and install from PyPI.
+- If you use Anaconda: ``conda install -c conda-forge metric-learn``. See more
+options `here <https://github.com/conda-forge/metric-learn-feedstock#installing-metric-learn>`_.
 
-Run ``python setup.py install`` for default installation.
+- To install from PyPi: ``pip install metric-learn``.
```
terrytangyuan: PyPi -> PyPI

README.rst
```diff
@@ -33,6 +33,8 @@ metric-learn contains efficient Python implementations of several popular superv
 
 Run ``pip install metric-learn`` to download and install from PyPI.
 
+Run ``TO COMPLETE`` to download and install in conda.
```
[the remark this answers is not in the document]
bellet: Thanks, I have added the simple instructions to install and pointed to this page for more options

README.rst
bellet: thanks, fixed!

bellet:
Just tested the conda recipe for both Python 2 & 3 and it works well

perimosocordiae:
Thanks again, merged.

bellet:
> Note: I also noticed that the PyPi entry is somewhat outdated, for instance the badges for travis and codecov do not work probably due to outdated URL. metric-learn appears in the "My projects" section but the "manage" button is greyed out. Can someone give me the management rights and/or fix the small problems in the PyPi entry? Thanks!

@perimosocordiae @terrytangyuan @wdevazelhes can one of you check this? It looks like at least the CI badge urls need to be updated

terrytangyuan:
@bellet It seems like I cannot update the description manually. The issue is that [our last release's README](https://github.com/scikit-learn-contrib/metric-learn/blob/v0.5.0/README.rst) has the outdated URLs. The only solution would be release a new version. Also we need to make sure our README is correct and up-to-date before submitting release to PyPI next time.

bellet:
OK, I did not realize this was directly rendering the README file!

terrytangyuan:
@bellet Just added you as owner on PyPI in case I am away. Hope we'll all be safe around COVID-19. Take care!

bellet:
Thanks!

Issue #289, opened 2020-05-21
Doesn't work with scikit-learn>=0.23.0 as logsumexp was removed from sklearn.utils.fixes.


filaPro:
#### Description
Doesn't work with scikit-learn>=0.23.0 as logsumexp was removed from sklearn.utils.fixes.

#### Steps/Code to Reproduce

from metric_learn import NCA

#### Actual Results

cannot import name 'logsumexp' from 'sklearn.utils.fixes'

perimosocordiae:
Looks like we're using the scikit-learn backport of `logsumexp`, which they removed in favor of using the official scipy version: https://github.com/scikit-learn/scikit-learn/pull/16725

PR #290, opened 2020-05-22
Use scipy's logsumexp function

perimosocordiae:
Fixes gh-289.


perimosocordiae:
The test failures are real, but unrelated to this change. I'm a bit confused as to when they started happening.

perimosocordiae:
Eh, looks like there are a couple errors at head. I'll merge this now and move to the next ones.

Issue #272, opened 2020-01-13
Unconsistent behaviour of SDML when ussing skggm with fixed seed.


grudloff:
#### Description
I was using SDML_Supervised() for a subsequent 2D visualization with UMAP (Similar to t-sne) and got large differences in the results on every fit instance while using the same data. Fixing the seed doesn't make a difference. I tracked down the problem to the call of quic() done when skggm is installed, reviewing their code I found there is a fixed seed but anyway the results from that function vary in every call.

note: I am using the latest version from Skggm, will try to reproduce later with the version indicated in the documentation.
 
#### Steps/Code to Reproduce

```python
from metric_learn import SDML_Supervised
from sklearn.datasets import load_wine
from sklearn.preprocessing import StandardScaler
from sklearn.model_selection import train_test_split
import numpy as np

wine=load_wine()
X, y = load_wine(return_X_y=True)

X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=42)
scaler = StandardScaler()
X_train = scaler.fit_transform(X_train)
X_test = scaler.transform(X_test)

SDML=SDML_Supervised(random_state=42)
X_transform=SDML.fit_transform(X_train,y_train)
print(np.sum(np.abs(X_transform - SDML.fit_transform(X_train,y_train))))
```

#### Expected Results
The two instances of SDML fit should have the same result, then the printed difference should be zero.

#### Actual Results
Large numbers in the order of 100 to 300.

#### Versions

Linux-5.0.0-37-generic-x86_64-with-Ubuntu-18.04-bionic
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
NumPy 1.18.1
SciPy 1.4.1
Scikit-Learn 0.22.1
Metric-Learn 0.5.0
Skggm 0.2.8

bellet:
I just realized that we had noticed this when refurbishing SDML. See the following link and in particular the last post by jasonlaska:
https://github.com/skggm/skggm/pull/122

It seems they have not made a new release since then, which is why our doc suggests to install a particular developer version. Doing this should fix this problem. @grudloff can you try and confirm?

grudloff:
Installed the indicated version and the behavior turns out to be the expected. Maybe the following code could be added to the documentation for ease of installing that version:
`pip install "git+https://github.com/skggm/skggm.git@a0ed406586c4364ea3297a658f415e13b5cbdaf8"`

[...]

PR #291, opened 2020-05-22
Drop support for python 2 and python 3.5

perimosocordiae:
Fixes gh-166, and resolves some errors related to newer versions of scikit-learn.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,11 @@ env:
     - SKGGM_VERSION=a0ed406586c4364ea3297a658f415e13b5cbdaf8
 matrix:
   include:
-  - name: "Pytest python 3.5 without skggm"
-    python: "3.5"
+  - name: "Pytest python 3.6 without skggm"
+    python: "3.6"
     before_install:
       - sudo apt-get install liblapack-dev
-      - pip install --upgrade pip "pytest<5"
+      - pip install --upgrade pip pytest
       - pip install wheel cython numpy scipy codecov pytest-cov scikit-learn
     script:
       - pytest test --cov;
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,19 +7,8 @@ env:
     - SKGGM_VERSION=a0ed406586c4364ea3297a658f415e13b5cbdaf8
 matrix:
   include:
-  - name: "Pytest python 2.7 with skggm"
-    python: "2.7"
-    before_install:
-      - sudo apt-get install liblapack-dev
-      - pip install --upgrade pip pytest
-      - pip install wheel cython numpy scipy codecov pytest-cov scikit-learn
-      - pip install git+https://github.com/skggm/skggm.git@${SKGGM_VERSION};
-    script:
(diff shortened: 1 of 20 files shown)


perimosocordiae:
The Travis failures are due to an inconsistency in the way scikit-learn prints the `__repr__` of its transformers. In the version that supports py3.5, it uses the old behavior (print all params), but in newer versions (py3.6+) it only prints params that have non-default values.

Maybe this is a sign that we should drop py3.5 in this PR as well? @bellet any thoughts?

bellet:
sklearn stopped supporting Python 3.5, and this is also the case for our optional dependency skggm, see #166 
So I think I'm for dropping py35 as well.

If we go for this we should probably update the travis builds to include one without skggm, to make sure the code is tested in this setup

We should also update the skggm instructions as we should be able to simply use the install version, am I right @wdevazelhes ?

wdevazelhes:
I agree that it makes sense to drop py35

> We should also simplify the skggm-related instructions in the builds and the README/doc as we should be able to just use the last version instead of a specific commit, am I right @wdevazelhes ?

I think there was some undeterministicness problems still present in the current release of skggm (see https://github.com/scikit-learn-contrib/metric-learn/issues/272), which fix (https://github.com/skggm/skggm/commit/a0ed406586c4364ea3297a658f415e13b5cbdaf8) hasn't been released yet... However skggm doesn't seem very active... Also, @bellet since you fixed scikit-learn's issues with graphical lasso (https://github.com/scikit-learn/scikit-learn/pull/13276), maybe now the cases where skggm is actually useful are very rare? (or inexistant ? I don't remember exactly but I think the cases to look at are the non SPD ones: cf. https://github.com/scikit-learn-contrib/metric-learn/blob/2d5a942c0f48a093f2e1c0dca2426e632626a9a6/metric_learn/sdml.py#L116)

bellet:
> I agree that it makes sense to drop py35
> 
> > We should also simplify the skggm-related instructions in the builds and the README/doc as we should be able to just use the last version instead of a specific commit, am I right @wdevazelhes ?
> 
> I think there was some undeterministicness problems still present in the current release of skggm (see #272), which fix ([skggm/skggm@a0ed406](https://github.com/skggm/skggm/commit/a0ed406586c4364ea3297a658f415e13b5cbdaf8)) hasn't been released yet... However skggm doesn't seem very active... Also, @bellet since you fixed scikit-learn's issues with graphical lasso ([scikit-learn/scikit-learn#13276](https://github.com/scikit-learn/scikit-learn/pull/13276)), maybe now the cases where skggm is actually useful are very rare? (or inexistant ? I don't remember exactly but I think the cases to look at are the non SPD ones: cf.
> 
> https://github.com/scikit-learn-contrib/metric-learn/blob/2d5a942c0f48a093f2e1c0dca2426e632626a9a6/metric_learn/sdml.py#L116
> )

Nope, I don't think my fix changed much about this because the problem was only occurring in 2D. So let's keep things the same regarding `skggm` for now?

PR #299, opened 2020-07-02
[MRG] Add python requires plus message for old pips

wdevazelhes:
Adds the code from @bellet's link from Django that also deals with old pips, 
Also change the version number to 0.6.1 to be ready for the next release


wdevazelhes:
You're right ! Just added it

PR #308, opened 2021-03-01
Learn rate and length scale for MLKR

Ribosome25:
Hi Dear Maintainers,

I found that MLKR does not support custom learning rate and length scale yet. Can I add these features to it? Thanks.

### DIFF
--- a/metric_learn/mlkr.py
+++ b/metric_learn/mlkr.py
@@ -109,7 +109,7 @@ class MLKR(MahalanobisMixin, TransformerMixin):
   """
 
   def __init__(self, n_components=None, init='auto',
-               tol=None, max_iter=1000, verbose=False,
+               tol=None, max_iter=1000, sigma=1, learn_rate=1e-4, verbose=False,
                preprocessor=None, random_state=None):
     self.n_components = n_components
     self.init = init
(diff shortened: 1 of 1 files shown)


wdevazelhes:
Hi @Ribosome25, thanks for your message, 
- **Regarding the learning rate:** Our current implementation of MLKR uses lbfgs-b as an optimizer, where there is no learning rate (this is an advantage: it removes an additional hyperparameter, and lbfgs-b has a lot of advantages which why it is often used as a method of choice).  See for instance this answer: https://stackoverflow.com/questions/60902315/mlpregressor-learning-rate-init-for-lbfgs-solver-in-sklearn, on a similar question about the MLP in sklearn that uses L-BFGS.
- **Regarding the scaling:** In theory, if such scaling is needed for achieving a good performance, the learned mahalanobis matrix should learn it already (to say it otherwise, `pairwise_dist(x, x') /sigma = sqrt((x-x').T.dot(M).(x - x'))/sigma = sqrt((x-x').T.dot(M/sigma^2).(x - x'))`), therefore, is there any particular reason to add it ?

Ribosome25:
Thank you for the explanation. That is very helpful.

Thanks.

wdevazelhes:
@Ribosome25  I'm glad it was helpful, 

I'll close the PR, feel free to open another one it if you think of some other contribution ! 

Cheers,

PR #313, opened 2021-04-07
[MRG+2] Update repo to work with both new and old scikit-learn

### LINKED ISSUE #311: Update metric-learn to work with the most recent scikit-learn when doing pytest
There seems to have been some changes in scikit-learn's new version (0.24): I have the following errors ``ModuleNotFoundError: No module named 'sklearn.utils.testing'`` when calling pytest on ``metric-learn`` (after a fresh clone of master on a conda env with  scikit-learn 0.24). The reason is the one from this comment https://github.com/uber/causalml/issues/285#issuecomment-762435858: 
>The sklearn.utils.testing module is deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.utils. Anything that cannot be imported from sklearn.utils is now part of the private API.

wdevazelhes:
Fixes #311 
- I added the workaround suggested by @bellet here: https://github.com/scikit-learn-contrib/metric-learn/issues/311#issuecomment-804229956, so that imports work both in sklearn <0.24 and >= 0.24 (EDIT: actually the old import strategy was already deprecated in 0.22, so I compared the version to 0.22)
- as well as an additional travis job to test (I chose the one with python 3.6 since it's the oldest python so I thought if someting goes wrong it might be this one, and also with skggm, since I also I'm thinking something might have more chance to go wrong when using another package... I could have put all the checks (python3.6+3.7, with/without skggm), but then I thought the travis test suite might take quite some time)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ matrix:
     after_success:
       - bash <(curl -s https://codecov.io/bash)
   - name: "Pytest python 3.6 with skggm + scikit-learn 0.20.3"
+    # checks that tests work for the oldest supported scikit-learn version
     python: "3.6"
     before_install:
       - sudo apt-get install liblapack-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,18 @@ matrix:
       - pytest test --cov;
     after_success:
       - bash <(curl -s https://codecov.io/bash)
+  - name: "Pytest python 3.6 with skggm + scikit-learn 0.20.3"
+    python: "3.6"
+    before_install:
+      - sudo apt-get install liblapack-dev
+      - pip install --upgrade pip pytest
+      - pip install wheel cython numpy scipy codecov pytest-cov
+      - pip install scikit-learn==0.20.3
+      - pip install git+https://github.com/skggm/skggm.git@${SKGGM_VERSION};
+    script:
+      - pytest test --cov;
+    after_success:
+      - bash <(curl -s https://codecov.io/bash)
   - name: "Syntax checking with flake8"
     python: "3.7"
     before_install:
--- a/metric_learn/sklearn_shims.py
+++ b/metric_learn/sklearn_shims.py
@@ -1,0 +1,27 @@
+"""This file is for fixing imports due to different APIs
+depending on the scikit-learn version"""
+import sklearn
+from packaging import version
+SKLEARN_AT_LEAST_0_22 = (version.parse(sklearn.__version__)
+                         >= version.parse('0.22.0'))
+if SKLEARN_AT_LEAST_0_22:
+    from sklearn.utils._testing import (set_random_state,
+                                        assert_warns_message,
+                                        ignore_warnings,
+                                        assert_allclose_dense_sparse,
+                                        _get_args)
+    from sklearn.utils.estimator_checks import (_is_public_parameter
+                                                as is_public_parameter)
+    from sklearn.metrics._scorer import get_scorer
+else:
+    from sklearn.utils.testing import (set_random_state,
+                                       assert_warns_message,
+                                       ignore_warnings,
+                                       assert_allclose_dense_sparse,
+                                       _get_args)
+    from sklearn.utils.estimator_checks import is_public_parameter
+    from sklearn.metrics.scorer import get_scorer
+
+__all__ = ['set_random_state', 'assert_warns_message', 'set_random_state',
+           'ignore_warnings', 'assert_allclose_dense_sparse', '_get_args',
+           'is_public_parameter', 'get_scorer']
--- a/test/test_base_metric.py
+++ b/test/test_base_metric.py
@@ -6,7 +6,8 @@ import numpy as np
 from sklearn import clone
 import sklearn
 from packaging import version
-if version.parse(sklearn.__version__) >= version.parse('0.22.0'):
+skversion = version.parse(sklearn.__version__)
+if skversion >= version.parse('0.22.0'):
     from sklearn.utils._testing import set_random_state
 else:
     from sklearn.utils.testing import set_random_state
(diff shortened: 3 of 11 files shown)


wdevazelhes:
Actually, tests are failing because of some LMNN problem similar to the discussion in #309 : a test is run on LMNN with classes that have too few labels.
The fact that this pb happens in only some travis tests could be due to some different random states depending on the order/quantity of runned tests depending on the particular travis test (with/without skggm) (+ I guess the change of library versions wrt to the past could maybe impact the seed? otherwise I don't know why tests were passing before) so I set the random seed in the failing test so that they should all either fail or pass
After having a consistent behaviour we can decide what to do (either adapt the test to have big enough classes, or adapt LMNN to accept classes with few labels)
I'll react on #309 accordingly since the decision should also impact the new LMNN PR (for instance we can decide to merge it as is since actually the current LMNN is even more strict than the newer PR (it needs at least `n_neighbors + 1` elements in each class, not just 2), so the new PR would actually allow to deal with more cases, and we could deal with singleton classes in another PR)

wdevazelhes:
Ok so there's just the travis test with the old scikit learn that is failing, because of the string representation of estimators, I'll look into that
But at least fixing the random seed indeed fixed the other tests (temporarily since we should not depend on the random seed)

test/test_base_metric.py
```diff
@@ -4,71 +4,170 @@
...
+    from sklearn.utils._testing import set_random_state
+else:
+    from sklearn.utils.testing import set_random_state
 
 
 def remove_spaces(s):
   return re.sub(r'\s+', '', s)
 
 
+def sk_repr_kwargs(def_kwargs, nndef_kwargs):
```
wdevazelhes, line 21: Here I thought it could be good to test the str repr for both the old (<=0.22) sklearn version and the newer ones, so I made a string representation that depends on the sklearn version

test/test_sklearn_compat.py
```diff
@@ -121,7 +131,8 @@ def test_array_like_inputs(estimator, build_dataset, with_preprocessor):
 
   # we subsample the data for the test to be more efficient
   input_data, _, labels, _ = train_test_split(input_data, labels,
-                                              train_size=20)
```
wdevazelhes, line 124: Test was failing here because of classes with too few labels in LMNN (see comments above), so in the end I created a toy example with a bit more samples (which I guess makes sense because the role of this particular test is not to test edge cases, but rather the fact that array-like objects work with our estimators),

wdevazelhes:
I guess the PR is ready to merge :)

test/test_sklearn_compat.py
```diff
@@ -52,37 +62,37 @@ def __init__(self, sparsity_param=0.01,
...
-    check_estimator(Stable_SDML_Supervised)
+    check_estimator(Stable_SDML_Supervised())
 
   def test_rca(self):
-    check_estimator(Stable_RCA_Supervised)
+    check_estimator(Stable_RCA_Supervised())
 
   def test_scml(self):
-    check_estimator(SCML_Supervised)
+    check_estimator(SCML_Supervised())
```
wdevazelhes, line 95: Here scikit-learn had return an error saying that checks should be run on estimators instances, not classes

.travis.yml
```diff
@@ -39,6 +39,18 @@ matrix:
       - pytest test --cov;
     after_success:
       - bash <(curl -s https://codecov.io/bash)
+  - name: "Pytest python 3.6 with skggm + scikit-learn 0.20.3"
```
terrytangyuan, line 42: Perhaps add a note here to clarify this additional test's purpose

terrytangyuan (approved):
Just a small comment. Otherwise LGTM

.travis.yml
wdevazelhes: I agree, done

wdevazelhes:
Thanks for the review @terrytangyuan! I adressed your comment

test/test_base_metric.py
```diff
@@ -4,71 +4,170 @@
...
+    def_kwargs.update(nndef_kwargs)
+    args_str = ",".join(f"{key}={strify(value)}"
+                        for key, value in def_kwargs.items())
+    return args_str
+
+
+def strify(obj):
+    """Function to add the quotation marks if the object
+    is a string"""
+    if type(obj) is str:
```
perimosocordiae, line 40: if isinstance(obj, str)

perimosocordiae: Depending on the inputs we expect to pass, it might be simpler to call `json.dumps(obj)` instead.

test/test_base_metric.py
```diff
@@ -4,71 +4,170 @@
...
 
+def sk_repr_kwargs(def_kwargs, nndef_kwargs):
+    """Given the non-default arguments, and the default
+    keywords arguments, build the string that will appear
+    in the __repr__ of the estimator, depending on the
+    version of scikit-learn.
+    """
+    if skversion >= version.parse('0.22.0'):
+        def_kwargs = ""
+    nndef_kwargs = eval(f"dict({nndef_kwargs})")
```
perimosocordiae, line 29: Rather than using eval, could we pass actual dicts here instead? The values are all simple literals, so it should be easier that way.

test/test_mahalanobis_mixin.py
```diff
@@ -11,7 +11,12 @@
 from sklearn.datasets import make_spd_matrix, make_blobs
 from sklearn.utils import check_random_state, shuffle
 from sklearn.utils.multiclass import type_of_target
-from sklearn.utils.testing import set_random_state
+import sklearn
+from packaging import version
+if version.parse(sklearn.__version__) >= version.parse('0.22.0'):
```
perimosocordiae, line 16: This seems like something we could do once during module initialization, then have a global constant like `SKLEARN_AT_LEAST_0_22` that we could use.

Or make a `sklearn_shims.py` file that does the conditional import, and have everything else import from that.

perimosocordiae (changes requested):

test/test_base_metric.py
wdevazelhes: Yes you're right, will do

test/test_base_metric.py
wdevazelhes: That's right, thanks ! Will do

test/test_mahalanobis_mixin.py
wdevazelhes: I agree, will do

wdevazelhes:
Thanks for the review @perimosocordiae ! I'll do the changes and put the PR to MRG as soon as I finish

test/test_base_metric.py
wdevazelhes: actually there was a pb for booleans (``json.dumps`` casts them to lowercase, i.e. ``json.dumps(True)`` returns ``true``), but I found online that ``repr`` can do the same without this pb

test/test_mahalanobis_mixin.py
wdevazelhes: done

test/test_base_metric.py
wdevazelhes: done

test/test_base_metric.py
wdevazelhes: done

metric_learn/sklearn_shims.py
```diff
@@ -0,0 +1,27 @@
...
+else:
+    from sklearn.utils.testing import (set_random_state,
+                                       assert_warns_message,
+                                       ignore_warnings,
+                                       assert_allclose_dense_sparse,
+                                       _get_args)
+    from sklearn.utils.estimator_checks import is_public_parameter
+    from sklearn.metrics.scorer import get_scorer
+
+__all__ = ['set_random_state', 'assert_warns_message', 'set_random_state',
```
wdevazelhes, line 25: writing this ``__all__`` lines were necessary for flake8 not to return errors

wdevazelhes:
@perimosocordiae I adressed your comments, I guess the PR is now ready to merge :)

bellet (approved):
+1 for merging, thanks @wdevazelhes!

bellet:
Somehow we do not see all checks (it happens sometimes), maybe do an empty commit to make sure @wdevazelhes?

wdevazelhes:
Thanks for the review @bellet ! 

I tried to do an empty commit but we still see 1 line for the check, indeed it's weird...  Although clicking on details we see all checks, so maybe it's just a change of display ?

wdevazelhes:
@perimosocordiae, do you approve my updates after your review ?

bellet:
> Thanks for the review @bellet !
> 
> I tried to do an empty commit but we still see 1 line for the check, indeed it's weird... Although clicking on details we see all checks, so maybe there was a change of display of travis checks on github ?

We don't see test coverage though for some reason

wdevazelhes:
> We don't see test coverage though for some reason

Really ? If I click on "show all checks" and then "details", here is what I see: 



image

perimosocordiae (approved):

perimosocordiae:
Merged, thanks!

bellet:
> > We don't see test coverage though for some reason
> 
> Really ? If I click on "show all checks" and then "details", here is what I see:
> 
> image

Where do you see test coverage (codecov)?

wdevazelhes:
> Where do you see test coverage (codecov)?

Ah yes that's right sorry ! Actually they sent me an email telling me that their bash uploader was hacked: https://about.codecov.io/security-update/?utm_medium=email&utm_source=marketo&ajs_uid=2386266&ajs_event=Clicked%20Email&mkt_tok=MzMyLUxWWC03NDEAAAF8dJ5mOA5H1E5ORbGsBiNKeiALhIlihm9VPW70i0qLTSuToRz_FC1YHHXfXcvCiatj0yHQxAfkDKCri3j9ksaM0_8pH36YziLCqWSOcdIt
So I guess the pb is probably because of that, I'll try to look at it see if they say any guideline for fixing the pb

wdevazelhes:
> Merged, thanks!

Thanks @perimosocordiae !
I'll try to look into the codecov pb noticed by @bellet (see above), to double check that this PR indeed checks the coverage requirement

wdevazelhes:
It seems the codecov is OK here is what I see on codecov in the "pull" section: 

image

bellet:
> > Where do you see test coverage (codecov)?
> 
> Ah yes that's right sorry ! Actually they sent me an email telling me that their bash uploader was hacked: https://about.codecov.io/security-update/?utm_medium=email&utm_source=marketo&ajs_uid=2386266&ajs_event=Clicked%20Email&mkt_tok=MzMyLUxWWC03NDEAAAF8dJ5mOA5H1E5ORbGsBiNKeiALhIlihm9VPW70i0qLTSuToRz_FC1YHHXfXcvCiatj0yHQxAfkDKCri3j9ksaM0_8pH36YziLCqWSOcdIt
> So I guess the pb is probably because of that, I'll try to look at it see if they say any guideline for fixing the pb

Good catch! They have guidelines on what to do on the above page

wdevazelhes:
I opened a draft PR here #314, running their recommended ``env`` command, and it doesn't seem to leak any sensitive information (what ``env`` prints can be seen for instance here ), (I was wondering about the ``SSH_CONNECTION`` variable that is printed but I checked here and it contains no password just IP adresses and port numbers)

As a precaution, on app.codecov.io, I clicked on "Regenerate" for the "Repository Upload Token", but I don't even think it had any effect, since I searched for "codecov" in our code, and I see nowhere any token

So I think we are safe,

However codecov still doesn't appear in the "all checks" of the test PR #314, so I think we can wait a little bit: they may be still fixing problems
In the meantime we can still check manually the code coverage for next PRs directly on the codecov website.

PR #322, opened 2021-09-14
Threshold must be a real number

### LINKED ISSUE #219: forbid any non-float value for the threshold
right now the threshold can be set by any value. We should forbid to set it by sth else than a float

mvargas33:
Closes #219

As stated in the [issue #219 ](https://github.com/scikit-learn-contrib/metric-learn/issues/219), the threshold variable for `_PairsClassifierMixin` must be a real number (float or integer).

A type check has been added in `set_threshold()` method, following the same format/guideline of type checking in other parts of the code.

The threshold set from `calibrate_threshold()` method is guaranteed to be float, so no additional type checking is needed.


terrytangyuan (approved):
LGTM. Others PTAL

bellet (changes requested):
Thanks @mvargas33!
It would be nice to add a couple of tests to ensure we have the expected behavior

@terrytangyuan @perimosocordiae @wdevazelhes: I can't find the automatic CI / coverage reports - is it me or something is broken there?

terrytangyuan:
From https://travis-ci.org/github/scikit-learn-contrib/metric-learn/builds, it says  `Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.`

We should probably switch to use GitHub Actions now. Many projects have already migrated.

bellet:
> From https://travis-ci.org/github/scikit-learn-contrib/metric-learn/builds, it says `Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.` (Travis is no longer free)
> 
> We should probably switch to use GitHub Actions now. Many projects have already migrated.

Indeed it seems CI / codecov checks are no longer working (last ones are 4 months ago). Switching to GitHub Actions sounds good - do you happen to know how to go about it? I've never used it.

terrytangyuan:
It's pretty straightforward to set up. Just create a `.yml` file in `.github/workflows` with the testing steps. A template can be found [here](https://github.com/actions/starter-workflows/blob/028df69d88fa6b986e3ec1f52b4ae52300e87c5a/ci/python-package-conda.yml).

terrytangyuan:
Created https://github.com/scikit-learn-contrib/metric-learn/issues/325 to track that separately. Feel free to pick it up.

metric_learn/base_metric.py
```diff
@@ -410,6 +410,10 @@ def set_threshold(self, threshold):
     """
     check_is_fitted(self, 'preprocessor_')
 
+    if threshold is None or not isinstance(threshold, (int, float)):
```
perimosocordiae, line 413: No need to check `is None` here, the isinstance check will handle None values as expected.

It may be even simpler to remove this checking and call `self.threshold_ = float(threshold)` instead. This will raise a ValueError if given invalid input, and also handles a broader range of float-like types.

perimosocordiae (changes requested):

metric_learn/base_metric.py
mvargas33: I agree with this simpler solution

bellet: this also removes the need for a test I guess

bellet: I agree as well. Perhaps writing a couple of test to make sure this raises the right error is still a good idea

mvargas33: Ok! Will do

mvargas33:
I made a test to check correct behavior. The `float(threshold)` works for almost all datatypes, but not for boolean as `float(True)=1.0`, so I had to add `isinstance()` just to check the boolean type.

metric_learn/base_metric.py
```diff
@@ -409,8 +409,10 @@ def set_threshold(self, threshold):
       The pairs classifier with the new threshold set.
     """
     check_is_fitted(self, 'preprocessor_')
-
-    self.threshold_ = threshold
+    if threshold is isinstance(threshold, (bool)):
+        raise ValueError('Parameter threshold must be a real number. '
+                         'Got {} instead.'.format(type(threshold)))
+    self.threshold_ = float(threshold)
```
bellet, line 415: Two comments:
- Not sure if it makes sense to make an exception for `bool`
- Maybe it is good to throw the custom error in all cases, which may be more informative for users? what do you think @perimosocordiae?

metric_learn/base_metric.py
```diff
@@ -409,8 +409,10 @@ def set_threshold(self, threshold):
       The pairs classifier with the new threshold set.
     """
     check_is_fitted(self, 'preprocessor_')
-
-    self.threshold_ = threshold
+    if threshold is isinstance(threshold, (bool)):
+        raise ValueError('Parameter threshold must be a real number. '
```
bellet, line 413: use 2-space indent to be consistent

metric_learn/base_metric.py
mvargas33: I agree, and the first approach was that, then I changed it to `float(threshold)`. But given that it works for al except `bool`, maybe its better to use the custom error for all cases as proposed initially.

metric_learn/base_metric.py
```diff
@@ -569,8 +569,11 @@ def set_threshold(self, threshold):
       The pairs classifier with the new threshold set.
     """
     check_is_fitted(self, 'preprocessor_')
-
-    self.threshold_ = threshold
+    try:
+      self.threshold_ = float(threshold)
+    except Exception:
```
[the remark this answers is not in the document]
mvargas33: I had to catch `ValueError` and `TypeError`. Most cases fall into `TypeError`, but `String` fall into `ValueError` with the funtion `float()`:
```
ValueError("could not convert string to float: 'ABC'")
```
Considering both types of exceptions, the test passes.

perimosocordiae (approved):

perimosocordiae:
Merged, thanks!

PR #323, opened 2021-09-14
[Minor] GUI for matplotlib to run examples

mvargas33:
If you try to run the examples on a fresh virtual environment, no figure will be shown and you will end up with the following error:

`UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.`

I propose to add `pyqt5` to 'demo' dependencies. This is a package with a GUI that works for matplotlib. This solves the issue.


terrytangyuan (approved):
LGTM

perimosocordiae:
I don't think we need to specify QT as a dependency here, there are several matplotlib backends that should all work reasonably well and I wouldn't want to make users download the QT backend if they aren't going to use it.

Maybe the best change here is a comment in the demo noting that a graphical matplotlib backend is recommended.

mvargas33:
I followed @perimosocordiae recommendation and added a note in the docs for `sandwich` and `algorithms walkthrough` instead. That should be enough.

perimosocordiae (approved):

PR #324, opened 2021-09-14
Rename variables, proposed by issue #257

### LINKED ISSUE #257: Improve consistency in parameter names
It would be nice to follow a consistent naming convention for parameters and be as consistent as possible with sklearn. For instance:
- In supervised versions of weakly supervised algorithms, `num_constraints` should be renamed `n_constraints`, `num_chunks` to `n_chunks`
- In LMNN, the parameter `k` could be renamed `n_neighbors` like in sklearn's `KNeighborsClassifier`
- There is also `tol` and `convergence_threshold` which are both used to refer to optimization tolerance (we should always use `tol` which is quite standard, cf `scipy.optimize)

mvargas33:
Closes #257 as suggested by @bellet . Tests pass as expected.

This refactor is relevant from a user perspective.

### DIFF
--- a/bench/benchmarks/iris.py
+++ b/bench/benchmarks/iris.py
@@ -11,7 +11,7 @@ CLASSES = {
     'LSML_Supervised': metric_learn.LSML_Supervised(n_constraints=200),
     'MLKR': metric_learn.MLKR(),
     'NCA': metric_learn.NCA(max_iter=700, n_components=2),
-    'RCA_Supervised': metric_learn.RCA_Supervised(dim=2, num_chunks=30,
+    'RCA_Supervised': metric_learn.RCA_Supervised(dim=2, n_chunks=30,
                                                   chunk_size=2),
     'SDML_Supervised': metric_learn.SDML_Supervised(n_constraints=1500)
 }
--- a/bench/benchmarks/iris.py
+++ b/bench/benchmarks/iris.py
@@ -5,15 +5,15 @@ import metric_learn
 
 CLASSES = {
     'Covariance': metric_learn.Covariance(),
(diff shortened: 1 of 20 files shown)


terrytangyuan (approved):
LGTM. PTAL @bellet

bellet (changes requested):
@mvargas33 We should avoid breaking compatibility right away: instead, you should keep older names with 'deprecated' value and have deprecation warnings for the next release (0.7.0), before we remove them permanently in 0.8.0.

See for instance what was done here: https://github.com/scikit-learn-contrib/metric-learn/pull/193

mvargas33:
> @mvargas33 We should avoid breaking compatibility right away: instead, you should keep older names with 'deprecated' value and have deprecation warnings for the next release (0.7.0), before we remove them permanently in 0.8.0.
> 
> See for instance what was done here: #193

@bellet 

So, the existing code is not broken anymore. Following #193 I've added a warning mentioning deprecation in each part of the code. These warnings are `FutureWarnings` in order to show them to end users. If I use `DeprecationWarning` the warning won't show up unless the user configure their program to show more warnings (according to [docs](https://docs.python.org/3/library/warnings.html) and tested myself).

To take note:

- The pattern is to change the old name with a 'deprecated' value by default. If its not, the warning is triggered, and the value is assigned to the variable with the new name.
- The old name must have a value, otherwise some tests get broken, for instance, the `check_estimator` test from sklearn, checks that all params received in the constructor are set in the `Estimator`, including the deprecated one, even if it is not used anywhere.
- All tests have the new name now, not the old one
- No additional test were made to check if warnings are being triggered
- Maybe the docs needs an additional modification, just like in #193, but I did not do modified the docs.

perimosocordiae:
Changes look good, merging.

PR #329, opened 2021-09-21
[WIP] Bilinear similarity

mvargas33:
Hi!

In the aim of implementing [OASIS](https://www.jmlr.org/papers/volume11/chechik10a/chechik10a.pdf) algorithm, I open this PR to discuss the implementation of the BilinearMixin class. Which is a requirement for the algorithm.

I've made some testing but in a file outside the test folder, as I am not familiar with the correct way to test things out. These are basic testing for bilinear similarity with random and handmade toy examples.

As the matrix W in the bilinear similarity is not guaranteed to be symmetric nor PSD, I did not extend from the class MetricTransformer, and did not implemented the `transform()` method.

Also, for `score_pairs()` I propose two implementations, and I am not sure which one perms better.


metric_learn/base_metric.py
```diff
@@ -161,6 +161,68 @@ def transform(self, X):
...
+      """
+      check_is_fitted(self, ['preprocessor_', 'components_'])
+      pairs = check_input(pairs, type_of_inputs='tuples',
+                          preprocessor=self.preprocessor_,
+                          estimator=self, tuple_size=2)
+      # Note: For bilinear order matters, dist(a,b) != dist(b,a)
+      # We always choose first pair first, then second pair
+      # (In contrast with Mahalanobis implementation)
+      # I dont know wich implementation performs better
+      return np.diagonal(np.dot(
```
bellet, line 188: I think this should be faster as it performs all operations in vectorized form
You could verify this using `%timeit` or something similar

metric_learn/base_metric.py
```diff
@@ -161,6 +161,68 @@ def transform(self, X):
...
+            The first point involved in the distance computation.
+
+          v : array-like, shape=(n_features,)
+            The second point involved in the distance computation.
+
+          Returns
+          -------
+          distance : float
+            The distance between u and v according to the new metric.
+          """
```
bellet, line 214: comments should be updated to reflect that we do not have a distance anymore but a similarity

metric_learn/oasis.py
```diff
@@ -0,0 +1,24 @@
+from .base_metric import BilinearMixin
+import numpy as np
+
+
+class OASIS(BilinearMixin):
```
bellet, line 5: for now you could move this to the test file and define a `DummyBilinearMixin` class there for testing purposes

test_bilinear.py
```diff
@@ -0,0 +1,44 @@
...
+    # Symetric example
+    u = np.array([0, 1, 2])
+    v = np.array([3, 4, 5])
+
+    mixin.components_ = np.identity(3)  # Identity matrix
+    dists = mixin.score_pairs([[u, v], [v, u]])
+    assert_array_almost_equal(dists, [14, 14])
+
+
+test_toy_distance()
```
bellet, line 44: not needed

test_bilinear.py
```diff
@@ -0,0 +1,44 @@
...
+    dists = mixin.score_pairs([[u, v], [v, u]])
+    assert_array_almost_equal(dists, [96, 120])
+
+    # Symetric example
+    u = np.array([0, 1, 2])
+    v = np.array([3, 4, 5])
+
+    mixin.components_ = np.identity(3)  # Identity matrix
+    dists = mixin.score_pairs([[u, v], [v, u]])
+    assert_array_almost_equal(dists, [14, 14])
```
bellet, line 41: here you are testing multiple things (not always clear what). It may be better to split between several tests. Also, you should probably take inspiration from `test_mahalanobis_mixin.py`: many ideas/toy examples can probably be re-used or adapted

bellet (commented):
Thanks @mvargas33 for getting this started! Here are some comments.

bellet:
You should also ensure the compatibility with pairs, triplets and quadruplet classifiers to make sure we can then build actual bilinear similarity learning algorithms. Probably many existing tests for these will just naturally pass with bilinear similarity, while others may need to be adapted

metric_learn/base_metric.py
bellet: @mvargas33 you should do simply this on your laptop and post here the results as a comment - there is no need to include it in a test :-)

bellet: Just realizing that the first option is bad because it computes the similarities between all `n_pairs*n_pairs` pairs!
There should be a way to avoid this while keeping things full vectorized

mvargas33: I ran a test with `timeit` computing similarity for pairs of for 1.000 samples of 10.000 features each. Got these results:
```
First method took 2.203281 s
Second method took 3.51009 s
```
Where first method is the one with `np.diagonal()`, and the second the one with `zip()`

Will stick to first one for now, will try to find a way to compute it more efficiently

bellet: Yes, because the problem with the first option is that it takes quadratic computation and memory in the number of pairs, which is clearly suboptimal

(I think if you increase the number of pairs in your timing test, you will see that the second method becomes faster)

This should be the fastest / more efficient way to compute it:
`np.sum(np.dot(M, pairs[:, 1, :].T) * pairs[:, 0, :].T, axis=0)`

mvargas33: Found a better solution [here](https://stackoverflow.com/questions/14758283/is-there-a-numpy-scipy-dot-product-calculating-only-the-diagonal-entries-of-the)

For 1e4 samples and 1e4 features these are the results:
```
First method took 38.912194710999984
Second method took 50.40531421000014
Third method took 23.76173380199998
```
Will stick to this better solution

metric_learn/base_metric.py
mvargas33: All comments have been updated stating this is in fact a similarity.

The only thing that makes me noise is the `get_metric` method. It might be `get_similarity`, but we can't do that as `get_metric` is inherited from `BaseMetricLearner`. But documentation should be enough.

test_bilinear.py
mvargas33: Deleted

test_bilinear.py
mvargas33: I moved all testing to a `test/` in a `test_bilinear_mixin.py` file.

metric_learn/oasis.py
mvargas33: I just moved all to a testing folder and created a `DummyBilinearMixin` for testing purposes

metric_learn/base_metric.py
bellet: Yes, this is fine I think

mvargas33:
@bellet @perimosocordiae @wdevazelhes @terrytangyuan 

There needs to be a broader discussion regarding adding BilinearSimilarity and Bilinear Similarity Learners such as OASIS.

By definition bilinear similarity works in the opposite direction as a metric such as the Mahalanobis.

Given two vectors `u, v` in `Rn`, if their bilinear similarity `S(u,v)` is positive and huge, it means the vectors are closer. On the opposite, if `S(u,v)` is negative, it means the vectors point in different directions, and are far away. In contrast with a metric where a huge value means the vectors are far far away.

Regarding the API, it means that `score_pairs()` works semantically in opposite directions for `MahalanobisMixin` and `BilinearMixin` while both implement it, as both inherit from `BaseMetricLearner`.

There are several paths that can be followed to deal with this:

1. Add a -1 multiplication at the result of `score_pairs` in `BilinearMixin`, so the global semantic of that function across the whole package remains the same. And add a warning at the documentation. 
2. Don't change `score_pairs()` and return the original bilinear similarity: This may conflict when comparing different models, as the user needs to take cake of making an exceptional case for Bilinear learners considering that huge values of `score_pairs` means closer points.
3. Don't inherit `BilinearMixin` from `BaseMetricLearner` as the linear similarity is not a metric, its semantically incorrect. Create a whole different class/part of the code for `SimilarityLearners`
4. Other options I don't see right now

This is also a concern for `Classifiers`, as they rely on `score_pairs()` at `decision_function()` (and at `predict()` as consequence) for verifying triplets/cuadriplets. And measuring _distance_ for pairs.

For example, if we choose 2) and not change `score_pairs()` for bilinear, we will have to use `isinstance()` at classifiers to make the difference for `MahalanobiasMixin` and the `BilinearMixin`. Another option is to build almost the same classifiers, but for `BilinearMixin` exclusively, duplicating a lot of code.

The decision made now, might imply many things for the future, moreover if there comes more bilinear learners implementations in the future, or if there are other similarities apart from bilinear and learners for those other similarities.

My humble opinion: I believe its more important to keep the global semantic of `score_pairs()`, even when the "metric" learned is in fact a similarity. I bet for the -1 multiplier in BilinearMixin and rely that the final user will see a huge and clear warning at Bilinear Learners. Maybe implement an extra function that returns the original similarities (without -1 mult.) although that idea seems silly.

Ps: By "semantic" I mean the meaning of what the function is intended to do. For instance, the current semantic of `score_pairs()` means that the lower the output value, the closer the pairs.

wdevazelhes:
> @bellet @perimosocordiae @wdevazelhes @terrytangyuan
> 
> There needs to be a broader discussion regarding adding BilinearSimilarity and Bilinear Similarity Learners such as OASIS.
> 
> By definition bilinear similarity works in the opposite direction as a metric such as the Mahalanobis.
> 
> Given two vectors `u, v` in `Rn`, if their bilinear similarity `S(u,v)` is positive and huge, it means the vectors are closer. On the opposite, if `S(u,v)` is negative, it means the vectors point in different directions, and are far away. In contrast with a metric where a huge value means the vectors are far far away.
> 
> Regarding the API, it means that `score_pairs()` works semantically in opposite directions for `MahalanobisMixin` and `BilinearMixin` while both implement it, as both inherit from `BaseMetricLearner`.
> 
> There are several paths that can be followed to deal with this:
> 
> 1. Add a -1 multiplication at the result of `score_pairs` in `BilinearMixin`, so the global semantic of that function across the whole package remains the same. And add a warning at the documentation.
> 2. Don't change `score_pairs()` and return the original bilinear similarity: This may conflict when comparing different models, as the user needs to take cake of making an exceptional case for Bilinear learners considering that huge values of `score_pairs` means closer points.
> 3. Don't inherit `BilinearMixin` from `BaseMetricLearner` as the linear similarity is not a metric, its semantically incorrect. Create a whole different class/part of the code for `SimilarityLearners`
> 4. Other options I don't see right now
> 
> This is also a concern for `Classifiers`, as they rely on `score_pairs()` at `decision_function()` (and at `predict()` as consequence) for verifying triplets/cuadriplets. And measuring _distance_ for pairs.
> 
> For example, if we choose 2) and not change `score_pairs()` for bilinear, we will have to use `isinstance()` at classifiers to make the difference for `MahalanobiasMixin` and the `BilinearMixin`. Another option is to build almost the same classifiers, but for `BilinearMixin` exclusively, duplicating a lot of code.
> 
> The decision made now, might imply many things for the future, moreover if there comes more bilinear learners implementations in the future, or if there are other similarities apart from bilinear and learners for those other similarities.
> 
> My humble opinion: I believe its more important to keep the global semantic of `score_pairs()`, even when the "metric" learned is in fact a similarity. I bet for the -1 multiplier in BilinearMixin and rely that the final user will see a huge and clear warning at Bilinear Learners. Maybe implement an extra function that returns the original similarities (without -1 mult.) although that idea seems silly.
> 
> Ps: By "semantic" I mean the meaning of what the function is intended to do. For instance, the current semantic of `score_pairs()` means that the lower the output value, the closer the pairs.

@mvargas33 I agree with you, IMHO I think it would be good to reuse the code and existing classes/methods as much as possible: 
- Regarding whether to multiply by -1 or not for `score_pairs` for `BilinearMixin`,I agree it would be good, so that indeed `score_pairs` is "in the same direction" for all metric learners, which would make the code for `Classifiers` keep working for `Bilinear` learners, and allow users to write some generic code with `score_pairs`. I agree that, as a consequence, we would need to ensure that users know what they're doing when using `score_pairs`, and a warning in the documentation sounds good.
&#8594; Maybe also an idea for even more clarity would be to revisit `score_pairs` by changing its name to something more self-explanatory like `pairs_remoteness` (this is probably a bad name but any synonym of distance would work (distance itself may maybe not be ideal if we want to return things that are not true distances), to keep the idea "small=closer, big=further away). And maybe having the opposite function as you suggested like say `pairs_similarity` would be good too: it can be good because in the documentation for the instantiated versions of those methods, we would for instance say for Bilinear learners that `pairs_similarity` is the actual bilinear similarity (from the math formula), and for Mahalanobis learners that `pairs_remoteness` is the actual mahalanobis distance (from the math formula), and for the remaining functions (`pairs_remoteness` for Bilinear learners and `pairs_similarity` for the Mahalanobis learners), we would just say it's the opposite of the other one,  given for convenience. This way may be a good compromise between having on one hand a generic code (useful for us in classes like `Classifier` and for users who write generic code), and on the other hand being very transparent and clear to the user on what function does what
And this way we may even not need the warning.
But maybe having two methods is a bit too much, I'm not sure, I guess either way are good (2 methods, or 1 method with the right warning in docs)

(Note: also regarding the name `pairs_remoteness`, I guess we could either have a generic name like this and say for Mahalanobis learners that it is equal to the true distance, OR have a precise name like `pairs_distance` and add a warning that says for `Bilinar` learners that it is not a true distance (in the mathematical sense, since (- bilinear similarity) can be negative) , I guess either way works)

NB: Probably the `get_metric` function would need to be adapted too, the same way as `score_pairs` ? 

- Regarding inheriting from `BaseMetricLearner`, I think it would be good indeed, we can even see `BaseMetricLearner` not as a generic class for distance metric learners, but like for learners of the package "metric-learn" :D, so in this case Bilinear learners fits in ;) (Or otherwise we could rename that abstract class too into just `BaseLearner` for instance ;) )

mvargas33:
@wdevazelhes 

I like the idea of `pairs_remoteness` and `pairs_similarity`, because it adds one layer of abstraction, making the code more generic and being able to support metrics and similarites without complicating things too much. No change will be needed in `Classifiers` that way.

Even if this looks like an _overkill_ right now, it guarantees that the code will be scalable and easily extendible in the future. For instance, imagine there are 5 bilinear similarity learners at some point, How this deals with the solution of -1 multiplier + warning? For me that sounds more like a patch.

Or what if there's support for data-structure similarities in the future, like tree-similarities or string-similarities. This adds more similarity types that could be easily handled with `pairs_remoteness` and `pairs_similarity`, no matter what the algorithm actually calculates in the end: closeness or remoteness.

I also like the idea of renaming `BaseMetricLearner` to `BaseLearner`, again, because the second its more abstract and semantically correct with metrics and similarities.

wdevazelhes:
@mvargas33 That sounds good ! Yes that's right, having two general functions like this would make the code quite generic and extensible

bellet:
Hi, thanks @mvargas33 and @wdevazelhes for kicking off the discussion. Here are some comments:
- I agree that we should act at the level of the abstract class `BaseMetricLearner` and make the semantic of the abstract methods consistent, so that everything downstream (e.g., classifiers) work seamlessly.
- I think we chose the current name of the method `score_pairs` in analogy to sklearn's  `score` method, but somehow the semantic is reversed compared to the latter when considering Mahalanobis learners (larger score means less positive), which is counter-intuitive. In hindsight I don't think we did the right choice here and I am in favor of changing this one way or another.
- Given the above, we have several choices.
  1. Either we stick to `score_pairs` but adjust the semantic to always follow the convention "larger = more similar". This is the easiest change but it means changing the current behavior for Mahalanobis learners, which may break existing code (of course we can add warnings but this may not be enough for people to become aware of the problem).
  2. Or we decide it is better to change the name of the method (and deprecate `score_pairs`, giving users time to adapt). We could use `pair_score` if we want to stick close to what we had. In this case, as suggested by @wdevazelhes, we could also consider introducing two methods (one for similarity and one for dissimilarity/distance). Maybe this could be the default and we could implement `pair_distance` only when the metric is actually a true (pseudo) distance? Otherwise, the two methods may be the same in practice up to a minus sign, which is not that useful.
- "Metric learning" have been used extensively in the literature as broad terms encompassing both metric learning and similarity learning. Note incidentally that in general Mahalanobis "distances" are not true metrics (only pseudo-distances). I think we should follow this terminology (probably explain this clearly in the doc), and thus for instance keep the name `BaseMetricLearner` (the name `BaseLearner` becomes way too generic IMO).
- Similarly, I am not convinced at this point that `get_metric` needs to be changed. BTW, is there any sklearn estimator that has an argument for providing a similarity instead of a metric?

Would be great to hear @perimosocordiae and @terrytangyuan's take on this!

wdevazelhes:
@bellet Yes ​I agree the deprecation sounds good, to  force users to change their code, and I agree one general method that every learner would implement (like similarity, which respects more the meaning of a "score" indeed) may be enough for having generic code (for us, and for our users), and then a method like `pairwise_distances` for (pseudo)distance metric learners is indeed cool to have: (they even have such a function with the same name in sklearn too: `https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html`)
- For `get_metric` you're right, in the end this would be mostly be used as a function in sklearn's estimators so no real need to extend it I guess
- Regarding your question about whether sklearn's estimators sometimes take in a similarity: I just checked and it seems it can, with for instance `GaussianProcesses` (see https://scikit-learn.org/stable/modules/gaussian_process.html#kernels-for-gaussian-processes): there's even an interesting paragraph in the user guide about API, I didn't fully understood how it all works but I seems like it could definitely be a source of inspiration: 
`
All Gaussian process kernels are interoperable with sklearn.metrics.pairwise and vice versa: instances of subclasses of Kernel can be passed as metric to pairwise_kernels from sklearn.metrics.pairwise. Moreover, kernel functions from pairwise can be used as GP kernels by using the wrapper class PairwiseKernel.
`
Also, it seems that bilinear metrics learn in fact a linear kernel in a linearly transformed space is that right ? If yes I guess it'd be cool if they return an object that can be used as a kernel in scikit-learn `GaussianProcess` !
Also, this part of the documentation seems interesting: https://scikit-learn.org/stable/modules/gaussian_process.html#gaussian-process-kernel-api It seems that in returning the kernel object, it is even possible to specify what are some hyperparameters (in this case maybe the matrix A in xTAy ?), and what is the gradient of the kernel function with respect to them: this may mean that scikit-learn could fine-tune a kernel previously learnt by say OASIS, in order to optimize a gaussian process, (not sure if that would work, even if that would be useful in some real life examples, but it definitely shows the power of the API!)

NB: this is also another scikit-learn doc that may be useful: https://scikit-learn.org/stable/modules/metrics.html#metrics

bellet:
@wdevazelhes the thing is that a bilinear similarity is not guaranteed to yield a data transformation (and thus be a valid kernel). This is only true when the matrix is PSD, which is often not enforced by the similarity learning algorithms to avoid the computational cost with the projection into the PSD cone. So basically, a bilinear metric learner will not always be a transformer or valid kernel (and in practice, may often not be). We could still implement `transform` and raise an error when the matrix is not PSD, but this may be confusing for users?

@mvargas33 at this point of the discussion I think we all agree that it is a good idea to:
- deprecate the `score_pairs` method
- add instead a `pairwise_similarities` method which follows the convention (the higher, the more similar)
- add a `pairwise_distances` method which is only implemented for metric learner that learn an actual (pseudo) distance, ie implemented for Mahalanobis but not bilinear similarity.

So perhaps you can go ahead with the above changes, and we can continue the discussion a bit regarding what to do for bilinear similarity (do we want to have specific things for when the matrix is actually PSD). @terrytangyuan @perimosocordiae your opinion is most welcome!

bellet:
This is always a feature we can add later if we feel it is needed.
@mvargas33 I think we have some kind of consensus now so you may go ahead and start implementing these changes

doc/introduction.rst
```diff
@@ -79,6 +79,34 @@ necessarily the identity of indiscernibles.
...
+.. _bilinear_similarity:
+
+Bilinear Similarity
+===================
+
+Some algorithms in the package don't learn a distance or pseudo-distance, but
+a similarity. The idea is that two pairs are closer if their similarity value
+is high, and viceversa. Given a real-valued parameter matrix :math:`W` of shape
+``(n_features, n_features)`` where ``n_features`` is the number features
+describing the data, the Bilinear Similarity associated with :math:`W` is
```
bellet, line 91: I would use lower-case: "bilinear similarity"
(the reason why Mahalanobis is upper-cased is because it is a name)

doc/introduction.rst
```diff
@@ -79,6 +79,34 @@ necessarily the identity of indiscernibles.
...
   the metric learning problem with respect to either :math:`M` or :math:`L`.
 
+.. _bilinear_similarity:
+
+Bilinear Similarity
+===================
+
+Some algorithms in the package don't learn a distance or pseudo-distance, but
+a similarity. The idea is that two pairs are closer if their similarity value
+is high, and viceversa. Given a real-valued parameter matrix :math:`W` of shape
```
bellet, line 89: "Some algorithms in the package learn bilinear similarity functions. These similarity functions are not pseudo-distances: they simply output real values such that the larger the similarity value, the more similar the two examples.

doc/introduction.rst
```diff
@@ -79,6 +79,34 @@ necessarily the identity of indiscernibles.
...
+``(n_features, n_features)`` where ``n_features`` is the number features
+describing the data, the Bilinear Similarity associated with :math:`W` is
+defined as follows:
+
+.. math:: S_W(x, x') = x^T W x'
```
bellet, line 94: ^T --> ^\top

doc/introduction.rst
```diff
@@ -79,6 +79,34 @@ necessarily the identity of indiscernibles.
...
+
+The matrix :math:`W` is not required to be positive semi-definite (PSD), so
+none of the distance properties are satisfied: nonnegativity, identity of
```
bellet, line 97: so the distance properties (nonnegativity, identity of indiscernibles, symmetry and triangle inequality) do not hold in general.

doc/introduction.rst
```diff
@@ -79,6 +79,34 @@ necessarily the identity of indiscernibles.
...
+Some algorithms in the package don't learn a distance or pseudo-distance, but
+a similarity. The idea is that two pairs are closer if their similarity value
+is high, and viceversa. Given a real-valued parameter matrix :math:`W` of shape
+``(n_features, n_features)`` where ``n_features`` is the number features
+describing the data, the Bilinear Similarity associated with :math:`W` is
+defined as follows:
+
+.. math:: S_W(x, x') = x^T W x'
+
+The matrix :math:`W` is not required to be positive semi-definite (PSD), so
```
bellet, line 96: is not required to be positive semi-definite (PSD) or even symmetric

doc/introduction.rst
```diff
@@ -79,6 +79,34 @@ necessarily the identity of indiscernibles.
...
+none of the distance properties are satisfied: nonnegativity, identity of
+indiscernibles, symmetry and triangle inequality.
+
+This allows some algorithms to optimize :math:`S_W` in an online manner using a
+simple and efficient procedure, and thus can be applied to problems with
+millions of training instances and achieves state-of-the-art performance
+on an image search task using :math:`k`-NN.
```
bellet, line 103: this is specific to OASIS. nothing says that these things are not possible also for Mahalanobis distances

I would remove this entirely and replace by a sentence at the beginning of the next paragraph: "The absence of PSD constraint can enable the design of more efficient algorithms."
(and then "It also...")

doc/introduction.rst
```diff
@@ -79,6 +79,34 @@ necessarily the identity of indiscernibles.
...
+
+It also allows to be applied in contexts where the triangle inequality is
+violated by visual judgements and the goal is to approximate perceptual
+similarity. For intance, a man and a horse are both similar to a centaur,
+but not to one another.
```
bellet, line 108: Here let's simplify a bit as well and only keep one sentence:
"It is also relevant in applications where the underlying notion of similarity does not satisfy the triangle inequality, as known to be the case for visual judgments".

doc/introduction.rst
```diff
@@ -103,6 +131,10 @@ examples (for code illustrating some of these use-cases, see the
   the data into a new embedding space before feeding it into another machine
   learning algorithm.
 
+The most common use-case of metric-learn would be to learn a Mahalanobis metric,
+then transform the data to the learned space, and then resolve one of the task
+above.
```
bellet, line 136: this is confusing, and somehow make it seems like bilinear similarity is not useful. It is also redundant with the last item of the previous list. I would just remove but modify slightly the last item (which refers to matrix L of Mahalanobis):

"More generally with Mahalanobis distances, the learned transformation L can be used to project the data into a new embedding space before feeding it into another machine learning algorithm."

doc/supervised.rst
```diff
@@ -41,32 +41,43 @@ two numbers.
...
-classes will be large. To do so, we fit the metric learner (example:
-`NCA`).
+classes will be large.
+
+But there are also some algorithms that learn a similarity, not a distance,
+thus the points cannot be transformed into a new space. In this case, the
+utility comes at using the similarity bewtween points directly.
+
+Mahalanobis learners can transform points into a new space, and to do so,
+we fit the metric learner (example:`NCA`).
```
bellet, line 54: We need to reformulate this: the initial sentence is too restrictive now, and keeping it induces some confusion.

The goal of supervised metric learning algorithms is to learn a (distance or similarity) metric such that two points from the same class will be similar (e.g., have small distance) and points from different classes will be dissimilar (e.g., have large distance).

To do so, we first need to fit the supervised metric learner on a labeled dataset, as in the example below with `NCA`.

doc/supervised.rst
```diff
@@ -41,32 +41,43 @@ two numbers.
...
 
-Also, as explained before, our metric learners has learn a distance between
-points. You can use this distance in two main ways:
+.. warning::
+    
+    If you try to use `transform` with a similarity learner, an error will
+    appear, as you cannot transform the data using them.
+
+Also, as explained before, mahalanobis metric learners learn a distance
+between points. You can use this distance in two main ways:
```
bellet, line 80: Need to change the order to first talk about the more generic things, and then talk about about specifics for Mahalanobis.
- Then present `pair_score` by saying "We can now use the learned metric to score new pairs of points (the larger the score, the more similar the pair)." ...
- Then say that for metric learners that learn a distance metric, there is also the `pair_distance` method
- Mention `get_metric`
- Finally, say that as explained in [link Mahalanobis distance section of "What is metric learning"], Mahalanobis distances are equivalent to the Euclidean distance in a transformed space, and can thus be used to transform data points in a new embedding space.

doc/supervised.rst
```diff
@@ -94,17 +106,48 @@ This is useful because `pair_score` matches the **score** semantic of
 scikit-learn's `Classification metrics
 <https://scikit-learn.org/stable/modules/model_evaluation.html#classification-metrics>`_.
 
+For similarity learners `pair_distance` is not available, as they don't learn
+a distance. Intead you use `pair_score` that has the same behaviour but
+for similarity.
```
bellet, line 111: based on the suggested organization above, this sentence should not be needed; you can simply re-use the warning you have below

doc/supervised.rst
```diff
@@ -94,17 +106,48 @@ This is useful because `pair_score` matches the **score** semantic of
...
+
+>>> algorithm.pair_score([[[3.5, 3.6], [5.6, 2.4]], [[1.2, 4.2], [2.1, 6.4]], [[3.3, 7.8], [10.9, 0.1]]])
+array([-0.2312, 705.23, -72.8])
+
+.. warning::
+    
+    If you try to use `pair_distance` with a similarity learner, an error
```
bellet, line 118: let's say "bilinear similarity learner" because "similarity learner" is perhaps not well defined

doc/supervised.rst
```diff
@@ -94,17 +106,48 @@ This is useful because `pair_score` matches the **score** semantic of
...
+    will appear, as they don't learn a distance nor a pseudo-distance.
```
bellet, line 119: will be thrown, as it does not learn a distance.

doc/supervised.rst
```diff
@@ -94,17 +106,48 @@ This is useful because `pair_score` matches the **score** semantic of
...
+available. You can interpret that this function returns the **score**
+between points: the more the **score**, the closer the pairs and vice-versa.
+For mahalanobis learners, it is equal to the inverse of the distance.
+
 .. note::
 
     If the metric learner that you use learns a :ref:`Mahalanobis distance
-    <mahalanobis_distances>` (like it is the case for all algorithms
-    currently in metric-learn), you can get the plain learned Mahalanobis
+    <mahalanobis_distances>`, you can get the plain learned Mahalanobis
```
bellet, line 136: remove "plain" and perhaps name the matrix M for clarity

doc/supervised.rst
```diff
@@ -94,17 +106,48 @@ This is useful because `pair_score` matches the **score** semantic of
...
     matrix using `get_mahalanobis_matrix`.
 
     >>> nca.get_mahalanobis_matrix()
     array([[0.43680409, 0.89169412],
            [0.89169412, 1.9542479 ]])
 
+    If the metric learner that you use learns a :ref:`Bilinear similarity
+    <bilinear_similarity>`, you can get the plain learned Bilinear
```
bellet, line 144: remove "plain" and perhaps name the matrix W

doc/weakly_supervised.rst
```diff
@@ -127,9 +127,16 @@ through the argument `preprocessor` (see below :ref:`fit_ws`)
...
+Generally, the goal of weakly-supervised metric-learning algorithms is to
+transform points in a new space, in which the tuple-wise constraints between
+points are respected.
+
+But there are also some algorithms that learn a similarity, not a distance,
+thus the points cannot be transformed into a new space. But the goal is the
+same: respect the maximum number of constraints while learning the similarity.
+
+Weakly-supervised mahalanobis learners can transform points into a new space,
+and to do so, we fit the metric learner (example:`MMC`).
```
bellet, line 139: Similar as for supervised learners:

The goal of weakly supervised metric learning algorithms is to learn a (distance or similarity) metric such that the tuple-wise constraints between points are respected.

To do so, we first need to fit the weakly supervised metric learner on a dataset of tuples, as in the example below with `MMC`.

doc/weakly_supervised.rst
```diff
@@ -144,18 +151,23 @@ Or alternatively (using a preprocessor):
...
 >>> mmc.transform(X_new)
 array([[-3.24667162e+01,  4.62622348e-07,  3.88325421e-08],
        [-3.61531114e+01,  4.86778289e-07,  2.12654397e-08]])
 
+.. warning::
+    
+    If you try to use `transform` with a similarity learner, an error will
+    appear, as you cannot transform the data using them.
+
 Also, as explained before, our metric learner has learned a distance between
```
bellet, line 171: re-organize as for supervised learners

bellet (changes requested):
Here's a review for the documentation part.

Generally we need to change a bit the vocabulary to accommodate both Mahalanobis and bilinear similarities. As common in the literature I propose to use the term "metric" as a generic term (encompassing both distances and similarities).

In addition to the changes mentioned below, some changes are needed in the beginning of "What is metric learning?":

Sec 1
- Many approaches in machine learning require a measure of distance (or similarity) between data points.
- choose a standard distance metric --> choose a standard metric 
- Distance metric learning (or simply, metric learning) --> Metric learning
- task-specific distance metrics --> task-specific metrics
- The learned distance metric --> the learned metric

Sec 1.1:
- the goal in this setting is to learn a distance metric --> a metric
- the goal is to learn a distance metric --> a metric
- find the parameters of a distance function --> of a metric function

**Please do spellcheck to avoid typos**!

bellet:
> 
>     * There are also some test that are identical for Mahalanobis and Bilinear learners, but they are in their respective test file. Should we have a test file for tests regarding all kinds of learners? Now there is none, but all learners are tested in classifiers tests. Take as an example: `test_pair_score_dim` and `test_deprecated_score_pairs_same_result`.

Yes, tests that are common to Mahalanobis and Bilinear linears (and thus most likely to any metric learner) should not be duplicated and placed in `test_base_metric.py` instead

doc/introduction.rst
```diff
@@ -79,6 +79,34 @@ necessarily the identity of indiscernibles.
   parameterizations are equivalent. In practice, an algorithm may thus solve
   the metric learning problem with respect to either :math:`M` or :math:`L`.
 
+.. _bilinear_similarity:
+
+Bilinear Similarity
```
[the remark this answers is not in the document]
mvargas33: Unified to plural

doc/introduction.rst
mvargas33: Changed to lower case

doc/introduction.rst
mvargas33: Changed

doc/introduction.rst
mvargas33: Changed.

doc/introduction.rst
mvargas33: Changed.

doc/introduction.rst
mvargas33: Yeah, I took that paragraph from the Metric Learning book, I guess it was the paragraph in between Bilinear Similarity and OASIS so that's why it's so specific. I removed it and added the sentence you mention in the next paragraph.

doc/introduction.rst
mvargas33: Changed.

doc/introduction.rst
mvargas33: Removed the redundant paragraph and changed the last item of the list.

doc/introduction.rst
mvargas33: Changed.

doc/supervised.rst
mvargas33: Changed.

doc/supervised.rst
mvargas33: Changed.

doc/supervised.rst
mvargas33: Changed.

doc/supervised.rst
mvargas33: Changed.

doc/supervised.rst
mvargas33: Changed.

doc/supervised.rst
mvargas33: Removed and simplified.

doc/supervised.rst
mvargas33: Did the reordering refactor, and rephrased some parts.

doc/weakly_supervised.rst
mvargas33: Reorganized just like supervised learners

mvargas33:
> Here's a review for the documentation part.
> 
> Generally we need to change a bit the vocabulary to accommodate both Mahalanobis and bilinear similarities. As common in the literature I propose to use the term "metric" as a generic term (encompassing both distances and similarities).
> 
> In addition to the changes mentioned below, some changes are needed in the beginning of "What is metric learning?":
> 
> Sec 1
> 
>     * Many approaches in machine learning require a measure of distance (or similarity) between data points.
> 
>     * choose a standard distance metric --> choose a standard metric
> 
>     * Distance metric learning (or simply, metric learning) --> Metric learning
> 
>     * task-specific distance metrics --> task-specific metrics
> 
>     * The learned distance metric --> the learned metric
> 
> 
> Sec 1.1:
> 
>     * the goal in this setting is to learn a distance metric --> a metric
> 
>     * the goal is to learn a distance metric --> a metric
> 
>     * find the parameters of a distance function --> of a metric function
> 
> 
> **Please do spellcheck to avoid typos**!

All observations were resolved for `introduction.rst`, `supervised.rst` and `weakly_supervised.rst`.

Will run the spellchecker soon as well as moving the common tests to `test_base_metric.py`

mvargas33:
- I ran the spellcheck and fixed a few typos.
- I moved `test_pair_score_dim` and `test_deprecated_score_pairs_same_result` to `test_base_metric.py`
- I refactored `test_same_with_or_without_preprocessor` to run with all metric learners and used `isinstance(estimator, MahalanobisMixin)` to extend it for `transform` and `pair_distance`, but avoiding having a duplicated/separated test for Mahalanobis/Bilinear only.

doc/supervised.rst
```diff
@@ -62,87 +58,71 @@ NCA(init='auto', max_iter=100, n_components=None,
...
 
 >>> similarity_fun = algorithm.get_metric()
 >>> similarity_fun([3.5, 3.6], [5.6, 2.4])
 -0.04752
 
-For similarity learners and mahalanobis learners, `pair_score` is
-available. You can interpret that this function returns the **score**
-between points: the more the **score**, the closer the pairs and vice-versa.
-For mahalanobis learners, it is equal to the inverse of the distance.
+Finally, as explained in :ref:`mahalanobis_distances`, these are equivalent to the Euclidean
```
bellet, line 99: you may need to replace "these" by "Mahalanobis distances" here depending on how the ref is displayed once the doc is compiled?

doc/weakly_supervised.rst
```diff
@@ -149,94 +147,78 @@ Or alternatively (using a preprocessor):
...
-  scikit-learn's `Classification metrics
-  <https://scikit-learn.org/stable/modules/model_evaluation.html#classification-metrics>`_.
+>>> similarity_fun = algorithm.get_metric()
+>>> similarity_fun([3.5, 3.6], [5.6, 2.4])
+-0.04752
 
-For similarity learners `pair_distance` is not available, as they don't learn
-a distance. Intead you use `pair_score` that has the same behaviour but
-for similarity.
+Finally, as explained in :ref:`mahalanobis_distances`, these are equivalent to the Euclidean
```
bellet, line 194: same question here

doc/weakly_supervised.rst
```diff
@@ -149,94 +147,78 @@ Or alternatively (using a preprocessor):
...
->>> mmc.fit(pairs_indice, y)
+>>> mmc.fit(pairs_indices, y)
 
 Now that the estimator is fitted, you can use it on new data for several
 purposes.
 
-First, your mahalanobis learner can transform the data in
-the learned space, using `transform`: Here we transform two points in
-the new embedding space.
+We can now use the learned metric to **score** new pairs of points with ``pair_score``
```
bellet, line 155: same here about now

test/test_base_metric.py
```diff
@@ -296,5 +300,56 @@ def test_score_pairs_warning(estimator, build_dataset):
...
+    model.pair_score(tuples[1])
+  assert str(raised_error.value) == msg
+
+
+@pytest.mark.parametrize('estimator, build_dataset', metric_learners,
+                         ids=ids_metric_learners)
+def test_deprecated_score_pairs_same_result(estimator, build_dataset):
+  """
+  Test that `pari_distance` gives the same result as `score_pairs` for
+  Mahalanobis learnes, and the same for `pair_score` and `score_paris`
```
bellet, line 331: typos: pari_distance and score_paris

bellet (commented):
Some nitpicks on the doc. One additional thing to update is in introduction.rst, in the part:

> Strictly speaking, Mahalanobis distances are "pseudo-metrics": they satisfy
three of the `properties of a metric <https://en.wikipedia.org/wiki/Metric_
(mathematics)>`_ (non-negativity, symmetry, triangle inequality) but not
necessarily the identity of indiscernibles.

For consistency of vocabulary, change "pseudo-metrics" into "pseudo-distances" and "properties of a metric" by "properties of a distance"

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
     """
 
 
+class BilinearMixin(BaseMetricLearner, metaclass=ABCMeta):
```
bellet, line 245: this is a detail but I would place this class after the Mahalanobis one, as it is currently less used in the package

bellet, line 245: this is a detail but I would place this class after the Mahalanobis one, as it is currently less used in the package

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+  r"""Bilinear similarity learning algorithms.
+
+  Algorithm that learns a Bilinear (pseudo) similarity :math:`s_M(x, x')`,
+  defined between two column vectors :math:`x` and :math:`x'` by: :math:
+  `s_M(x, x') =  x M x'`, where :math:`M` is a learned matrix. This matrix
```
bellet, line 250: You call the matrix W in the doc, let's stick to the same convention

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+  is not guaranteed to be symmetric nor positive semi-definite (PSD). Thus
+  it cannot be seen as learning a linear transformation of the original
+  space like Mahalanobis learning algorithms.
+
+  Attributes
+  ----------
+  components_ : `numpy.ndarray`, shape=(n_components, n_features)
+    The learned bilinear matrix ``M``.
```
bellet, line 258: W

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+  """
+
+  def score_pairs(self, pairs):
+    dpr_msg = ("score_pairs will be deprecated in release 0.7.0. "
+               "Use pair_score to compute similarity scores, or "
+               "pair_distances to compute distances.")
+    warnings.warn(dpr_msg, category=FutureWarning)
+    return self.pair_score(pairs)
```
bellet, line 266: As this is a new class, I would say that we do not need this method at all as there is no risk that previous code gets broken (since this class did not exist before the switch to `pair_score`).
If you prefer to keep it, then copy the whole docstring, deprecation notice etc as in Mahalanobis case.

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+
+  def pair_distance(self, pairs):
+    """
+    Returns an error, as bilinear similarity learners don't learn a
+    pseudo-distance nor a distance. In consecuence, the additive inverse
+    of the bilinear similarity cannot be used as distance by construction.
+    """
+    msg = ("This learner doesn't learn a distance, thus ",
+           "this method is not implemented. Use pair_score instead")
+    raise Exception(msg)
```
bellet, line 276: Is there a way to implement this behavior by default in `BaseMetric`, so we avoid to have this for all classes that don't implement it and instead only override it for classes that do? Would that be a better practice than what we currently have?
@wdevazelhes @perimosocordiae @terrytangyuan

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
     """
 
 
+class BilinearMixin(BaseMetricLearner, metaclass=ABCMeta):
+  r"""Bilinear similarity learning algorithms.
+
+  Algorithm that learns a Bilinear (pseudo) similarity :math:`s_M(x, x')`,
```
bellet, line 248: Bilinear --> bilinear

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+    Returns an error, as bilinear similarity learners don't learn a
+    pseudo-distance nor a distance. In consecuence, the additive inverse
+    of the bilinear similarity cannot be used as distance by construction.
+    """
+    msg = ("This learner doesn't learn a distance, thus ",
+           "this method is not implemented. Use pair_score instead")
+    raise Exception(msg)
+
+  def pair_score(self, pairs):
+    r"""Returns the learned Bilinear similarity between pairs.
```
bellet, line 279: Bilinear --> bilinear

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+  def score_pairs(self, pairs):
+    dpr_msg = ("score_pairs will be deprecated in release 0.7.0. "
+               "Use pair_score to compute similarity scores, or "
+               "pair_distances to compute distances.")
+    warnings.warn(dpr_msg, category=FutureWarning)
+    return self.pair_score(pairs)
+
+  def pair_distance(self, pairs):
+    """
+    Returns an error, as bilinear similarity learners don't learn a
```
bellet, line 270: do not

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+    pseudo-distance nor a distance. In consecuence, the additive inverse
+    of the bilinear similarity cannot be used as distance by construction.
+    """
+    msg = ("This learner doesn't learn a distance, thus ",
```
bellet, line 274: does not

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+           "this method is not implemented. Use pair_score instead")
+    raise Exception(msg)
+
+  def pair_score(self, pairs):
+    r"""Returns the learned Bilinear similarity between pairs.
+
+    This similarity is defined as: :math:`s_M(x, x') =  x^T M x'`
+    where ``M`` is the learned Bilinear matrix, for every pair of points
+    ``x`` and ``x'``.
```
bellet, line 283: M --> W
Bilinear --> bilinear

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+    See Also
+    --------
+    get_metric : a method that returns a function to compute the similarity
+      between two points. The difference with `pair_score` is that it
+      works on two 1D arrays and cannot use a preprocessor. Besides, the
+      returned function is independent of the similarity learner and hence
+      is not modified if the similarity learner is.
+
+    :ref:`Bilinear_similarity` : The section of the project documentation
+      that describes Bilinear similarity.
```
bellet, line 306: Check that the ref works (you renamed this part as Bilinear Similarities)
Bilinear --> bilinear

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+    ----------
+    pairs : array-like, shape=(n_pairs, 2, n_features) or (n_pairs, 2)
+      3D Array of pairs to score, with each row corresponding to two points,
+      for 2D array of indices of pairs if the similarity learner uses a
+      preprocessor.
+
+    Returns
+    -------
+    scores : `numpy.ndarray` of shape=(n_pairs,)
+      The learned Bilinear similarity for every pair.
```
bellet, line 295: Bilinear --> bilinear

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+    # (In contrast with Mahalanobis implementation)
+    return np.sum(np.dot(pairs[:, 0, :], self.components_) * pairs[:, 1, :],
+                  axis=-1)
+
+  def get_metric(self):
+    check_is_fitted(self, 'components_')
+    components = self.components_.copy()
+
+    def similarity_fun(u, v):
+      """This function computes the similarity between u and v, according to the
```
bellet, line 323: +bilinear

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+        The similarity between u and v according to the new similarity.
+      """
+      u = validate_vector(u)
+      v = validate_vector(v)
+      return np.dot(np.dot(u.T, components), v)
+
+    return similarity_fun
+
+  def get_bilinear_matrix(self):
+    """Returns a copy of the Bilinear matrix learned by the similarity learner.
```
bellet, line 346: Bilinear --> bilinear

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+
+    Returns
+    -------
+    M : `numpy.ndarray`, shape=(n_features, n_features)
+      The copy of the learned Bilinear matrix.
```
bellet, line 351: Bilinear --> bilinear

bellet (changes requested):
Some small comments on the implementation of the class, but otherwise looks good! I still need to review the test part, which is the big chunk

test/test_mahalanobis_mixin.py
```diff
@@ -1,3 +1,7 @@
+"""
+Tests all functionality for Mahalanobis Learners. Correctness, use cases,
```
bellet, line 2: Learners --> learners

test/test_mahalanobis_mixin.py
```diff
@@ -100,28 +106,9 @@ def test_pair_distance_finite(estimator, build_dataset):
...
-  msg = ("3D array of formed tuples expected{}. Found 2D array "
-         "instead:\ninput={}. Reshape your data and/or use a preprocessor.\n"
-         .format(context, tuples[1]))
-  with pytest.raises(ValueError) as raised_error:
-    model.pair_distance(tuples[1])
-  assert str(raised_error.value) == msg
-
-
 def check_is_distance_matrix(pairwise):
+  """Returns True if the matrix is positive, symmetrc, the diagonal is zero,
```
bellet, line 110: typo: symmetrc

test/test_mahalanobis_mixin.py
```diff
@@ -247,21 +236,8 @@ def test_get_metric_is_pseudo_metric(estimator, build_dataset):
             np.isclose(metric(a, c), metric(a, b) + metric(b, c), rtol=1e-20))
 
 
-@pytest.mark.parametrize('estimator, build_dataset', metric_learners,
-                         ids=ids_metric_learners)
-def test_get_metric_compatible_with_scikit_learn(estimator, build_dataset):
```
bellet, line 252: where did this test go?
it should probably be kept here only because the output of `get_metric` only has the correct semantic (of a distance function) for Mahalanobis learners, not for bilinear learners (which returns a similarity function)

test/test_mahalanobis_mixin.py
```diff
@@ -750,3 +726,35 @@ def test_deterministic_initialization(estimator, build_dataset):
...
+  with pytest.raises(NotFittedError):
+    estimator.pair_score(input_data)
+  with pytest.raises(NotFittedError):
+    estimator.pair_distance(input_data)
+  with pytest.raises(NotFittedError):
+    estimator.get_metric()
+  with pytest.raises(NotFittedError):
+    estimator.get_mahalanobis_matrix()
+  with pytest.raises(NotFittedError):
+    estimator.transform(input_data)
```
bellet, line 760: Some of these methods are present for all metric learners (`score_pairs`,  `pair_score`, `get_metric`) so they shoud be tested directly in `test_base_metric.py` instead of here

Additionally, there should be a similar test in `test_bilinear_mixin.py` for methods specific to bilinear similarities (such as `get_bilinear_matrix`)

test/test_pairs_classifiers.py
```diff
@@ -545,11 +539,12 @@ def breaking_fun(**args):  # a function that fails so that we will miss
...
 
 
-@pytest.mark.parametrize('estimator, build_dataset', pairs_learners,
-                         ids=ids_pairs_learners)
+@pytest.mark.parametrize('estimator, build_dataset', pairs_learners_m,
+                         ids=ids_pairs_learners_m)
 def test_accuracy_toy_example(estimator, build_dataset):
   """Test that the accuracy works on some toy example (hence that the
-  prediction is OK)"""
+  prediction is OK). This test is designed for Mahalanobis learners only,
```
bellet, line 546: maybe we can easily design a similar test for bilinear learners, eg by using half the similarity score as the threshold?

test/test_quadruplets_classifiers.py
```diff
@@ -31,21 +36,27 @@ def test_predict_only_one_or_minus_one(estimator, build_dataset,
...
 
 
-@pytest.mark.parametrize('estimator, build_dataset', quadruplets_learners,
-                         ids=ids_quadruplets_learners)
+@pytest.mark.parametrize('estimator, build_dataset', quadruplets_learners_m,
+                         ids=ids_quadruplets_learners_m)
 def test_accuracy_toy_example(estimator, build_dataset):
   """Test that the default scoring for quadruplets (accuracy) works on some
-  toy example"""
+  toy example. This test is designed for Mahalanobis learners only,
```
bellet, line 58: similarly here

test/test_sklearn_compat.py
```diff
@@ -110,6 +112,54 @@ def generate_array_like(input_data, labels=None):
   return input_data_changed, labels_changed
 
 
+# TODO: Find a better way to run this test and the next one, to avoid
+# duplicated code.
```
bellet, line 116: does this TODO still holds, or you have precisely done that?

test/test_triplets_classifiers.py
```diff
@@ -56,7 +62,7 @@ def test_no_zero_prediction(estimator, build_dataset):
     assert_array_equal(X[1], x)
   with pytest.raises(AssertionError):
     assert_array_equal(X[1], y)
-  # Assert the distance is the same for both
+  # Assert the distance is the same for both -> Wont work for b. similarity
```
bellet, line 65: b. --> bilinear to be more explicit

bellet: But anyway, is it so important to test this distance? (not really the purpose of the test) Because otherwise it looks like this test would pass also for bilinear learners

test/test_triplets_classifiers.py
```diff
@@ -75,21 +81,27 @@ def test_no_zero_prediction(estimator, build_dataset):
...
 
 
-@pytest.mark.parametrize('estimator, build_dataset', triplets_learners,
-                         ids=ids_triplets_learners)
+@pytest.mark.parametrize('estimator, build_dataset', triplets_learners_m,
+                         ids=ids_triplets_learners_m)
 def test_accuracy_toy_example(estimator, build_dataset):
   """Test that the default scoring for triplets (accuracy) works on some
-  toy example"""
+  toy example. This test is designed for Mahalanobis learners only,
```
bellet, line 103: and again here

test/test_utils.py
```diff
@@ -1,3 +1,8 @@
+"""
+Tests preprocesor, warnings, errors. Also made util functions to build datasets
```
bellet, line 2: typo: preprocesor

test/test_utils.py
```diff
@@ -28,14 +34,109 @@
...
+  """
+  Basic array 'X, y' for testing when using a preprocessor, for instance,
+  for clustering. For supervised learners.
+
+  If no preprocesor: 'data' are raw points, 'target' are dummy labels,
+  'preprocesor' is None, and 'to_transform' are points.
+
+  If preprocessor: 'data' are point indices, 'target' are dummy labels,
+  'preprocessor' are unique points, 'to_transform' are points.
+  """
```
bellet, line 139: same comment on quotes

test/test_utils.py
```diff
@@ -46,7 +147,16 @@ def build_classification(with_preprocessor=False):
...
-  """Basic array for testing when using a preprocessor"""
+  """
+  Basic array 'X, y' for testing when using a preprocessor, for regression.
+  For supervised learners.
+
+  If no preprocesor: 'data' are raw points, 'target' are dummy labels,
+  'preprocesor' is None, and 'to_transform' are points.
+
+  If preprocessor: 'data' are point indices, 'target' are dummy labels,
+  'preprocessor' are unique points, 'to_transform' are points.
```
bellet, line 158: same comment on quotes

test/test_utils.py
```diff
@@ -58,6 +168,8 @@ def build_regression(with_preprocessor=False):
 
 
 def build_data():
+  """Aux function: Returns 'X, pairs' taken from the iris dataset, where
```
bellet, line 171: same comment on quotes

test/test_utils.py
```diff
@@ -70,7 +182,17 @@ def build_data():
...
+  """
+  For all pair weakly-supervised learners.
+
+  Returns: data, target, preprocessor, to_transform.
+
+  If no preprocesor: 'data' are raw pairs, 'target' are dummy labels,
+  'preprocesor' is None, and 'to_transform' are points.
+
+  If preprocessor: 'data' are pair indices, 'target' are dummy labels,
+  'preprocessor' are unique points, 'to_transform' are points.
```
bellet, line 194: same comment on quotes

test/test_utils.py
```diff
@@ -85,6 +207,17 @@ def build_pairs(with_preprocessor=False):
...
+  """
+  For all triplet weakly-supervised learners.
+
+  Returns: data, target, preprocessor, to_transform.
+
+  If no preprocesor: 'data' are raw triplets, 'target' are dummy labels,
+  'preprocesor' is None, and 'to_transform' are points.
+
+  If preprocessor: 'data' are triplets indices, 'target' are dummy labels,
+  'preprocessor' are unique points, 'to_transform' are points.
```
bellet, line 219: same comment on quotes

test/test_utils.py
```diff
@@ -112,59 +255,130 @@ def build_quadruplets(with_preprocessor=False):
...
+                 (NCA(), build_classification),
+                 (RCA(), build_classification),
+                 (ITML_Supervised(max_iter=5), build_classification),
+                 (LSML_Supervised(), build_classification),
+                 (MMC_Supervised(max_iter=5), build_classification),
+                 (RCA_Supervised(num_chunks=5), build_classification),
+                 (SDML_Supervised(prior='identity', balance_param=1e-5),
+                 build_classification),
+                 (SCML_Supervised(n_basis=80), build_classification)]
+ids_classifiers_m = list(map(lambda x: x.__class__.__name__,
```
bellet, line 292: suffix `m` is not very explicit. I suggest `maha`

test/test_utils.py
```diff
@@ -112,59 +255,130 @@ def build_quadruplets(with_preprocessor=False):
...
+tuples_learners_m = pairs_learners_m + triplets_learners_m + \
+                  quadruplets_learners_m
+ids_tuples_learners_m = ids_pairs_learners_m + ids_triplets_learners_m \
+                      + ids_quadruplets_learners_m
+
+supervised_learners_m = classifiers_m + regressors_m
+ids_supervised_learners_m = ids_classifiers_m + ids_regressors_m
+
+metric_learners_m = tuples_learners_m + supervised_learners_m
+ids_metric_learners_m = ids_tuples_learners_m + ids_supervised_learners_m
```
bellet, line 309: same here

test/test_utils.py
```diff
@@ -112,59 +255,130 @@ def build_quadruplets(with_preprocessor=False):
...
+tuples_learners_b = pairs_learners_b + triplets_learners_b + \
+                  quadruplets_learners_b
+ids_tuples_learners_b = ids_pairs_learners_b + ids_triplets_learners_b \
+                      + ids_quadruplets_learners_b
+
+supervised_learners_b = classifiers_b
+ids_supervised_learners_b = ids_classifiers_b
+
+metric_learners_b = tuples_learners_b + supervised_learners_b
+ids_metric_learners_b = ids_tuples_learners_b + ids_supervised_learners_b
```
bellet, line 345: suffix `b` is not very explicit. I suggest `bilin`

test/test_utils.py
```diff
@@ -112,59 +255,130 @@ def build_quadruplets(with_preprocessor=False):
...
+                                   [learner for (learner, _) in
+                                    triplets_learners_b]))
+
+pairs_learners_b = [(MockPairIdentityBilinearLearner(), build_pairs)]
+ids_pairs_learners_b = list(map(lambda x: x.__class__.__name__,
+                                [learner for (learner, _) in
+                                 pairs_learners_b]))
+# -- Supervised
+classifiers_b = [(RandomBilinearLearner(), build_classification),
+                 (IdentityBilinearLearner(), build_classification)]
```
bellet, line 330: you could add those as regressors as well

test/test_utils.py
```diff
@@ -28,14 +34,109 @@
...
+      super().__init__(preprocessor=preprocessor)
+
+  def fit(self, triplets):
+    triplets = self._prepare_inputs(triplets, type_of_inputs='tuples')
+    self.d_ = np.shape(triplets[0][0])[-1]
+    self.components_ = np.identity(self.d_)
+    return self
+
+
+class MockQuadrpletsIdentityBilinearLearner(BilinearMixin,
```
bellet, line 108: typo: Quadrplets

test/test_utils.py
```diff
@@ -112,59 +255,130 @@ def build_quadruplets(with_preprocessor=False):
...
+                             [learner for (learner, _) in
+                              classifiers_b]))
+# -- Bilinear sets
+tuples_learners_b = pairs_learners_b + triplets_learners_b + \
+                  quadruplets_learners_b
+ids_tuples_learners_b = ids_pairs_learners_b + ids_triplets_learners_b \
+                      + ids_quadruplets_learners_b
+
+supervised_learners_b = classifiers_b
+ids_supervised_learners_b = ids_classifiers_b
```
bellet, line 342: and add them here

test/test_utils.py
```diff
@@ -112,59 +255,130 @@ def build_quadruplets(with_preprocessor=False):
...
+# -- Weakly Supervised + Supervised learners
+metric_learners = metric_learners_m + metric_learners_b
+ids_metric_learners = ids_metric_learners_m + ids_metric_learners_b
+
+# -- For sklearn pipeline: Pair + Supervised learners
+metric_learners_pipeline = pairs_learners_m + pairs_learners_b + \
+                         supervised_learners_m + supervised_learners_b
+ids_metric_learners_pipeline = ids_pairs_learners_m + ids_pairs_learners_b +\
+                             ids_supervised_learners_m + \
+                             ids_supervised_learners_b
```
bellet, line 374: I forgot the rationale for this list. What is the criterion for inclusion?
My first thought would be that it needs to implement the `transform` method. In which case bilinear learners should not be part of this. But at the same time we include pair learners and not others (although they can all implement transform as long as they are Mahalanobis)
Are you able to clarify? And if so, add a comment to explain this explicitly

metric_learn/base_metric.py
```diff
@@ -242,6 +242,118 @@ def transform(self, X):
...
+  Algorithm that learns a Bilinear (pseudo) similarity :math:`s_M(x, x')`,
+  defined between two column vectors :math:`x` and :math:`x'` by: :math:
+  `s_M(x, x') =  x M x'`, where :math:`M` is a learned matrix. This matrix
+  is not guaranteed to be symmetric nor positive semi-definite (PSD). Thus
+  it cannot be seen as learning a linear transformation of the original
+  space like Mahalanobis learning algorithms.
+
+  Attributes
+  ----------
+  components_ : `numpy.ndarray`, shape=(n_components, n_features)
```
bellet, line 257: Two important comments:
- The shape is incorrect: it will always be (n_features, n_features)
- I am not sure we should name this `components_` because this is typically used to denote a projection matrix used to transform the data (as in PCA, or Mahalanobis learners). I think `bilinear_matrix_` would be more appropriate. Not sure if having a different name would be at odds with the generality of some tests?

bellet (changes requested):
Here's finally the review of the test part, with important additional comments on `components_` name for bilinear similarities.

Could you also verify that the new class `BilinearMixin` is included in the doc in the same way as `MahalanobisMixin`?

test/test_mahalanobis_mixin.py
mvargas33: I moved it to `test_sklearn_compat.py` as it's sklearn-specific, but yes, it's for Mahalanobis learners only, so I can keep it here

test/test_mahalanobis_mixin.py
mvargas33: Should I build one generic test in `test_base_metric.py` for all kinds of leraners, at the cost of using `isinstance()` (as in `test_deprecated_score_pairs_same_result`),

or

Should I build one test for common functions in `test_base_metric.py`, and replicate the same test for specific functions in mahalanobis, bilinear and classifiers test files?

test/test_sklearn_compat.py
mvargas33: I can use `isinstance()` and join the tests `test_array_like_inputs_mahalanobis` and `test_array_like_inputs` that is just next to the first.

or

Have these two tests separately, which I dont like because there is duplicated code.

Take `test_deprecated_score_pairs_same_result` at `test_base_metric.py` as an example test using `isinstance()`

test/test_sklearn_compat.py
```diff
@@ -461,5 +498,18 @@ def test_dont_overwrite_parameters(estimator, build_dataset,
...
+  """Check that the metric returned by get_metric is compatible with
+  scikit-learn's algorithms using a custom metric, DBSCAN for instance"""
+  input_data, labels, _, X = build_dataset()
+  model = clone(estimator)
+  set_random_state(model)
+  model.fit(*remove_y(estimator, input_data, labels))
+  clustering = DBSCAN(metric=model.get_metric())
+  clustering.fit(X)
+
+
```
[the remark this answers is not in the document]
mvargas33: Ah! Ok, I just have to change the list of metric learners

test/test_mahalanobis_mixin.py
mvargas33: I read the comment below

doc/introduction.rst
```diff
@@ -4,17 +4,16 @@
...
 (Euclidean, City-Block, Cosine, etc.) using a priori knowledge of the
 domain. However, it is often difficult to design metrics that are well-suited
 to the particular data and task of interest.
 
-Distance metric learning (or simply, metric learning) aims at
-automatically constructing task-specific distance metrics from (weakly)
-supervised data, in a machine learning manner. The learned distance metric can
-then be used to perform various tasks (e.g., k-NN classification, clustering,
-information retrieval).
+Metric learning (or simply, metric learning) aims at automatically constructing
```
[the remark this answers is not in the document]
mvargas33: Removed

doc/introduction.rst
```diff
@@ -25,19 +24,19 @@ of supervision available about the training data:
...
   at the tuple level (typically pairs, triplets, or quadruplets of
   data points). A classic example of such weaker supervision is a set of
-  positive and negative pairs: in this case, the goal is to learn a distance
+  positive and negative pairs: in this case, the goal is to learn a
   metric that puts positive pairs close together and negative pairs far away.
 
 Based on the above (weakly) supervised data, the metric learning problem is
 generally formulated as an optimization problem where one seeks to find the
-parameters of a distance function that optimize some objective function
+parameters of a function that optimize some objective function
```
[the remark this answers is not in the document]
mvargas33: Changed.

doc/supervised.rst
```diff
@@ -62,87 +58,71 @@ NCA(init='auto', max_iter=100, n_components=None,
...
-  do that, use the `get_metric` method.
-
->>> metric_fun = nca.get_metric()
->>> metric_fun([3.5, 3.6], [5.6, 2.4])
-0.4962707194621285
-
-- Alternatively, you can use `pair_score` to return the **score** between
-  pairs of points (the larger the score, the more similar the pair).
-  For Mahalanobis learners, it is equal to the opposite of the distance.
+We can now use the learned metric to **score** new pairs of points with ``pair_score``
```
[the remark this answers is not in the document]
mvargas33: Changed, changed in `weakly_supervised.rst` as well

doc/supervised.rst
mvargas33: It's rendered as "Mahalanobis distances", so it felt redundant to repeat it again. See the render [here](https://imgur.com/a/I4Iw6mb)

doc/weakly_supervised.rst
mvargas33: Same, its rendered as "Mahalanobis distances", so its a bit redundant to repeat the noun

test/test_base_metric.py
mvargas33: Fixed

metric_learn/base_metric.py
mvargas33: Duplicated comment

metric_learn/base_metric.py
mvargas33: Changed to W

metric_learn/base_metric.py
mvargas33: Changed

metric_learn/base_metric.py
mvargas33: It's still needed for the tests where all kinds of learners are involved, otherwise an "not implemented method" may appear.

This will change when we remove `socre_pairs` in one or two releses in the future.

I added the docstring

metric_learn/base_metric.py
mvargas33: Changed.

metric_learn/base_metric.py
mvargas33: changed.

metric_learn/base_metric.py
mvargas33: changed.

metric_learn/base_metric.py
mvargas33: Changed, in the tests as well

metric_learn/base_metric.py
mvargas33: Changed.

metric_learn/base_metric.py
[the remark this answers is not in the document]
mvargas33: Changed.

metric_learn/base_metric.py
mvargas33: Changed the ref everywhere to `bilinear_similarities` to be consistent.

Changed Bilinear --> bilinear everywhere as well.

metric_learn/base_metric.py
mvargas33: Changed.

metric_learn/base_metric.py
mvargas33: Added.

metric_learn/base_metric.py
mvargas33: Changed.

metric_learn/base_metric.py
mvargas33: Changed.

bellet:
@wdevazelhes @perimosocordiae @terrytangyuan it would be great if one of you can review this PR as well in the next couple of weeks (@mvargas33 will be off for ~2 weeks - so he can finalize the PR once he is back)

perimosocordiae:
I resolved merge conflicts, but there are still a bunch of review items to address here.

PR #330, opened 2021-09-21
[WIP] OASIS algorithm implementation

mvargas33:
Hi!

I am currently implementing the [OASIS](https://www.jmlr.org/papers/volume11/chechik10a/chechik10a.pdf) algorithm and I open this PR to make the implementation transparent while working on it. Any discussion, question or comments is very welcomed.

This PR is under the WIP (Work In Progress) tag because as of now, I have a draft implementation of the algorithm out-of-the-package itself. It's a file in the root directory, with a test file in root as well.

Over these days I will move the algorithm to metric_learn folder to make it compatible with the current API. Same for testing.

Current testing only checks that nothing is broken, I'll make some test regarding KNN tasks to verify that the algorithm performs better at least for a handmade toy test.

This PR depends on the Bilinear PR #329 acceptance beforehand.


bellet:
@mvargas33 Let us know when some part of the code is ready for a first round of review!

metric_learn/oasis.py
```diff
@@ -0,0 +1,293 @@
...
+    # (pi+ - pi-)
+    diff = np.subtract(triplet[1], triplet[2])  # Shape (, d)
+    result = []
+
+    # For each scalar in first triplet, multiply by the diff of pi+ and pi-
+    for v in triplet[0]:
+        result.append(v * diff)
+
+    return np.array(result)  # Shape (d, d)
+
```
mvargas33, line 137: @bellet So this is the core part of OASIS, I've separated `_loss`, `_frobenius_squared` and `_vi_matrix` on purpose to make the code more readable.

Small comments:

`self.indices` has the order in which the triplets will be seen
`self.n_iter` matches the size of self.indices. It's the number of iterations
`self.components_` is in this case the whole M matrix of bilinear similarity, as there is no transformation here
`partial_fit` does the same as fit but with new triplets and reusing the existing `self.components_` in the object

Paper is [here ](https://www.jmlr.org/papers/volume11/chechik10a/chechik10a.pdf)

mvargas33 (commented):
It will be good to take a deep look at the core part of OASIS, comparing it directly with the paper

metric_learn/oasis.py
mvargas33: Another note: In line 90, `loss / fs` can give a warning about zero division, when `fs` is zero, because `_vi` is zero. This happens particularly when a triplet with zeros in all features is seen.

In that case `loss/fs` will become `+inf`, but `np.minimum` will always choose `self.c`

metric_learn/oasis.py
```diff
@@ -0,0 +1,375 @@
...
+    """
+    # (pi+ - pi-)
+    diff = np.subtract(triplet[1], triplet[2])  # Shape (, d)
+    result = []
+
+    # For each scalar in first triplet, multiply by the diff of pi+ and pi-
+    for v in triplet[0]:
+        result.append(v * diff)
+
+    return np.array(result)  # Shape (d, d)
```
mvargas33, line 134: This part is correct, just checked manually, but I've found a better way to compute it. This is equivalent to:
```python
return np.outer(triplet[0], np.subtract(triplet[1], triplet[2]))
```
For d=100, and executing this function n=1.000.000 times, I get the following timings:
```
Current solution: 139.7970213389999 [s]
Using numpy outer: 13.34107805200074 [s]
```
I've changed it for this optimized approach

PR #331, opened 2021-10-01
Disallow 0 on Triplets predictions

mvargas33:
Right now, the zero value is allowed when predicting triplets, because of `np.sign` wich returns -1 if the input is below zero, +1 if the input is higher than zero, and 0 if the input is zero.

Triplets of the form: (x, y, y), (x, x, x) and (x, u, v), with u and v orthogonal to x with the same norm, cause trouble.

This is because the distance between x and the second point, x and third point is the same, causing `decision_function` to be zero, and then causing `predict` to be zero in the `_TripletsClassifierMixin`.

For bilinear similarity it's even more relaxed, as the norm of u and v does not need to be the same.

By definition, the semantic of a triplet says that the distance between the first and the second point should be **strictly higher** than the distance between the first and the third point, then any zero prediction is invalid.

I propose to force zero predictions to be -1 instead, as the definition of a triplet is not satisfied.

In practice, there should not be that many border cases.

Finally, I also made a test to show this problem.


metric_learn/base_metric.py
```diff
@@ -602,7 +602,9 @@ def predict(self, triplets):
     prediction : `numpy.ndarray` of floats, shape=(n_constraints,)
       Predictions of the ordering of pairs, for each triplet.
     """
-    return np.sign(self.decision_function(triplets))
+    return np.array([-1 if (t <= 0) else 1 for t in
+                    self.decision_function(triplets)])
+    # return np.sign(self.decision_function(triplets))
```
bellet, line 607: remove this line

metric_learn/base_metric.py
```diff
@@ -602,7 +602,9 @@ def predict(self, triplets):
     prediction : `numpy.ndarray` of floats, shape=(n_constraints,)
       Predictions of the ordering of pairs, for each triplet.
     """
-    return np.sign(self.decision_function(triplets))
+    return np.array([-1 if (t <= 0) else 1 for t in
+                    self.decision_function(triplets)])
```
bellet, line 606: How about just `2 * (self.decision_function(triplets) > 0) -1`?

bellet (changes requested):
Thanks @mvargas33. I have not looked at the tests yet (some tests are failing)

metric_learn/base_metric.py
mvargas33: Done

metric_learn/base_metric.py
mvargas33: Just changed, this way is much simpler

mvargas33:
This PR is ready to be merged

test/test_triplets_classifiers.py
```diff
@@ -63,3 +64,48 @@ def test_accuracy_toy_example(estimator, build_dataset):
...
+
+@pytest.mark.parametrize('estimator, build_dataset', triplets_learners,
+                         ids=ids_triplets_learners)
+def test_no_zero_prediction(estimator, build_dataset):
+  """
+  Test that all predicted values are in {-1, 1}, even when the
+  distance d(x,y) and d(x,z) is the same for a triplet of the
+  form (x, y, z).
+  """
+  # Dummy fit
```
bellet, line 77: remove

test/test_triplets_classifiers.py
```diff
@@ -63,3 +64,48 @@ def test_accuracy_toy_example(estimator, build_dataset):
...
+  triplets, _, _, X = build_dataset(with_preprocessor=False)
+  # Force 3 dimentions only, to use cross product and get easy orthogonal vec.
+  triplets = np.array([[t[0][:3], t[1][:3], t[2][:3]] for t in triplets])
+  X = np.array([x[:3] for x in X])
```
bellet, line 81: this should just be a matrix, so you can do `X = X[:, :3]` no?

test/test_triplets_classifiers.py
```diff
@@ -63,3 +64,48 @@ def test_accuracy_toy_example(estimator, build_dataset):
...
+  # Dummy fit
+  estimator = clone(estimator)
+  set_random_state(estimator)
+  estimator.fit(triplets)
+  # we force the transformation to be identity, to force euclidean distance
+  estimator.components_ = np.eye(X.shape[1])
+
+  # Get two orthogonal vectors in respect to X[1]
+  k = X[1]/np.linalg.norm(X[1])  # Normalize first vector
```
bellet, line 90: spaces around `/`

test/test_triplets_classifiers.py
```diff
@@ -63,3 +64,48 @@ def test_accuracy_toy_example(estimator, build_dataset):
...
+  # Form the three scenarios where predict() gives 0 with numpy.sign
+  triplets_test = np.array(  # Critical examples
+    [[X[0], X[2], X[2]],
+     [X[1], X[1], X[1]],
+     [X[1], x, y]])
+  # Predict
+  predictions = estimator.predict(triplets_test)
+  # Count non -1 or 1 values
+  not_valid = [e for e in predictions if e not in [-1, 1]]
+  assert len(not_valid) == 0
```
bellet, line 111: this does not really match the semantics of the name of the test (which is to verify that there is no zero value). so either you just check that there is no zero value (and adapt the docstring of the test accordingly), or you can try to merge this with the already existing test `test_predict_only_one_or_minus_one`.

PR #333, opened 2021-10-01
[WIP] score_pairs refactor

mvargas33:
See #329 

In the aim of making metric learn compatible with similarity learners like OASIS and pseudo-distance Mahalanobis learners (all current learners) I take a step forward deprecating `score_pairs` with a `FutureWarning`, and replacing it with `pair_score` and `pair_distance`, as discussed in #329 .

After thinking a while between `pairwise_similarity` and `pair_similarity` names, I think that using ''pairwise'' gives the user the intuition of similarity between all pairs, so a matrix is expected as output, not a list. Just like sklearn's `pairwise_distances` (See [here](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html)).

Right now it does not makes too much sense to have `pair_distances` and `pair_similarity` only for Mahalanobis learners because `pair_similarity` it's just the inverse of the distance. But, this PR is needed to proceed with #329 and then with #330.

As discussed before, this change makes the code very extendible for new types of learners, for instance a `StringEdit` learner could fit in the library seamlessly extending from `BaseLearner` regardless the algorithm learns a similarity or a pseudo-distance.

It also does not affect `Classifiers` that much, as they can work with `pair_distance` from now on.

In contrast to this proposal, using `isinstance()` for each kind of learner inside the current `score_pairs` does not look like a good solution, in my humble opinion.

Ps: If the discussion goes in favor, I'll check all current tests, one by one, to make sure they don't become broken.


mvargas33:
The general functionality is ready. What is left to be done is to produce a better documentation. I'm on it.

mvargas33:
> @mvargas33 is this ready to review?

Yes, it is ! The code and the docs are ready

bellet:
Merged, thanks @mvargas33!

PR #338, opened 2021-10-14
[MRG][DOC] Fixes almost all warnings in the docs

### LINKED ISSUE #226: deprecations should appear in red
In scikit-learn or imbalanced-learn for instance, when there is a deprecation in one of the docstrings, it appears in red. I think it's cool to catch the eye, but still couldn't find how to do that

### LINKED ISSUE #305: [DOC] Remove `any` flag
In the end the `any` flag set here: https://github.com/scikit-learn-contrib/metric-learn/blob/66a12edbfba74774d84e6cd0f4f0fd5d438a79f8/doc/conf.py#L42
may cause more harm than good, because it displays a lot of errors when building the doc and sometimes it's hard to find true errors, so in the end we should maybe come back to a manual linking in the doc

mvargas33:
Closes #226
Closes #305

While doing the docs for #333 #329 and #330 I've encountered an enormous amount of warnings while executing `make html`. This PR solves most of these warnings while leaving a small number of warnings open to discussion. Here's a detail of the changes:

1. **Correct dependencies**: In `setup.py` solved a typo in `shinx_rtd_theme` for `sphinx_rtd_theme`. Also added `sphinx-gallery `and `matplotlib`, both needed to render the docs for the examples.
2. **Fixed deprecated functions**: The docs won't compile as they are. In `conf.py`, at line 82 and 83 `add_stylesheet` need to be changed for `add_css_file`, and `add_javascript` for `add_js_file`. Also `html4_writer = True` at line 72 is deprecated by an explicit warning from `sphinx_rtd_theme` , so it's commented now.
3. **Most warnings come from the `any` flag**: As stated in #305 the `any` flag is causing more trouble than benefit. So I commented it. Tested some refs in the docs and they seem to work as usual.
4. **`Topic` directive without body**: Each `topic` directive regarding ''Example Code'' had no body, mainly in `supervised.rst` and `weakly_supervised.rst`. Introduced a brief description of the code whenever necessary, like _"Example Code: A basic usage of this model with the Iris dataset from scikit-learn._"
5. **Minor issues**: Some warnings were solved simply by using the right indentation.
6. **Last remaining warnings**: In `supervised.rst`, `weakly_supervised.rst` and `unsupervised.rst` there is still an issue:

- By definition, while using the directive `.. topic:: References:` it means that something was referred previously. But in this case, this section is used as "find the original paper here" or "base code was taken from here" at the end of each algorithm description. None of the references is mention earlier in any text body, and because of that this kind of warning shows up: `/metric-learn/doc/unsupervised.rst:39: WARNING: Footnote [1] is not referenced.`
- Also, in a single `rst` file the numeration of references must strictly increase, but in this case, each algorithm has references from 1-2 or 1-3, and because of that there are a lot of warnings of the form `/metric-learn/doc/supervised.rst:378: WARNING: Duplicate explicit target name: "1".` Which makes sense, because if I want to reference the target [1] in a text body, there are 6 or 7 references with that name.

Some possible solutions:
- Don't use  `.. topic:: References:` but just a list instead, this will delete all warnings.
- Use the reference as they are intended to be used: use them in the text body of each algorithm, and use a unique incresing number id for each reference in the same file.

Besides this last problem, all warnings were fixed.

### DIFF
--- a/doc/_static/css/styles.css
+++ b/doc/_static/css/styles.css
@@ -21,3 +21,16 @@
   margin-left: 0px !important;
   margin-bottom: 0px !important;
 }
+.deprecated {
+  color: #b94a48;
+  background-color: #F3E5E5;
+  border-color: #eed3d7;
+  margin-top: 0.5rem;
+  padding: 0.5rem;
+  border-radius: 0.5rem;
+  margin-bottom: 0.5rem;
+}
+
+.deprecated p {
+  margin-bottom: 0 !important;
+}
--- a/doc/_static/css/styles.css
+++ b/doc/_static/css/styles.css
@@ -1,0 +1,23 @@
+.hatnote {
+    border-color: #e1e4e5 ;
+    border-style: solid ;
+    border-width: 1px ;
+    font-size: x-small ;
+    font-style: italic ;
+    margin-left: auto ;
+    margin-right: auto ;
+    margin-bottom: 24px;
+    padding: 12px;
+}
+.hatnote-gray {
+  background-color: #f5f5f5 
+}
+.hatnote li {
+  list-style-type: square;
+  margin-left: 12px !important;
+}
+.hatnote ul {
+  list-style-type: square;
+  margin-left: 0px !important;
+  margin-bottom: 0px !important;
+}
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -81,6 +81,7 @@ autosummary_generate = True
 def setup(app):
    app.add_js_file('js/copybutton.js')
    app.add_css_file("basic.css")
+   app.add_css_file("css/styles.css")
 
 
 # Remove matplotlib agg warnings from generated doc when using plt.show
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -38,9 +38,6 @@ html_theme = 'sphinx_rtd_theme'
 html_static_path = ['_static']
 htmlhelp_basename = 'metric-learndoc'
 
-# Option to only need single backticks to refer to symbols
-# default_role = 'any'
-
 # Option to hide doctests comments in the documentation (like # doctest:
 # +NORMALIZE_WHITESPACE for instance)
 trim_doctest_flags = True
@@ -67,11 +64,6 @@ sphinx_gallery_conf = {
 # generate autosummary even if no references
 autosummary_generate = True
 
-# Switch to old behavior with html4, for a good display of references,
-# as described in https://github.com/sphinx-doc/sphinx/issues/6705
-# html4_writer = True
-
-
 # Temporary work-around for spacing problem between parameter and parameter
 # type in the doc, see https://github.com/numpy/numpydoc/issues/215. The bug
 # has been fixed in sphinx (https://github.com/sphinx-doc/sphinx/pull/5976) but
@@ -80,8 +72,7 @@ autosummary_generate = True
 # https://github.com/readthedocs/sphinx_rtd_theme/pull/747/files
 def setup(app):
    app.add_js_file('js/copybutton.js')
-   app.add_css_file("basic.css")
-   app.add_css_file("css/styles.css")
+   app.add_css_file('css/styles.css')
 
 
 # Remove matplotlib agg warnings from generated doc when using plt.show
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -64,6 +64,7 @@ sphinx_gallery_conf = {
 # generate autosummary even if no references
 autosummary_generate = True
 
+
 # Temporary work-around for spacing problem between parameter and parameter
 # type in the doc, see https://github.com/numpy/numpydoc/issues/215. The bug
 # has been fixed in sphinx (https://github.com/sphinx-doc/sphinx/pull/5976) but
@@ -71,8 +72,8 @@ autosummary_generate = True
 # In an ideal world, this would get fixed in this PR:
 # https://github.com/readthedocs/sphinx_rtd_theme/pull/747/files
 def setup(app):
-   app.add_js_file('js/copybutton.js')
-   app.add_css_file('css/styles.css')
+  app.add_js_file('js/copybutton.js')
+  app.add_css_file('css/styles.css')
 
 
 # Remove matplotlib agg warnings from generated doc when using plt.show
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -39,7 +39,7 @@ html_static_path = ['_static']
 htmlhelp_basename = 'metric-learndoc'
 
 # Option to only need single backticks to refer to symbols
-#default_role = 'any'
+# default_role = 'any'
 
 # Option to hide doctests comments in the documentation (like # doctest:
 # +NORMALIZE_WHITESPACE for instance)
@@ -69,7 +69,7 @@ autosummary_generate = True
 
 # Switch to old behavior with html4, for a good display of references,
 # as described in https://github.com/sphinx-doc/sphinx/issues/6705
-#html4_writer = True
+# html4_writer = True
 
 
 # Temporary work-around for spacing problem between parameter and parameter
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -39,7 +39,7 @@ html_static_path = ['_static']
 htmlhelp_basename = 'metric-learndoc'
 
 # Option to only need single backticks to refer to symbols
-default_role = 'any'
+#default_role = 'any'
 
 # Option to hide doctests comments in the documentation (like # doctest:
 # +NORMALIZE_WHITESPACE for instance)
@@ -69,7 +69,7 @@ autosummary_generate = True
 
 # Switch to old behavior with html4, for a good display of references,
 # as described in https://github.com/sphinx-doc/sphinx/issues/6705
-html4_writer = True
+#html4_writer = True
 
 
 # Temporary work-around for spacing problem between parameter and parameter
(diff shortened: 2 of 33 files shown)


mvargas33:
> > Also, the references and example environments are a bit ugly visually (just a simple border with possibly overlarge margins). Maybe there is a way to make this a bit more nice?
> 
> I've changed the CSS of Example Code and References for a custom one. You can notice the differences [here](https://imgur.com/a/JPAEC43) before and after this PR.

Forgot to mention that this approach don't use `.. topic:: References` but a simple fot list instead, **finally removing all warnings from compilation**.

Also, with this last commit, **I've added all methods from all classes to their respective `Methods` directive.** I've also checked that the methods appear in alphabetical order, just as they are rendered by `autosummary`.

So in the end, each class has a summary of all of its methods (in addition to the attributes and references and so on) and then the detailed description of each method.

I also added the `autosummary` for `MetricTransformer` and `MahalanobisMixin` that was missing before at `metric_learn.rst`. I don't know if there was some reason for this, but as `BilinearMixin` will be incorported in #329 and #333 is already merged adding `pair_score` and `pair_distance`, I think its important to have these classes in the docs. For instance, the user can notice **from the docs** that `pair_distance` is not implemented in `BilinearMixin`.

This PR is ready for a code review. Summary:

- All warnings are removed from building the docs
- Add `Methods` directive to all classes, wich is a summary of all methods from that class.
- Changed CSS for _Example Code_ and _References_
- Update dependencies to build the docs (Actually they don't compile as they are before this PR)
- Removed deprecated functions in `conf.py` (for js and css)
- Removed _**any**_ flag (Fixes #305 )

doc/conf.py
```diff
@@ -69,7 +69,7 @@
 
 # Switch to old behavior with html4, for a good display of references,
 # as described in https://github.com/sphinx-doc/sphinx/issues/6705
-html4_writer = True
+# html4_writer = True
```
perimosocordiae, line 72: If this is no longer needed, can we delete it?

Same goes for default_role above.

doc/conf.py
```diff
@@ -79,8 +79,9 @@
 # In an ideal world, this would get fixed in this PR:
 # https://github.com/readthedocs/sphinx_rtd_theme/pull/747/files
 def setup(app):
-    app.add_javascript('js/copybutton.js')
-    app.add_stylesheet("basic.css")
+   app.add_js_file('js/copybutton.js')
```
perimosocordiae, line 82: Either use single or double quotes consistently, but not both (unless it helps to avoid escaping nested quotes)

metric_learn/sdml.py
```diff
@@ -285,6 +331,39 @@ class SDML_Supervised(_BaseSDML, TransformerMixin):
     The linear transformation ``L`` deduced from the learned Mahalanobis
     metric (See function `components_from_metric`.)
 
+  Methods
```
perimosocordiae, line 334: I may have missed it in your PR description, but why do we need a full method listing in each class docstring? I expect the doc building process to handle method listings automatically.

mvargas33: I expected that as well but it does not happen. So it all started because there is **a lot** of warnings like these:
```bash
/home/max/inria/metric-learn/metric_learn/sdml.py:docstring of metric_learn.sdml.SDML_Supervised:82: WARNING: autosummary: stub file not found 'metric_learn.SDML_Supervised.get_params'. Check your autosummary_generate setting.
```
See these [image](https://imgur.com/a/1ohSdHT) as reference.

And I've found the solution in [stackoverflow](https://stackoverflow.com/questions/65198998/sphinx-warning-autosummary-stub-file-not-found-for-the-methods-of-the-class-c) from a person that was also extending from the `Estimator` class and had exactly the same errors. So I followed his advice and added the `Method` directive in every class, and that removed all these bunch of warnings.

Without explicitly adding this directive, no method summary is shown in the docs, just the detailed method list.

I also thought that the `Method` directive could be incremental, so for each class I only had to add the methods description of that class, and not the method description of his parent, and grandparent class as well. But sadly it doesn't work that way either.

It may look like an overkill right now, but thinking about it, when someone has to add another algorithm in the future, you already have the docstrings from other algorithms as reference, so it won't be overwhelming.

Also, if someone wants to extend the docs, it's better for them to see the warnings/errors that only they are producing, because right now its really really hard to debug the docs with all these warnings. @wdevazelhes says it at #305 as well (in this case for the 'any' flag, the second bunch of warnings).

doc/conf.py
mvargas33: Yes. Deleted it

doc/conf.py
mvargas33: Fixed

mvargas33:
[Here](https://imgur.com/a/6yf4KO7) is a quick preview of the `Methods` directive. Its a summary between the `Attributes` and the `Method detailed descriptions`, in every class.

metric_learn/sdml.py
bellet: Regarding this: how about the second answer in the stackoverflow post you linked: https://stackoverflow.com/questions/65198998/sphinx-warning-autosummary-stub-file-not-found-for-the-methods-of-the-class-c
This could provide a way to remove the warnings. We would not have any method summary but perhaps it is not big deal, considering the effort required to have this (of course the best solution would be that it is constructed automatically, but you seem to say that it is not possible)

doc/_static/css/styles.css
```diff
@@ -0,0 +1,23 @@
...
+}
+.hatnote li {
+  list-style-type: square;
+  margin-left: 12px !important;
+}
+.hatnote ul {
+  list-style-type: square;
+  margin-left: 0px !important;
+  margin-bottom: 0px !important;
+}
```
bellet, line 23: I think it would be nice to have a numbered list, so we can still cite a reference in the text by its number if we need to in the future

doc/supervised.rst
```diff
@@ -152,7 +152,9 @@ neighbors (with same labels) of :math:`\mathbf{x}_{i}`, :math:`y_{ij}=0`
 indicates :math:`\mathbf{x}_{i}, \mathbf{x}_{j}` belong to different classes, 
 :math:`[\cdot]_+=\max(0, \cdot)` is the Hinge loss.
 
-.. topic:: Example Code:
+.. rubric:: Example Code
+
+A basic usage of this model with the Iris dataset from scikit-learn.
```
bellet, line 157: With the use of `rubric` instead of `topic`, would we still get a warning if we did not have this description sentence? If not, we could probably remove it in most cases to make things more concise, as it is often clear from the code what it is doing (here, using LMNN on iris). We can of course keep a description for examples where the code is not self-explanatory or we want to emphasize more what it is doing.

doc/supervised.rst
```diff
@@ -231,13 +235,14 @@ the sum of probability of being correctly classified:
     nca = NCA(max_iter=1000)
     nca.fit(X, Y)
 
-.. topic:: References:
+.. rubric:: References
+
+
+.. container:: hatnote hatnote-gray
```
bellet, line 241: it would be nice it there was a way to define a custom rubric or something so we only need to write something like 
``.. rubric:: References``
and the appropriate CSS style would be automatically applied

bellet: (no big deal if there is no easy way to do this)

bellet (commented):
Here are some additional comments

doc/_static/css/styles.css
mvargas33: Changed for a numbered list

doc/supervised.rst
mvargas33: After researching for while, there is not a clear path to this at all. Also, I think that even if its possible all `.. rubric::` stuff will have the same CSS, including the Examples, so I guess its not a good idea to have this CSS globally, because we don't want grey boxes everywhere, just in the references.

doc/supervised.rst
mvargas33: Just checked, we can remove the body of `.. rubirc::` and don't get any warning, so I removed all simple comments from almost everywhere except in `weakly_supervised.rst` line 67 and 100 to make a little bit a emphasis.

metric_learn/sdml.py
mvargas33: I've been hours looking to build the methods summary automatically without any warning, without success.

We either use `numpydoc_show_class_members = False` removing the warnings but loosing the automatic method summary, or

We keep the implicit `numpydoc_show_class_members = True` generating the automatic method summary but with a bunch of warnings.

Yes, I realized that the method summary its actually being generated with the docs as they are. So I will remove the hard coded method summary that I've made, as it's a lot of work and can be done automatically, despite the warnings.

mvargas33:
So I've removed the `Method` directive everywhere because the method summary can be generated automatically by `sphinx autosummary`, at a cost of having one warning per function in each class (lots of them).

I've hardly tried to keep the methods summary generation without warnings, but I failed to do so.

I think its ok for now to keep this kind of warnings, because hard-coding the `Method` directive is too much work, and less flexible while extending the code in the future.

mvargas33:
Closes #226 

I've added some extra CSS to show deprecation warnings in red as requested in this other issue. The final result looks like [this](https://imgur.com/a/Vfwhc7I).

There is no extra work from now on, you can use `.. deprecated::` directive directly and it will have the custom style.

perimosocordiae (approved):

bellet (approved):
LGTM, thanks a lot!

PR #341, opened 2021-10-28
[Tests][Warnings] Cut all warnings from SCML using a minimal solution

mvargas33:
As suggested by @bellet in #339, it's better to cut SCML warnings from the root by specifying the `n_basis` value when its possible.

As all test run the iris dataset for SCML though `build_triplets` at `test_utils.py`, then we just need to scpecify `n_basis` for `SCML` and `SCML_Supervised` in the lists.

Going through the code of `SCML`, if `n_basis` is `None`, then the following default is assigned:

Default for SCML_Supervised: ([lines 579-583](https://github.com/scikit-learn-contrib/metric-learn/blob/aaf8d44b8d31d6ea418f0bd80ef86958e5081b4c/metric_learn/scml.py#L583))
```python
n_basis = min(20*n_features, n_samples*2*min(n_classes-1, n_features) - 1)
```
Default for SCML: ([line 234](https://github.com/scikit-learn-contrib/metric-learn/blob/aaf8d44b8d31d6ea418f0bd80ef86958e5081b4c/metric_learn/scml.py#L234))
```python
n_basis = n_features*80
```
Iris dataset has 150 samples, 3 classes, 4 features. So the value is 80 for `SCML_Supervised` and 320 for `SCML`.

I changed it, but got the following tests failing.

- In `test_components_is_2D` at `test_mahalanobis_mixin.py` there is an error because of a border case. I got this warning from SCML:
```
The number of bases with nonzero weight is less than the number of features of the input, in consequence the learned transformation reduces the dimension to 0
```
And this error from the test:
```
>     assert model.components_.shape == (1, 1)  # the components must be 2D
E     assert (0, 1) == (1, 1)
E       At index 0 diff: 0 != 1
E       Use -v to get the full diff
```
So the test want to test `fit` for just one feature. Right now the feature selected from Iris is the 1st one, but  if we use 2nd dimension instead, we can avoid this border case and make the test pass. The behavior expected is the same.
```python
# trunc_data = input_data[..., :1]  # Old
trunc_data = input_data[..., 1:2]  # New
```

- There is also an error in `test_sklearn_compat.py`, `TestSklearnCompat` class, `test_scml`, `check_estimator` function if `n_basis` is arbitrary, because under the hood it calls `fit()` wich calls `_generate_bases_LDA`, and if `n_basis` is wrong, it may fail. Thus, we need to let `SCML` calc `n_basis` for us as we don't know wich random `X, y` dataset is being used. Thus, we need to use `pytest.warn` to catch the warning.

- Not an error, but an obervation in `test_triplet_diffs` and `test_lda`: These tests needs an explicitly `n_basis=None` to check if it is being set correctly internally, thus, in this case the warning needs to be caught as well.

Besides this observations, **all warnings from SCML were removed. Only 300-ish warnings are shown now across all tests.**

Also, this PR removes the `isinstance()` overkill made in `test_triplets_clasifiers.py` previously.

Best! 🎃

### DIFF
--- a/test/metric_learn_test.py
+++ b/test/metric_learn_test.py
@@ -95,7 +95,7 @@ class TestSCML(object):
                                n_informative=60, n_redundant=0, n_repeated=0,
                                random_state=42)
     X = StandardScaler().fit_transform(X)
-    scml = SCML_Supervised(random_state=42)
+    scml = SCML_Supervised(random_state=42, n_basis=399)
     scml.fit(X, y)
     csep = class_separation(scml.transform(X), y)
     assert csep < 0.7
@@ -106,7 +106,7 @@ class TestSCML(object):
                                                          [2, 0], [2, 1]]),
                                                np.array([1, 0, 1, 0])))])
   def test_bad_basis(self, estimator, data):
-    model = estimator(basis='bad_basis')
+    model = estimator(basis='bad_basis', n_basis=33)  # n_basis doesn't matter
     msg = ("`basis` must be one of the options '{}' or an array of shape "
            "(n_basis, n_features)."
            .format("', '".join(model._authorized_basis)))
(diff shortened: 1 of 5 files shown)


perimosocordiae (approved):

terrytangyuan (approved):

bellet:
Thanks a lot, @mvargas33!

PR #345, opened 2021-12-08
[WIP] Add warm_start  parameter to SCML

maxi-marufo:
Because [SCML](http://researchers.lille.inria.fr/abellet/papers/aaai14.pdf) optimization procedure is based on stochastic subgradient descent, we can save the weights after fitting the model, and use them in a following fit call (with a different set of triplets). The decision to use the `warm_start` parameter instead of a new `partial_fit` method is because [partial_fit in scikit-learn will only fit 1 epoch, whereas fit will fit for multiple epochs (until the loss converges or max_iter is reached)](https://stackoverflow.com/questions/38052342/what-is-the-difference-between-partial-fit-and-warm-start), which is the case also for SCML.

### DIFF
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -23,7 +23,8 @@ class _BaseSCML(MahalanobisMixin):
 
   def __init__(self, beta=1e-5, basis='triplet_diffs', n_basis=None,
                gamma=5e-3, max_iter=10000, output_iter=500, batch_size=10,
-               verbose=False, preprocessor=None, random_state=None):
+               verbose=False, preprocessor=None, random_state=None,
+               warm_start=False):
     self.beta = beta
     self.basis = basis
     self.n_basis = n_basis
@@ -34,6 +35,7 @@ class _BaseSCML(MahalanobisMixin):
     self.verbose = verbose
     self.preprocessor = preprocessor
     self.random_state = random_state
+    self.warm_start = warm_start
     super(_BaseSCML, self).__init__(preprocessor)
 
   def _fit(self, triplets, basis=None, n_basis=None):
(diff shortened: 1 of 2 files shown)


perimosocordiae:
This change looks fine to me, though I'm not sure when this warm-start option is useful in practice. Sorry for the extreme delay in reviewing!

@grudloff want to take a look?

grudloff:
LGTM!

bellet (commented):
Thanks @maxi-marufo !
Maybe it is possible to design a better test, something in the spirit https://github.com/scikit-learn/scikit-learn/blob/80598905e517759b4696c74ecc35c6e2eb508cff/sklearn/linear_model/tests/test_sgd.py#L274
where the idea is to check that calling `warm_start=True` is equivalent to manually setting the parameters?

bellet (approved):
LGTM @maxi-marufo, but I would like to rerun the CIs, not sure how to do this - @perimosocordiae @terrytangyuan do you know?

One simple way sould be for you @maxi-marufo to update your branch to the current state of master and push an empty commit

grudloff:
@bellet For a manual run you need to have a trigger on `workflow_dispatch ` but currently it is only set up with push and pull requests to master. So it would not be possible rn
(see https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow)

terrytangyuan:
This just needs a rebase to trigger the CI.

terrytangyuan:
CI triggered

bellet:
@maxi-marufo @terrytangyuan well looks like checks have been cancelled and/or failed?

terrytangyuan:
I think some failed which triggered cancellation of the rest.

PR #350, opened 2022-04-07
SCML: Raise ValueError if n_features larger than n_triplets

nikosmichas:
Related discussion https://github.com/scikit-learn-contrib/metric-learn/issues/348
cc @grudloff

### DIFF
--- a/metric_learn/scml.py
+++ b/metric_learn/scml.py
@@ -240,6 +240,12 @@ class _BaseSCML(MahalanobisMixin):
       raise ValueError("n_basis should be an integer, instead it is of type %s"
                        % type(self.n_basis))
 
+    if n_features > n_triplets:
+      raise ValueError(
+        "Number of features (%s) is greater than the number of triplets(%s).\n"
+        "Consider using dimensionality reduction or using another basis "
+        "generation scheme." % (n_features, n_triplets))
+
     basis = np.zeros((n_basis, n_features))
 
     # get all positive and negative pairs with lowest index first
--- a/test/test_triplets_classifiers.py
+++ b/test/test_triplets_classifiers.py
@@ -2,7 +2,12 @@ import pytest
 from sklearn.exceptions import NotFittedError
 from sklearn.model_selection import train_test_split
 
-from test.test_utils import triplets_learners, ids_triplets_learners
+from metric_learn import SCML
+from test.test_utils import (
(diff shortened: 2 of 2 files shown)


metric_learn/scml.py
```diff
@@ -240,6 +240,12 @@ def _generate_bases_dist_diff(self, triplets, X):
       raise ValueError("n_basis should be an integer, instead it is of type %s"
                        % type(self.n_basis))
 
+    if n_features > n_triplets:
+      raise ValueError(
+        "Number of features (%s) is greater than the nuber of triplets(%s).\n"
```
grudloff, line 245: small typo

nikosmichas: fixed

metric_learn/scml.py
```diff
@@ -240,6 +240,12 @@ def _generate_bases_dist_diff(self, triplets, X):
...
+        "Consider using a dimensionality reduction preprocessing or create "
+        "a new basis generation scheme." % (n_features, n_triplets))
+
```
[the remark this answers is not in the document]
nikosmichas: thanks, updated

bellet (changes requested):
Thanks for the work @nikosmichas !
This looks good to me, except that I realize that the docstring entry for the `‘triplet_diffs’` option says:

```
The basis set is constructed from the differences between points of n_basis positive or negative pairs taken from the triplets constrains.
```

Shoudn't it be something like:

```
Each basis is constructed from the differences between points of n_features positive or negative pairs taken randomly from the triplets constraints.
```

nikosmichas:
> Thanks for the work @nikosmichas ! This looks good to me, except that I realize that the docstring entry for the `‘triplet_diffs’` option says:
> 
> ```
> The basis set is constructed from the differences between points of n_basis positive or negative pairs taken from the triplets constrains.
> ```
> 
> Shoudn't it be something like:
> 
> ```
> Each basis is constructed from the differences between points of n_features positive or negative pairs taken randomly from the triplets constraints.
> ```

@bellet not sure about the exact wording. I can make the docstring change if you think that it's better. Do you thing that the change is related to this PR ?

perimosocordiae:
Merging now, we can fix the docstring separately.

bellet:
Thanks @nikosmichas !

PR #355, opened 2023-10-09
update doc for 0.7.0

wdevazelhes:


wdevazelhes:
made that PR by mistake, docs should be already pushed in the gh-pages branch 👍

PR #356, opened 2024-07-26
[DOC] Fixed small error in Code Snipplet in Docs

DaminK:
The verbose (and almost all other parameters) need to be passed to the constructor and not the .fit() function. This was not reflected in the example snipplets for LMNN in the docs and let new users (like me) to make errors based on the docs.


perimosocordiae:
Thanks!

PR #357, opened 2024-08-02
[MRG] Fix test failures due to updated packages: deprecated pytest.warns(None) syntax + GLasso update in sklearn

bellet:
This PR fixes two problems due to updates in `pytest` and `sklearn`:
- `sklearn` has changed the name of the function implementing graphical lasso, which make SDML (and associated tests) fail when SKGGM is not installed. This PR makes a quick (and dirty?) fix that does the job - @perimosocordiae @terrytangyuan if you have a better solution let me know
- `pytest` has deprecated the syntax `pytest.warns(None)` and it raises an error since v8. I have fixed this following the guidelines provided by `pytest` (https://github.com/pytest-dev/pytest/issues/9404#issue-1076710891), which is to use `warnings.catch_warnings(record=True)` instead. This is what has been done for instance in `scipy` (https://github.com/scipy/scipy/pull/15192)

### DIFF
--- a/metric_learn/sdml.py
+++ b/metric_learn/sdml.py
@@ -6,7 +6,11 @@ import warnings
 import numpy as np
 from sklearn.base import TransformerMixin
 from scipy.linalg import pinvh
-from sklearn.covariance import graphical_lasso
+try:
+  from sklearn.covariance import _graphical_lasso as graphical_lasso
+except ImportError:
+  from sklearn.covariance import graphical_lasso
+
 from sklearn.exceptions import ConvergenceWarning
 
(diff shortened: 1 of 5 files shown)


bellet:
This is ready to review/merge.

Note: I have not added the `warnings.simplefilter("error")` to filter errors as in all case we are either checking for a particular type of warning or to seek catch any warning/error.

terrytangyuan (approved):
LGTM. Thanks!

perimosocordiae (approved):
Importing the private function isn't ideal, but for now it should be ok.

